PR #1, opened 2015-01-07
Add a manpage for mgba.

bentley:
My first instinct when about to use a new emulator is to check the manpage, so I can see the command line flags and default keybindings.

I don’t know what magic CMake needs to install it in the man directories.

Gave it MPL 2.0. Would prefer something simpler like MIT honestly, but figured it was better to be consistent with the rest of mgba (unless you’re willing to relicense…).

### DIFF
--- a/doc/mgba-qt.6
+++ b/doc/mgba-qt.6
@@ -12,6 +12,7 @@
 .Sh SYNOPSIS
 .Nm mgba-qt
 .Op Fl b Ar biosfile
+.Op Fl l Ar loglevel
 .Op Fl p Ar patchfile
 .Op Fl s Ar n
 .Ar file
@@ -25,6 +26,33 @@ Specify a BIOS file to use during boot.
 If this flag is omitted,
 .Nm
 will use a high\(hylevel emulated BIOS.
+.It Fl l Ar loglevel
+Log messages during emulation.
+.Ar loglevel
+is a bitmask defining which types of messages to log:
+.Bl -inset
+.It 0x01
+fatal errors
+.It 0x02
+errors
+.It 0x04
+warnings
+.It 0x08
+informative messages
+.It 0x10
+debugging messages
+.It 0x20
+stub messages
+.It 0x100
+in\(hygame errors
+.It 0x200
+software interrupts
+.It 0x400
+emulator status messages
+.It 0x800
+serial I/O messages
+.El
+The default is to log warnings, errors, fatal errors, and status messages.
 .It Fl p Ar patchfile , Fl -patch Ar patchfile
 Specify a patch file in IPS or UPS format.
 .It Fl s Ar n , Fl -frameskip Ar n
--- a/doc/mgba-qt.6
+++ b/doc/mgba-qt.6
@@ -11,10 +11,27 @@
 .Nd Game Boy Advance emulator
 .Sh SYNOPSIS
 .Nm mgba-qt
+.Op Fl b Ar biosfile
+.Op Fl p Ar patchfile
+.Op Fl s Ar n
 .Ar file
 .Sh DESCRIPTION
 .Nm
 is a Game Boy Advance emulator.
+The options are as follows:
+.Bl -tag -width Ds
+.It Fl b Ar biosfile , Fl -bios Ar biosfile
+Specify a BIOS file to use during boot.
+If this flag is omitted,
+.Nm
+will use a high\(hylevel emulated BIOS.
+.It Fl p Ar patchfile , Fl -patch Ar patchfile
+Specify a patch file in IPS or UPS format.
+.It Fl s Ar n , Fl -frameskip Ar n
+Skip every
+.Ar n
+frames.
+.El
 .Sh CONTROLS
 The default controls are as follows:
 .Bl -hang -width "Frame advance" -compact
--- a/doc/mgba-qt.6
+++ b/doc/mgba-qt.6
@@ -93,6 +93,15 @@ configuration file.
 Default
 .Nm mgba-qt
 configuration file.
+.It Pa portable.ini
+If this file exists in the current directory,
+.Nm
+will read
+.Pa config.ini
+and
+.Pa qt.ini
+from the current directory instead of
+.Pa ~/.config/mgba .
 .El
 .Sh AUTHORS
 .An Jeffrey Pfau Aq Mt jeffrey@endrift.com
--- a/doc/mgba-qt.6
+++ b/doc/mgba-qt.6
@@ -43,7 +43,7 @@ informative messages
 .It 16
 debugging messages
 .It 32
-stub messages
+stub messages for unimplemented features
 .It 256
 in\(hygame errors
 .It 512
--- a/doc/mgba-qt.6
+++ b/doc/mgba-qt.6
@@ -25,7 +25,8 @@ The options are as follows:
 Specify a BIOS file to use during boot.
 If this flag is omitted,
 .Nm
-will use a high\(hylevel emulated BIOS.
+will use the BIOS specified in the configuration file,
+or a high\(hylevel emulated BIOS if none is specified.
 .It Fl l Ar loglevel
 Log messages during emulation.
 .Ar loglevel
--- a/doc/mgba-qt.6
+++ b/doc/mgba-qt.6
@@ -31,25 +31,25 @@ Log messages during emulation.
 .Ar loglevel
 is a bitmask defining which types of messages to log:
 .Bl -inset
-.It 0x01
+.It 1
 fatal errors
-.It 0x02
+.It 2
 errors
-.It 0x04
+.It 4
 warnings
-.It 0x08
+.It 8
 informative messages
-.It 0x10
+.It 16
 debugging messages
-.It 0x20
+.It 32
 stub messages
-.It 0x100
+.It 256
 in\(hygame errors
-.It 0x200
+.It 512
 software interrupts
-.It 0x400
+.It 1024
 emulator status messages
-.It 0x800
+.It 2048
 serial I/O messages
 .El
 The default is to log warnings, errors, fatal errors, and status messages.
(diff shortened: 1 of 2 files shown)


endrift:
It looks a bit bare, but it's probably a good start. I'll look into this a bit later, although it's worth mentioning that mgba(6) and mgba-qt(6) are different commands and have (slightly) different flags. It looks like the usage spew is out of date, too.

bentley:
Updated usage spew. What is dirmode/-D for?

endrift:
You didn't update the summary for -5/-6. Dirmode is for future usage when I add a library view. In regular mode, savegames, savestates and screenshots take the loaded filename and crops off the extension, then adds .sav, .ss1 or .png or whatever. In dirmode, you pass the name of a directory and it loads the first .gba file it finds, and drops whatever it feels like in that directory. Savegames are sram.sav, screenshots are screenshot.png, etc and savestates are savestate.ss1, etc. It's currently undocumented because there's no real end-user use case at the moment.

Similarly, -l is undocumented because the log levels aren't stable at the moment and may change in the future.

bentley:
Oops, rebased for -5/-6.

bentley:
Okay, noticed some keybindings I didn’t know about before:

* backslash to advance frame
* tab to accelerate

Are there any other keybindings I’ve missed?

Do the default keybindings differ between mgba and mgba-qt?

How do the command line flags differ between mgba and mgba-qt?

endrift:
F1–F9 load savestate, Shift-F1–F9 save savestate. In the SDL version, [ rewinds, but it's ` in the Qt version. I'm going to change the SDL version to ` soon to match, though. The Qt version doesn't have \ for frame advance (it's Ctrl-N in the Qt version, but that also works in the SDL version). There are some other miscellaneous things, too.

Some of the command line flags, e.g. -d, don't work in mgba-qt, and others I just haven't implemented yet. In general, most of them should be shared.

endrift:
We want to get an updated version of this in for 0.2.2 so that it can be pushed to Debian/Ubuntu package repos. We should chat in IRC about the right things to add.

endrift:
Not sure this is going to be ready in time for 0.2.2 :(

bentley:
I’ve added some more options to the manual. Now everything in the usage output for SDL is in there.

Is there anything missing that should be in there? And what command‐line options does mgba-qt take?

endrift:
I'll take a closer look tomorrow, but it's looking better so far. I'm not sure which of the options that mgba-qt takes actually *work* so I need to double check that. Also, the official site is mgba.io, which will cease being a redirect around the time I release 0.3.0, and the bugzilla listed is deprecated, so that might want to be fixed.

bentley:
I think the manuals are fairly complete at this point. Tweaks can continue to be made but these are a good baseline.

I don’t know CMake at all, so I’ll leave that part to you. The manuals need to be installed to `${mandir}/man6`. `mandir` should default to one of ${prefix}/share/man or ${prefix}/man (the former is what Debian uses)—whatever the default, `mandir` itself should be customizable by whoever’s building the software.

doc/mgba-qt.6
```diff
@@ -0,0 +1,97 @@
...
+.Bl -tag -width Ds
+.It Fl b Ar biosfile , Fl -bios Ar biosfile
+Specify a BIOS file to use during boot.
+If this flag is omitted,
+.Nm
+will use a high\(hylevel emulated BIOS.
+.It Fl l Ar loglevel
+Log messages during emulation.
+.Ar loglevel
+is a bitmask defining which types of messages to log:
```
endrift: These need to be specified in decimal. The parsing doesn't support hexadecimal at this time.

doc/mgba-qt.6
```diff
@@ -0,0 +1,97 @@
...
+.It 0x02
+errors
+.It 0x04
+warnings
+.It 0x08
+informative messages
+.It 0x10
+debugging messages
+.It 0x20
+stub messages
```
endrift: This should probably be rephrased to better explain what a stub is.

doc/mgba-qt.6
```diff
@@ -0,0 +1,97 @@
...
+.It 0x200
+software interrupts
+.It 0x400
+emulator status messages
+.It 0x800
+serial I/O messages
+.El
+The default is to log warnings, errors, fatal errors, and status messages.
+.It Fl p Ar patchfile , Fl -patch Ar patchfile
+Specify a patch file in IPS or UPS format.
```
endrift: +BPS

doc/mgba-qt.6
```diff
@@ -0,0 +1,97 @@
...
+.Sh DESCRIPTION
+.Nm
+is a Game Boy Advance emulator.
+The options are as follows:
+.Bl -tag -width Ds
+.It Fl b Ar biosfile , Fl -bios Ar biosfile
+Specify a BIOS file to use during boot.
+If this flag is omitted,
+.Nm
+will use a high\(hylevel emulated BIOS.
```
endrift: The default depends on the stored config. What's described is the default if no stored config is present.

bentley:
Updated based on your comments. Also mentioned the configuration files I see under `~/.config/mgba`; any other files to worry about?

endrift:
Only other file is portable.ini, which can be created in the current working directory to tell it to load the config files from the current working directory instead of the mentioned directory. Weird that I didn't use `XDG_CONFIG_HOME`. I should fix that.

doc/mgba.6
```diff
@@ -0,0 +1,241 @@
...
+.It Cm b Ns Oo Cm reak Oc Ns Cm /t Ar address
+Set a breakpoint \(en ARM
+.Pq Ql /a ,
+Thumb
+.Pq Ql /t ,
+or the current CPU mode \(en at
+.Ar address .
+.It Cm c Ns Op Cm ontinue
+Continue execution.
+.It Cm d Ns Oo elete Oc Ar address
```
endrift: The formatting on `elete` is different from the rest of the commands.

doc/mgba.6
```diff
@@ -0,0 +1,241 @@
...
+Delete a breakpoint at
+.Ar address .
+.It Cm dis Ns Oo Cm asm Oc Ar address Op Ar count
+.It Cm dis Ns Oo Cm asm Oc Ns Cm /a Ar address Op Ar count
+.It Cm dis Ns Oo Cm asm Oc Ns Cm /t Ar address Op Ar count
+.It Cm dis Ns Oo Cm assemble Oc Ar address Op Ar count
+.It Cm dis Ns Oo Cm assemble Oc Ns Cm /a Ar address Op Ar count
+.It Cm dis Ns Oo Cm assemble Oc Ns Cm /t Ar address Op Ar count
```
endrift: The `address` expression is optional--it will disassemble the current instruction if omitted.

doc/mgba.6
```diff
@@ -0,0 +1,241 @@
...
+or the current CPU mode.
+If
+.Ar count
+is not specified, only disassemble the instruction at
+.Ar address .
+.It Cm h Ns Op Cm elp
+Print help.
+.It Cm i Ns Op Cm nfo
+.It Cm status
+Print the current contents of general\(hypurpose registers.
```
endrift: Also the current program state register, and disassembling the current instruction.

doc/mgba.6
```diff
@@ -0,0 +1,241 @@
...
+or word
+.Pq Ql /4
+from
+.Ar address .
+.It Cm w Ns Oo Cm atch Oc Ar address
+Set a watchpoint at
+.Ar address .
+.It Cm w/1 Ar address
+.It Cm w/2 Ar address
+.It Cm w/4 Ar address
```
endrift: These all need a second argument, the data to be written.

doc/mgba.6
```diff
@@ -0,0 +1,241 @@
...
+or word
+.Pq Ql /4
+to
+.Ar address .
+.It Cm w/r Ar register
+Write a word to
+.Ar register .
+.It Cm x/1 Ar address
+.It Cm x/2 Ar address
+.It Cm x/4 Ar address
```
endrift: These can take a `count` argument for the number of bytes/halfwords/words.

doc/mgba.6
```diff
@@ -0,0 +1,241 @@
...
+Examine bytes
+.Pq Ql /1 ,
+halfwords
+.Pq Ql /2 ,
+or words
+.Pq Ql /4
+from
+.Ar address .
+.It Cm \&!\ \&
+Break into the attached debugger.
```
endrift: This phrasing could probably be clarified a bit about how it breaks into the debugger attached to mGBA (if present), not to mGBA's built-in debugger.

bentley:
Updated based on your comments, except the last one. Not sure how `!` works exactly. Would “Break into a gdb(1) session previously started by -g” work? Or is that not correct?

endrift:
That is incorrect. `!` is for a debugger external to mGBA, e.g. If running mGBA inside gdb, `!` raises a SIGTRAP to trap into gdb. I should probably disable it in release builds, actually.

doc/mgba.6
```diff
@@ -0,0 +1,259 @@
...
+.Ar data
+as a byte
+.Pq Ql /1 ,
+halfword
+.Pq Ql /2 ,
+or word
+.Pq Ql /4
+to
+.Ar address .
+.It Cm w/r Ar register
```
endrift: This also requires an argument for the data to be written into the register

doc/mgba.6
```diff
@@ -0,0 +1,259 @@
...
+halfwords
+.Pq Ql /2 ,
+or words
+.Pq Ql /4
+from
+.Ar address .
+If
+.Ar count
+is not specified, examine 16 bytes, 8 halfwords, or 4 words.
+.It Cm \&!\ \&
```
endrift: I've disabled this command in release builds, as it's not useful without debugging symbols, and is only useful for debugging mGBA itself, so this should probably be removed from the man page.

bentley:
Updated based on your comments.

karasuhebi:
Yay merged! :D

PR #5, opened 2015-03-12
Add additional overrides

clobber:
These are the ones I could verify so far that fix game loading and save issues. There could still be others needed from VBA's vba-over.ini that have problems due to AP further in-game that require more effort to verify.


src/gba/supervisor/overrides.c
```diff
@@ -28,6 +35,9 @@ static const struct GBACartridgeOverride _overrides[] = {
 	// Final Fantasy Tactics Advance
 	{ "AFXE", SAVEDATA_FLASH512, HW_NONE, 0x8000428 },
 
+	// F-Zero - Climax
+	{ "BFTJ", SAVEDATA_FLASH1M, HW_NONE, IDLE_LOOP_NONE },
```
endrift: This seems to be detected fine for me even without the override.

endrift:
Everything else looks good. If you make a new version of this to remove that one line, or if you find an issue with it on your end that mandates it, please put "GBA: " at the beginning of the commit message to indicate that this is a GBA-specific commit.

PR #9, opened 2015-04-14
Add an About Dialog

megumisonoda:
Tested and working fine on Arch Linux. I put 'Help' in the menubar because that's where About Pages usually sit, and I couldn't think of a better place to stick it. I thought that perhaps we could add a link to the support forums under Help as well, but that's something for another commit.


src/platform/qt/AboutDialog.h
```diff
@@ -0,0 +1,22 @@
+/* Copyright (c) 2013-2015 Jeffrey Pfau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include <QDialog>
```
endrift: Needs header guards.

src/platform/qt/AboutDialog.h
```diff
@@ -0,0 +1,22 @@
...
+Q_OBJECT
+
+public:
+	AboutDialog(QDialog* parent = nullptr);
+
+private:
+	Ui::AboutDialog m_ui;
+};
+
+}
```
endrift: Make sure there's a newline at the end of the file; some editors get angry if it's missing.

endrift:
Not sure if this is what it was supposed to look like, but on OS X it looks a bit...bad. The text is a bit verbose for an about dialog (what user cares what it's written in when they're looking at an about dialog?) and it could use the logo on the side (which is already present as `:/res/logo.png` IIRC).
OS X

Apart from two nitpicks, the code itself looks good. I'd recommend taking a look at [Dolphin PR 2016](https://github.com/dolphin-emu/dolphin/pull/2016) for a bit of insight into how an about dialog might be designed.

megumisonoda:
I updated the design based on the Dolphin PR, fixed the nitpicks, as well as making the window fixed-size. You should test it again, because the fonts & font rendering on my Linux install right now are garbage.

endrift:
Well, the design is certainly a lot better, but I'm really not fond of how it's implemented.

* The dialog should be using layouts throughout instead of many fixed-position elements.
* The logo should be a QLabel with the pixmap set to `:/res/mgba-1024.png`, with scaledContents enabled. (This enables the same logo to be reused without having to add filesize to the project by adding a resized version of the logo.) Doing something like this is good:

        m_ui.logo->setDevicePixelRatio(devicePixelRatio());

    This allows it to be displayed as higher than 1x DPI on HiDPI devices.
* The project name and version number should be separate QLabels, and set dynamically at runtime to be `m_ui.projectName->setText(PROJECT_NAME);` and `m_ui.projectVersion->setText(PROJECT_VERSION);`. This allows the name and version number to be specified at compile time instead of being hardcoded.
* Using QLabels with HTML isn't really a good approach. One QLabel per paragraph, with wordwrap enabled, and a tiny bit of styling, is probably a better approach, especially once layouts are used.
* The copyright text is a size or two too small.
* The links don't do anything, at least on OS X. Not sure why this is. Not even use having links there is useful, but it may be. Also, the license link should go to the LICENSE file in the repo, just in case it's ever relicensed.

Overall though, this is much better. Thanks! I have a few more inline comments I'll add in a second.

src/platform/qt/AboutDialog.cpp
```diff
@@ -0,0 +1,18 @@
+/* Copyright (c) 2013-2015 Jeffrey Pfau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include "AboutDialog.h"
+
+#include <QDesktopServices>
+#include <QUrl>
```
endrift: Are these includes needed?

src/platform/qt/AboutDialog.h
```diff
@@ -0,0 +1,25 @@
...
+#define QGBA_ABOUT_DIALOG
+#include "ui_AboutDialog.h"
+
+namespace QGBA {
+
+class AboutDialog : public QDialog {
+Q_OBJECT
+
+public:
+	AboutDialog(QDialog* parent = nullptr);
```
endrift: `parent` should be a QWidget, not a QDialog.

src/platform/qt/AboutDialog.h
endrift: All includes should be after the header guards.

src/platform/qt/AboutDialog.cpp
megumisonoda: QDesktopServices and QUrl allow opening the user's default web browser upon clicking the links.

megumisonoda:
I'll work on it more then, thanks for the input!

src/platform/qt/AboutDialog.cpp
endrift: Presumably you need to call something in those files instead of just including them, though?

megumisonoda: Hmm, it looks like you're right. Does it open in the expected web browser for you as-is? If it does, I'll just remove the includes. Otherwise, I'll implement it properly.

megumisonoda: Hm, QLabel doesn't have DevicePixelRatio as far as I can tell. Should I be using something else for the logo?

endrift: Oh sorry, try doing that to `m_ui.logo->pixmap()->setDevicePixelRatio()`—it's a property of the pixmap.

megumisonoda: That still doesn't work because QPixmap is constant, and the setDevicePixelRatio method isn't.

endrift: Ugh, ok, just make the QPixmap in the constructor, call setDevicePixelRatio, and then setPixmap on the QLabel, then.

megumisonoda:
I pushed another update, no device pixel ratio changes quite yet, but everything else (including layout changes) is here.

src/platform/qt/Window.cpp
```diff
@@ -929,6 +935,13 @@ void Window::setupMenu(QMenuBar* menubar) {
 		m_inputController.setAllowOpposing(value.toBool());
 	}, this);
 
+	QMenu* helpMenu = menubar->addMenu(tr("&Help"));
+	m_shortcutController->addMenu(helpMenu);
+	QAction* about = new QAction(tr("&About mGBA"), helpMenu);
+	connect(about, SIGNAL(triggered()), this, SLOT(openAboutDialog()));
+	addControlledAction(helpMenu, about, "aboutDialog");
+
+
```
endrift: Too many newlines

src/platform/qt/Window.cpp
```diff
@@ -929,6 +935,13 @@ void Window::setupMenu(QMenuBar* menubar) {
 		m_inputController.setAllowOpposing(value.toBool());
 	}, this);
 
+	QMenu* helpMenu = menubar->addMenu(tr("&Help"));
+	m_shortcutController->addMenu(helpMenu);
+	QAction* about = new QAction(tr("&About mGBA"), helpMenu);
```
endrift: tr("&About %1").args(PROJECT_NAME)

src/platform/qt/AboutDialog.h
```diff
@@ -0,0 +1,25 @@
+/* Copyright (c) 2013-2015 Jeffrey Pfau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#ifndef QBGA_ABOUT_DIALOG
```
endrift: Header guard is spelled wrong

endrift:
Much better now. Apart from some really small nitpicks now, it's basically done. Note that it was fine being a QDialog before (and it probably should still be), it was the `parent` parameter that was the wrong type. Also, please set the layout for the root to be a vertical layout—it's absent at the moment, but that's really the only major problem with it now. There's a bit much spacing on the copyright footer, though, but that's not a big problem.

src/platform/qt/AboutDialog.ui
```diff
@@ -0,0 +1,217 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>AboutDialog</class>
+ <widget class="QWidget" name="AboutDialog">
+  <property name="windowModality">
+   <enum>Qt::WindowModal</enum>
```
endrift: The window should not be modal. I couldn't even quit the program until dismissing it thanks to that!

megumisonoda: My mistake, I misunderstood what the option was for.

megumisonoda:
Updated again. It's still a QWidget, is that alright? If not I can change it back to a QDialog. It'll just take a bit more time. ^^;

endrift:
The vertical layout for the root is not applied properly. You placed as a child of the widget instead of as the layout for the widget. You need to actually click on the widget, then press the vertical layout button in the toolbar to get it to be the layout for the widget.

You also did not address using `QString::args(PROJECT_NAME)` instead of hardcoding the name mGBA anywhere.

megumisonoda:
Ah, my mistake. I'll fix it later tonight then, as I have some errands to run.

endrift:
Any progress on this recently?

karasuhebi:
bump

karasuhebi:
Please be aware that at the 1x frame size, the Help submenu entry would not fit and would probably either be cut off or be pushed to a second line. Here's a screenshot of how it looks now:
http://i.imgur.com/GwsWqtL.png

endrift:
Bump. Gonna be writing this up myself soon if this PR isn't done in the next week or two.

endrift:
Per a twitter conversation with @megumisonoda, closing this PR. Let me know if you want to reopen if before I get to it myself.

PR #21, opened 2015-05-12
fix build issues on ubuntu

ikeboy:
This fixes https://github.com/mgba-emu/mgba/issues/18 for me.

### DIFF
--- a/src/platform/qt/AudioProcessorSDL.h
+++ b/src/platform/qt/AudioProcessorSDL.h
@@ -16,7 +16,6 @@ extern "C" {
 namespace QGBA {
 
 class AudioProcessorSDL : public AudioProcessor {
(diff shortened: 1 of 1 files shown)


endrift:
This fix is not correct. Q_OBJECT is necessary for any object that has slots, as this does.

ikeboy:
I'm not sure what exactly the problem is, but removing that line lets it build and run fine, while there are errors with it.

endrift:
That's a coincidence. The real problem appears to be that the file is being built without `BUILD_SDL` being defined. I'm assuming that with this being built as it is, the SDL option does not show up in Settings > Audio driver?

ikeboy:
>SDL option does not show up in Settings > Audio driver?

It does, actually.

PR #34, opened 2015-06-05
Qt: Rename rewind variables

tony971:
Slightly more intuitive in my opinion


src/platform/qt/SettingsView.ui
```diff
@@ -327,7 +327,7 @@
        <item row="5" column="0">
         <widget class="QLabel" name="label_4">
          <property name="text">
-          <string>Rewind interval:</string>
+          <string>Rewind snapshot:</string>
```
endrift: I have no idea what this is supposed to mean. What would a snapshot value mean? This is the number of frames between rewind states.

PR #40, opened 2015-06-11
All: initial work on MSVC support.

waddlesplash:
The only reason this fails now is because of the macro expansion stuff.


CMakeLists.txt
```diff
@@ -1,7 +1,12 @@
 cmake_minimum_required(VERSION 2.6)
 project(mGBA C)
 set(BINARY_NAME mgba CACHE INTERNAL "Name of output binaries")
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -std=c99")
+if(NOT MSVC)
+	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -std=c99")
+else()
+	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Dsnprintf=_snprintf -Dinline=__inline")
```
endrift: Please do this in src/util/common.h, as there are already some defines like this in there.

src/util/common.h
```diff
@@ -20,6 +20,8 @@
 #include <string.h>
 
 #ifdef _MSC_VER
+#define snprintf _snprintf
+#define inline __inline
```
endrift: I'd put these a bit lower for consistency's sake.

endrift:
Please rebase this. I removed a ton of code from the thumb instruction table that will hopefully fix most if not all of the MSVC macro issues.

waddlesplash:
@endrift OK, rebased. The build now halts because MSVC does not have `getopt`.

PR #57, opened 2015-07-02
Qt: (GameController) Add a missing va_end call to the log handler lambda

lioncash:


endrift:
I was never sure if `va_copy` needed to be matched with `va_end`, but it looks like cplusplus.com corroborates.

PR #62, opened 2015-07-19
The library directory is set to "lib" or "lib64" depending on the system.

duganchen:
This should better accommodate systems (such as Slackware, which I run) that expect 64-bit libraries to be in /usr/lib64.

It allows GNUInstallDirs to set the CMAKE_INSTALL_LIBDIR variable, then uses that variable in every place where it would help (for example. in places previously hardcoded to "lib").

I've tested it on my 64-bit system, and it results in the library directory being set correctly to /usr/local/lib64 if you don't specify any CMake options.

I also tested using -DCMAKE_INSTALL_LIBDIR=lib128 to make sure that overriding it works, and it does.


endrift:
Unfortunately, not all distros use lib64 (e.g. Ubuntu, Arch, etc.), and this seems to have CMake set it to lib64 regardless of the distro. I can't merge this if it'll break other distros instead of just fixing lib64 distros. You might want to file a bug with CMake upstream.

duganchen:
I've just pushed a new revision. The library directory now defaults "lib" (even on my 64-bit Slackware system; I just tested) unless you override it with -DLIBDIR=lib64.

endrift:
Sounds good, but please squash the commits before I merge this.

duganchen:
I just squashed the commits and did a forced push.

PR #63, opened 2015-07-21
Review PR

fr500:


src/gba/supervisor/config.c
```diff
@@ -166,7 +166,7 @@ void GBAConfigDirectory(char* out, size_t outLength) {
 	char* home = getenv("HOME");
 	snprintf(out, outLength, "%s/.config", home);
 	mkdir(out, 0755);
-	snprintf(out, outLength, "%s/.config/%s", home, binaryName);
+        snprintf(out, outLength, "%s/.config/%s", home, binaryName);
```
endrift: Might want to fix this so the indentation matches. That would remove this from the diff entirely.

src/platform/posix/memory.c
```diff
@@ -7,6 +7,14 @@
...
 #include <sys/mman.h>
 
+#ifndef MAP_ANONYMOUS
+#define MAP_ANONYMOUS 0x20
+#endif
+
+#ifndef MAP_ANON
+#define MAP_ANON MAP_ANONYMOUS
+#endif
+
```
endrift: On what platforms is this needed?

src/platform/libretro/memory.c
```diff
@@ -0,0 +1,38 @@
...
+
+void mappedMemoryFree(void* memory, size_t size) {
+#ifdef _WIN32
+	UNUSED(size);
+	// size is not useful here because we're freeing the memory, not decommitting it
+	VirtualFree(memory, 0, MEM_RELEASE);
+#else
+	munmap(memory, size);
+#endif
+}
```
endrift: All of this code already exists in other places. I can vaguely understand the desire to have it all in one file and then have the preprocessor take care of it, but it's a needless duplication of code. At worst, you can use the preprocessor to include the individual files (e.g.

    #ifdef _WIN32
    #include "platform/windows/memory.c"
    #else
    #include "platform/posix/memory.c"
    #endif

...but there's a reason these directories exist in the first place. Since the goal is to support as many platforms as possible, you should probably just include each file in the Makefile as needed, as determined by the compilation target.

Makefile.common
```diff
@@ -0,0 +1,54 @@
...
+					$(CORE_DIR)/gba/cheats/parv3.c \
+					$(CORE_DIR)/gba/cheats/codebreaker.c \
+					$(CORE_DIR)/gba/gba.c \
+					$(CORE_DIR)/gba/hardware.c \
+					$(CORE_DIR)/gba/sio.c \
+					$(CORE_DIR)/gba/sio/lockstep.c \
+					$(CORE_DIR)/gba/video.c \
+					$(CORE_DIR)/gba/serialize.c \
+					$(CORE_DIR)/gba/supervisor/rr.c \
+					$(CORE_DIR)/gba/supervisor/cli.c \
```
endrift: Are cli.c and rr.c actually necessary? I need to make it possible to compile without those two files if that's not the case.

Makefile.common
```diff
@@ -0,0 +1,54 @@
...
+					$(CORE_DIR)/arm/isa-arm.c \
+					$(CORE_DIR)/arm/decoder-arm.c \
+					$(CORE_DIR)/arm/decoder-thumb.c \
+					$(CORE_DIR)/arm/decoder.c \
+					$(CORE_DIR)/debugger/debugger.c \
+					$(CORE_DIR)/debugger/memory-debugger.c \
+					$(CORE_DIR)/gba/savedata.c \
+					$(CORE_DIR)/gba/io.c \
+					$(CORE_DIR)/gba/rr/vbm.c \
+					$(CORE_DIR)/gba/rr/mgm.c \
```
endrift: The rr folder should absolutely not be necessary.

Makefile.common
```diff
@@ -0,0 +1,54 @@
+INCFLAGS := -I$(CORE_DIR) -I$(CORE_DIR)/arm
+
+SOURCES_C :=   $(CORE_DIR)/arm/arm.c \
+					$(CORE_DIR)/arm/isa-thumb.c \
+					$(CORE_DIR)/arm/isa-arm.c \
+					$(CORE_DIR)/arm/decoder-arm.c \
+					$(CORE_DIR)/arm/decoder-thumb.c \
+					$(CORE_DIR)/arm/decoder.c \
+					$(CORE_DIR)/debugger/debugger.c \
+					$(CORE_DIR)/debugger/memory-debugger.c \
```
endrift: I'm not sure if the debugger folder is necessary. If it is, I should fix it to not be.

Makefile.common
```diff
@@ -0,0 +1,54 @@
...
+					$(CORE_DIR)/gba/audio.c \
+					$(CORE_DIR)/gba/memory.c \
+					$(CORE_DIR)/gba/cheats.c \
+					$(CORE_DIR)/gba/cheats/gameshark.c \
+					$(CORE_DIR)/gba/cheats/parv3.c \
+					$(CORE_DIR)/gba/cheats/codebreaker.c \
+					$(CORE_DIR)/gba/gba.c \
+					$(CORE_DIR)/gba/hardware.c \
+					$(CORE_DIR)/gba/sio.c \
+					$(CORE_DIR)/gba/sio/lockstep.c \
```
endrift: The lockstep driver isn't useful unless there's a way to run multiple cores in the same libretro process, and I know it can be compiled without it.

Makefile.common
```diff
@@ -0,0 +1,54 @@
...
+					$(CORE_DIR)/util/table.c \
+					$(CORE_DIR)/util/vfs/vfs-mem.c \
+					$(CORE_DIR)/util/string.c \
+					$(CORE_DIR)/util/hash.c \
+					$(CORE_DIR)/util/configuration.c \
+					$(CORE_DIR)/util/crc32.c
+					
+SOURCES_C += $(CORE_DIR)/util/circle-buffer.c
+#SOURCES_C += $(CORE_DIR)/debugger/gdb-stub.c
+SOURCES_C  += $(CORE_DIR)/third-party/blip_buf/blip_buf.c
```
endrift: Why are these on separate lines?

src/platform/libretro/memory.c
twinaphex: I'd rather libretro be treated as its own platform for this 'memory' subsection and for us just to supply solutions per platform instead of having to deal with separate platform memory implementation files like this. I think this is really not a major point of contention and that you should allow us some leeway on this. There's going to be lots of platforms we cater to that you don't and there's no reason to introduce codebase bloat and put them all in separate files like this when a lot of lines of code are just going to be identical with some minor variations.

src/gba/supervisor/config.c
twinaphex: I'll check it out.

Makefile.common
twinaphex: Omitted now.

Makefile.common
twinaphex: Omitted.

Makefile.common
twinaphex: Omitted.

Makefile.common
twinaphex: Took care of that now.

twinaphex:
Some other portability issues right now in the codebase that will have to get addressed upstream -

https://github.com/mgba-emu/mgba/blob/master/src/util/common.h#L34

The PS3 GCC toolchain doesn't have strings.h, usually in codebases I just remove this as it doesn't seem to serve any benefit and most of the prototypes/defines are catered to by other headers anyway. I'll let you make the decision here I guess. For now I just removed it in the commits.

Also, you might not be a fan of the struct/union names I chose in commit revision 9b4ef89, however, the main objective was just to get it working. You can come up with something better if you intend to but do keep in mind that anon unions/structs will have to go if we care about OSX PowerPC toolchain compatibility, and I think it'd be foolhardy not to since it opens the emulator codebase up to even more platforms.

twinaphex:
OK, I see you did this for the Wii which is more or less what I was thinking of doing anyway I guess -

https://github.com/mgba-emu/mgba/blob/4f04c30f708f30f51f5b90aff2e320cd607955ea/src/platform/wii/memory.c

Makefile.common
twinaphex: debugger files do seem to be a hard dependency yeah.

endrift:
Have you tried `-fm a-extensions` on gcc 4.0? I used to be able to get it compiling for PPC Mac just fine on gcc 4.2, and only recently stopped supporting it. I mean honestly, who still runs PPC Mac? Furthermore, PPC is not fully supported yet due to a handful of issues with save states and some cartridge hardware. It's premature to add support for PPC platforms, as I still need to add those bugs.

In no places do I allocate executable memory. There is no dynarec at the moment.

I don't have a PS3 that can run homebrew but I guess I can download a tool chain and try to compile it, fixing things without taking a hammer to something that isn't actually a nail.

src/platform/libretro/memory.c
endrift: Most of the "common lines of code" are braces or the function definition header. As it turns out, you have more lines of code in common if you have the distinct file due to repeating actual code instead of just punctuation.

twinaphex:
>  I used to be able to get it compiling for PPC Mac just fine on gcc 4.2, and only recently stopped supporting it. I mean honestly, who still runs PPC Mac? 

I like to be able to support those niches though, and I want to make it a point for them to be able to have most of the codebases compilable using nothing more than the base SDK/tooling that is already available on the platform. So yeah, if this is an issue where you don't want to follow suit, I will be forced to diverge from it in my own fork. Hopefully we can come to a suitable compromise where we just give structs/unions always a name, it's a relatively little thing to do really and the benefits outweigh the negatives.

BTW, I did get mgba running on RetroArch Wii today. Problem is, it's much slower than VBA Next. Not sure if those idle speedhacks will give it the edge over VBA Next but I find it quite worrying that we even have to resort to that given it's at least half of fullspeed right now on games like Super Mario Advance and Advance Wars (around 35-38fps) vs. well over 70-80 on VBA Next (VBA being pretty far from being an optimized emulator itself). Coincidentally, on the PowerPC Mac (Powerbook G4), I got 90/100fps on Super Mario Advance vs. the Wii's 38fps, so there's something to be said for those old PowerPC Macs that apparently nobody uses vs. a Wii.

twinaphex:
> I don't have a PS3 that can run homebrew but I guess I can download a tool chain and try to compile it, fixing things without taking a hammer to something that isn't actually a nail.

Umm, what? I already got it compiled and running on PS3. What about this in any way is about "taking a hammer to something that isn't actually a nail"? Sorry but this is getting rather annoying now, by your responses it seems like you don't respect or appreciate what I'm doing or something and think it's 'inferior'.

endrift:
I expect you to not remove header includes entirely since it is necessary on at least one platform. I expect you to ifdef them out. That's really the only problem I can find for the PS3 part.

I also expect you to not manhandle code for an infinitesimal payoff. Anyone who is still using 10.5 is going to be used to not being able to run anything newer than 5 years ago. If you use a slightly newer tool chain (4.2 comes to mind, which I used for quite a while) and distribute it without overhauling a large part of my code, cool. And it seems to me that not all of your supported targets (PS3?) you expect users to be able to compile for, so I don't know why you care.

Makefile.common
endrift: I'll see about fixing that.

twinaphex:
>I also expect you to not manhandle code for an infinitesimal payoff. 

LOL, and since when do you get to dictate that? You want to be *less* backwards compatible for no damn reason? OK, you've officially lost me then. I'll just do the stuff on my fork and do it properly in my mind, and you can do it *properly* on upstream according to whatever you believe it should be done. Makes no real difference to me. I have no desire to be engaging in all sorts of ego feuds over this stuff, either the contributions are welcome or they're not, and I guess you've already made that decision there.

endrift:
I see now why you have a reputation for creating drama. I tried to cooperate, but you made large changes with flimsy justification (spoilers, performance is irrelevant if there's no userbase). You are entitled and hardheaded. I've talked to a lot of people who aren't fond of how you hard fork projects. You're welcome to keep having your fans brow nose up to your ego, but this is my code and I actually can dictate what goes upstream. I wanted to avoid this, but your contributions are getting too heavy handed for a very small payoff. If this is why you want to fork it, over something so small and petty, so be it. I can't stop you. But you're making a foolish decision.

Good day.

twinaphex:
OK, so since the previous attempt to merge in stuff you didn't approve of met with resistance, can we try this again please?

This time it'd be limited to things you agree on and we'll do it your way. I'll have one shallow fork where we leave things unaltered and we just push stuff to you that is of direct benefit to the libretro port and I'll have another unshallow fork where we'll have certain compatibility patches I guess for things like PowerPC/PS3 compilers which you don't care about.

Is this something you'd want to give a chance? I'm sorry about how things previously fell through but you have to understand that I just have a certain passion towards making certain things work on these old platforms, probably the same reason why you have certain things you don't feel comfortable merging anyway. We all have this. I just thought this would be a good idea for your project as well since it'd require minimal changes, it's apparent you disagreed, and we fell out over it. Anyway, I'd like for us to work together instead of to have some split like this which benefits nobody. Let me know. Perhaps I previously went about this the wrong way and if so I apologize.

endrift:
Assuming cool heads can be kept, I'm more than willing to continue to work on this together. For the nonshallow fork I may suggest a branch instead of a separate repo, but that's an implementation detail.

My biggest problems with the changes made in the fork were twofold:

- Changes were made to support other platforms that broke existing ones. This prevents me from upstreaming it one way or another. This was mostly the removal of various header includes which could be ifdef'd out.
- Decreasing readability in central parts of the code base just to make it compile with a 10 year old compiler. I am not inherently opposed to supporting these operating systems, but if it means kludging up the code just to support an ancient compiler, I'm not happy.

Everything else was rather minor and I think could be resolved easily with more communication and collaboration. I definitely have no interest in screwing over users, but I'd imagine that there are no people using PPC Macs as their emulation machines. I realized that you meant that this also applied to the PS3 compiler, which is unfortunate and I guess a good reason for someone other than me to maintain a more-compatible version. I'm not fond of it being in master, but perhaps a branch that's merged often is a good idea.

I am absolutely interested in supporting the Wii, since a lot of people use that for emulation, and my early experiments seemed to show it running full speed in games. Although I didn't have sound working, it definitely seemed faster than what you're saying it is. I'll need to look further into that soon. PPC full support is an open ticket on the bugzilla and has been for ages. The biggest problem right now is the lack of compatibility between big endian savestates and little endian savestates (BE cannot read or write LE states at the moment, and I'd like all savestates to be LE), so I'd like to resolve that before something that can make savestates is released.

twinaphex:
OK, thanks for this response. I really appreciate it. I will try to get back to radius and we'll see how we can go about pushing a PR that is a bit less of a mess.

PR #70, opened 2015-07-27
Add debian packaging.

sergiobenrocha2:
To make a deb package, you'll need:

mv tools/debian/ .
debuild -b -uc -us

You need to install devscripts and build-essential, as well all packages listed in the debian/control file.


tools/debian/control
```diff
@@ -0,0 +1,78 @@
...
+               libqt5opengl5-dev,
+               libsdl2-dev,
+               libswscale-dev,
+               libzip-dev,
+               pkg-config,
+               qtbase5-dev,
+               qtmultimedia5-dev,
+               zlib1g-dev
+Standards-Version: 3.9.6
+Homepage: https://endrift.com/mgba/
```
endrift: mgba.io will probably be the domain name as soon as I release 0.3.0

tools/debian/control
```diff
@@ -0,0 +1,78 @@
...
+
+Package: libmgba
+Architecture: any
+Multi-Arch: same
+Depends: ${misc:Depends}, ${shlibs:Depends}
```
endrift: I'm a bit wary of this, since it will probably pull in libavcodec, when libavcodec-extra is preferable. However, either of them can be used.

sergiobenrocha2: There's no "libavcodec-extra-dev" package, it's up to the user install the libavcodec-extra

endrift: I'm aware; this comment was about th shlibs detection line for libmgba, which I was afraid would set a hard dependency on libavcodec and not allow -extra.

sergiobenrocha2: No problem, take a look:

libavcodec54 (>= 6:9.1-1) | libavcodec-extra-54 (>= 6:9.18)

This is from the deb package.

PR #79, opened 2015-08-08
Qt: Really fix forums link

bentley:


endrift:
Er, oops, that's what I get for editing in a laggy editor hastily. Thanks for catching that.

PR #90, opened 2015-08-18
Qt: Fix installation path of XDG desktop file

bentley:
Without this, the desktop file gets installed outside the DESTDIR prefix.


endrift:
This wasn't the right way to fix the bug, so I went ahead and fixed it properly in 1f10613d1b7e96d362660c81c37ccf78c2276f7d.

PR #93, opened 2015-08-20
Port/wii, added wiimote + classic controller support and windows makefile for devkitppc

askotx:


endrift:
There are three big problems here, as far as I can tell:

- You merged instead of rebasing. Please rebase instead.
- You're trying to do too many things in one pull request. Please split it up into separate PRs for the individual bits that are needed.
- You're doing things that I don't really want to merge, for different reasons per bit. I can split those out if you want.

I can't merge this as-is due to these three things listed.

askotx:
Ok, I will, but I'm no expert on git, but I'll try, first I'm going to close this pull request, but what can I pull?

endrift:
What you should do is create a branch based on the upstream port/wii branch for each separate thing you want to file a PR for, and make sure not to merge in other branches, as that will pull in the merge commits into the PR.

Unfortunately, without some git wizardry, it'll be difficult to do anything if you can't build it, and your "build on Windows" fixes should definitely be a distinct PR. I'd highly suggest using the official CMake steps, as that removes the need for redundant build systems. If you pop into #mgba on FreeNode IRC, I can help you get that working, but I'm going to bed shortly so it'll have to be some other time.

Issue #15, opened 2015-04-24
Video filters


ROMaster2:
[Enhancement]

Video filters like hq2x, hq3x, and 2xSaI would be nice when the resolution is larger, so it wouldn't look quite as pixelated.

Great job on everything else so far!

purpasmart96:
Those filters are kinda garbage as of 2015, xbrz or just plain xbr is in my opinion superior to those. but a plain bilinear filter or just some support for retroarch shaders would be nice.

CarterLi:
+1 for xbrz

endrift:
In bugzilla already: https://endrift.com/mgba/bugs/show_bug.cgi?id=192

ROMaster2:
While I'm certain it won't be a problem, I'll point it out since puNES doesn't support it (yet).

When enabling the video filters (like 2xBRZ, 3xBRZ), allow game/window scale be independent of it, at least in integer multiples. I like using x6 scale with a x2 or x3 video filter.

elcorsico:
how enable video filter ? just resample vide for me
add xBRZ and other please..

Zetchzie:
xBRZ tends to be very slow on low end Computers AdvanceMame2X filter or HQ3X is as high as I can go On my old Rig with VBA-M, ever since 2.0.1 Beta it's been a very demanding emulator

ROMaster2:
Well it would be optional. That, and once video filters are added in it'll be easier to add more.

DemonEyesBob:
From what I understand, right now mGBA runs (30%?) faster than VBA-M to begin with, so even more intensive filters wouldn't be as noticeable. Not to mention it doesn't make sense to not implement better filters just because they won't work on low end hardware. That's what having multiple options is for :)

Zetchzie:
Of course all hardware has its limits. As long as we have some good non demanding filters to choose from i'm happy lol xBRZ just tends to be very slow if you don't got a modern computer. No reason not to add them just hope we get some low end friendly Filters too.

endrift:
xBR added, but I won't be adding xBRZ as it can't be done on the GPU.

endrift:
Closing, since all of the major featurework is done.

inukaze:
Hi there, someone can tell me if this emulator have Video Filters / Shaders , like : lq /hq ( 2x 4x ) SuperEagle , 2xSai , Super2XSai , Bilinear , Bilinear Plus , AdvanceMAME 2X , xBRZ 2x ?

[...]

PR #100, opened 2015-08-23
gba filter video hq ??

elcorsico:
please add gba filter video xBRZ ,ect..
thanks


endrift:
This isn't how pull requests work. Video filters were filed in #15 four months ago, so you're a bit late to the party anyway.

PR #111, opened 2015-09-06
Use the 3ds-cmake for better compatibility

Lectem:
This helps to clean the CMakeLists.txt a bit, and fix the build for those under Windows.
This is based on https://github.com/Lectem/3ds-cmake


endrift:
Been thinking about this, and I think it'd be quite a bit overkill, and widely asymmetrical with the other platforms which have similar problems to solve but don't go above and beyond like this. It's a nice framework, but I don't think mGBA has much use for the cruft that comes along with it above what's already done (and was just done recently).

Sorry.

Lectem:
No problem, but would it be possible to at least fix the toolchain file for windows users?

endrift:
I did that last night:

Lectem:
Oh right, didn't see it

PR #137, opened 2015-10-14
software renderer: Silence -Wlogical-not-parentheses warnings

lioncash:
Silences a load of warnings when compiling with GCC 5.2.1


endrift:
I was tempted to just turn that warning off since it seems really silly to me, but I guess this works too.

PR #188, opened 2015-12-15
Unpause the emulator when resetting.

ghost:
mgba-qt hangs if you reset the emulator while it is paused.  Subsequently unpausing the emulator does not resolve the hang, although reloading the rom does.

The immediate cause of the hang is that the CPU thread is waiting for the audioRequiredCond condition in GBASyncProduceAudio.  This condition never gets raised because AudioProcessorSDL has paused the audio.  Because the thread's state was set to THREAD_RUNNING by the reset, none of the unpause handlers will get called.

I've solved this by unpausing the emulator before resetting it.

### DIFF
--- a/src/platform/qt/GameController.cpp
+++ b/src/platform/qt/GameController.cpp
@@ -474,6 +474,7 @@ void GameController::setPaused(bool paused) {
 }
 
 void GameController::reset() {
+	setPaused(false);
 	GBAThreadReset(&m_threadContext);
 }
 


snickerbockers:
(BTW this pull request is from me, the reason it says ghost is that I accidentally screwed up and made the request from an alternate account).

endrift:
I'm not particularly fond of this fix, as there are real reasons to start the emulator paused. There are probably better ways to fix this; I'll probably take a look at an alternate fix soon. Thanks for the debugging, though!

endrift:
This fix was requisite for my extended fix, so I merged it first.

PR #229, opened 2016-02-17
OpenGL: Correct boolean vector strcmp strings for uniforms

lioncash:


CHANGES
```diff
@@ -6,6 +6,7 @@ Bugfixes:
  - Qt: Fix sending gameStopped twice
  - Qt: Fix hang if audio sync is enabled and audio fails to initialize
  - GBA BIOS: Fix RegisterRamReset setting DISPCNT to the wrong value
+ - GLES2: Correct boolean vector strcmp strings for uniforms
```
endrift: This should be OpenGL, not GLES2. (Same for the commit message.)

PR #257, opened 2016-04-16
All: Working MSVC support. 

waddlesplash:


src/platform/qt/InputProfile.h
```diff
@@ -48,15 +48,9 @@ class InputProfile {
...
-	                       AxisValue axes[GBA_KEY_MAX] = (AxisValue[GBA_KEY_MAX]) {
-	                       	                             {}, {}, {}, {}, 
-	                                                     { GamepadAxisEvent::Direction::POSITIVE, 0 },
-	                                                     { GamepadAxisEvent::Direction::NEGATIVE, 0 },
-	                                                     { GamepadAxisEvent::Direction::NEGATIVE, 1 },
-	                                                     { GamepadAxisEvent::Direction::POSITIVE, 1 },
-	                                                     {}, {}},
+						   const ShortcutButton* shortcutButtons,
+						   const ShortcutAxis* shortcutAxes,
+						   AxisValue axes[GBA_KEY_MAX],
```
waddlesplash: I took this out because I was getting
> error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax

src/platform/qt/InputProfile.cpp
```diff
@@ -154,6 +155,7 @@ const InputProfile InputProfile::s_defaultMaps[] = {
 		}
 	},
 };
+#endif
```
waddlesplash: Same C4576 here

src/platform/qt/OverrideView.cpp
```diff
@@ -56,12 +56,12 @@ void OverrideView::saveOverride() {
...
 
 void OverrideView::updateOverrides() {
-	m_override = (GBACartridgeOverride) {
+	/*m_override = (GBACartridgeOverride) {
 		"",
 		static_cast<SavedataType>(m_ui.savetype->currentIndex() - 1),
 		HW_NO_OVERRIDE,
 		IDLE_LOOP_NONE
-	};
+	};*/
```
waddlesplash: And here

src/util/nointro.c
```diff
@@ -158,7 +158,8 @@ struct NoIntroDB* NoIntroDBLoad(struct VFile* vf) {
 	struct {
 		char* key;
 		struct NoIntroItem* item;
-	} keyStack[KEY_STACK_SIZE] = {};
+	} keyStack[KEY_STACK_SIZE];
+	//keyStack = {{}};
```
waddlesplash: I think this was a different error but now I can't recall

src/platform/msvc/dirent.h
```diff
@@ -0,0 +1,917 @@
+/*
```
endrift: Please write a VDir for Win32 instead of using a dirent shim. The whole point of VDir is to have platform-specific implementations instead of having to reimplement all of dirent each time.

src/platform/qt/PaletteView.h
```diff
@@ -13,8 +13,6 @@
 
 #include "ui_PaletteView.h"
 
-class GameController;
```
endrift: Why?

waddlesplash: Uh, you include GameController.h on line 11 so this is redundant and it causes a compiler error?

endrift:
Please rebase on top of 3a51c59768a0e4bbba9dc516b0554eb520474c05; it adds native VDir stuff so you don't need to worry about the dirent shim anymore.

waddlesplash:
Sorry, this week has been madness for me. It looks like you already merged most of this? I'll get the rest over the weekend.

endrift:
I only merged the removal of the forward declaration. Everything else that would be a conflicting change I did somewhat differently, e.g. I do actually include Windows.h in common.h now. Also, I wrote the Win32 VDir so you can remove the dirent shim entirely.

waddlesplash:
As soon as https://github.com/mgba-emu/mgba/blob/master/src/platform/qt/OverrideView.cpp#L59 is switched off of C99 syntax, this will make mGBA master work with MSVC 100%.

CMakeLists.txt
```diff
@@ -147,6 +147,10 @@ if(WIN32)
 	list(APPEND CORE_VFS_SRC ${CMAKE_SOURCE_DIR}/src/util/vfs/vfs-fd.c ${CMAKE_SOURCE_DIR}/src/platform/windows/vfs-w32.c)
 	file(GLOB OS_SRC ${CMAKE_SOURCE_DIR}/src/platform/windows/*.c)
 	source_group("Windows-specific code" FILES ${OS_SRC})
+	if(MSVC)
+		include_directories(SYSTEM ${CMAKE_SOURCE_DIR}/src/platform/msvc)
```
endrift: Why did you make a new folder for this, when platform/windows already exists? getopt.h is only included in one place, so there's no need to add it to the include path either as specializing that include should work.

endrift:
Removing that bit of C99ism could be part of this PR. Just remove the struct name and parens and it should work.

CMakeLists.txt
waddlesplash: Because this is only for MSVC. MinGW and friends have their own getopt.h, and I didn't even test with those; for all I can tell this getopt was designed for and only for MSVC.

waddlesplash: I suppose I can just put getopt in windows as "getopt_msvc" and add some ifdefs.

src/platform/commandline.c
```diff
@@ -20,7 +20,11 @@
 #include "util/string.h"
 
 #include <fcntl.h>
+#ifdef _MSC_VER
+#include "platform/windows/getopt_msvc.h"
```
endrift: Why did you rename the file? platform/windows isn't in the search path for this file, so it shouldn't matter that the names collide.

src/platform/windows/getopt_msvc.h
```diff
@@ -0,0 +1,607 @@
+#ifndef __GETOPT_H__
```
endrift: Please add a mention of this file to the copyright section of the README.

PR #272, opened 2016-06-03
Android build project

zminhquanz:
this is beginning of Android build project , i hope developer interesting that


endrift:
I don't get it. You just appear to have merged master into 0.4.

zminhquanz:
i will be to contribute Android version , what do you think

PR #290, opened 2016-06-22
GBA Memory: Add emulation of Vast Fame protected carts

tzlion:
This adds support for the copy protection used in games by the Taiwanese unlicensed company Vast Fame. These use somewhat unique cartridge hardware which allows for various manipulations of SRAM writes and ROM reads to thwart attempts at dumping and emulation.

There aren't any dumps of these games publicly available (at least not the original versions; a couple dumps of hacked reprints are floating around, but they have their own problems) - I have three which I intend to release on my blog once emulator support is in place, and those seem to run fine with my changes. If you'd like a copy of the roms for testing let me know.

### DIFF
--- a/src/gba/memory.c
+++ b/src/gba/memory.c
@@ -81,6 +81,8 @@ void GBAMemoryInit(struct GBA* gba) {
 	cpu->memory.activeNonseqCycles16 = 0;
 	gba->memory.biosPrefetch = 0;
 	gba->memory.mirroring = false;
+
+	GBAVFameInit(&gba->memory.vfame);
 }
 
 void GBAMemoryDeinit(struct GBA* gba) {
@@ -383,6 +385,8 @@ static void GBASetActiveRegion(struct ARMCore* cpu, uint32_t address) {
 		LOAD_32(value, address & (SIZE_CART0 - 4), memory->rom); \
 	} else if (memory->mirroring && (address & memory->romMask) < memory->romSize) { \
 		LOAD_32(value, address & memory->romMask, memory->rom); \
+	} else if (memory->vfame.cartType) { \
+		value = GBAVFameGetPatternValue(address, 32); \
 	} else { \
 		mLOG(GBA_MEM, GAME_ERROR, "Out of bounds ROM Load32: 0x%08X", address); \
 		value = ((address & ~3) >> 1) & 0xFFFF; \
@@ -515,6 +519,8 @@ uint32_t GBALoad16(struct ARMCore* cpu, uint32_t address, int* cycleCounter) {
 			LOAD_16(value, address & (SIZE_CART0 - 2), memory->rom);
 		} else if (memory->mirroring && (address & memory->romMask) < memory->romSize) {
 			LOAD_16(value, address & memory->romMask, memory->rom);
+		} else if (memory->vfame.cartType) {
+			value = GBAVFameGetPatternValue(address, 16);
 		} else {
 			mLOG(GBA_MEM, GAME_ERROR, "Out of bounds ROM Load16: 0x%08X", address);
 			value = (address >> 1) & 0xFFFF;
@@ -528,6 +534,8 @@ uint32_t GBALoad16(struct ARMCore* cpu, uint32_t address, int* cycleCounter) {
 			LOAD_16(value, address & (SIZE_CART0 - 2), memory->rom);
 		} else if (memory->mirroring && (address & memory->romMask) < memory->romSize) {
 			LOAD_16(value, address & memory->romMask, memory->rom);
+		} else if (memory->vfame.cartType) {
+			value = GBAVFameGetPatternValue(address, 16);
 		} else {
 			mLOG(GBA_MEM, GAME_ERROR, "Out of bounds ROM Load16: 0x%08X", address);
 			value = (address >> 1) & 0xFFFF;
@@ -613,6 +621,8 @@ uint32_t GBALoad8(struct ARMCore* cpu, uint32_t address, int* cycleCounter) {
 			value = ((uint8_t*) memory->rom)[address & (SIZE_CART0 - 1)];
 		} else if (memory->mirroring && (address & memory->romMask) < memory->romSize) {
 			value = ((uint8_t*) memory->rom)[address & memory->romMask];
+		} else if (memory->vfame.cartType) {
+			value = GBAVFameGetPatternValue(address, 8);
 		} else {
 			mLOG(GBA_MEM, GAME_ERROR, "Out of bounds ROM Load8: 0x%08X", address);
 			value = (address >> 1) & 0xFF;
@@ -873,7 +883,11 @@ void GBAStore8(struct ARMCore* cpu, uint32_t address, int8_t value, int* cycleCo
 		if (memory->savedata.type == SAVEDATA_FLASH512 || memory->savedata.type == SAVEDATA_FLASH1M) {
 			GBASavedataWriteFlash(&memory->savedata, address, value);
 		} else if (memory->savedata.type == SAVEDATA_SRAM) {
-			memory->savedata.data[address & (SIZE_CART_SRAM - 1)] = value;
+			if (memory->vfame.cartType) {
+				GBAVFameSramWrite(&memory->vfame, address, value, memory->savedata.data);
+			} else {
+				memory->savedata.data[address & (SIZE_CART_SRAM - 1)] = value;
+			}
 			memory->savedata.dirty |= SAVEDATA_DIRT_NEW;
 		} else if (memory->hw.devices & HW_TILT) {
 			GBAHardwareTiltWrite(&memory->hw, address & OFFSET_MASK, value);
--- a/src/gba/memory.h
+++ b/src/gba/memory.h
@@ -12,6 +12,7 @@
 
 #include "gba/hardware.h"
 #include "gba/savedata.h"
+#include "gba/vfame.h"
 
 enum GBAMemoryRegion {
 	REGION_BIOS = 0x0,
@@ -118,6 +119,7 @@ struct GBAMemory {
 
 	struct GBACartridgeHardware hw;
 	struct GBASavedata savedata;
+	struct GBAVFameCart vfame;
 	size_t romSize;
 	uint32_t romMask;
 	uint16_t romID;
--- a/src/gba/vfame.c
+++ b/src/gba/vfame.c
@@ -1,0 +1,298 @@
+/* Copyright (c) 2016 taizou
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "vfame.h"
+#include "gba/gba.h"
+#include "gba/memory.h"
+
+static const uint8_t ADDRESS_REORDERING[4][16] = {
+	{ 15, 14, 9, 1, 8, 10, 7, 3, 5, 11, 4, 0, 13, 12, 2, 6 },
+	{ 15, 7, 13, 5, 11, 6, 0, 9, 12, 2, 10, 14, 3, 1, 8, 4 },
+	{ 15, 0, 3, 12, 2, 4, 14, 13, 1, 8, 6, 7, 9, 5, 11, 10 }
+};
+static const uint8_t ADDRESS_REORDERING_GEORGE[4][16] = {
+	{ 15, 7, 13, 1, 11, 10, 14, 9, 12, 2, 4, 0, 3, 5, 8, 6 },
+	{ 15, 14, 3, 12, 8, 4, 0, 13, 5, 11, 6, 7, 9, 1, 2, 10 },
+	{ 15, 0, 9, 5, 2, 6, 7, 3, 1, 8, 10, 14, 13, 12, 11, 4 }
+};
+static const uint8_t VALUE_REORDERING[4][16] = {
+	{ 5, 4, 3, 2, 1, 0, 7, 6 },
+	{ 3, 2, 1, 0, 7, 6, 5, 4 },
+	{ 1, 0, 7, 6, 5, 4, 3, 2 }
+};
+static const uint8_t VALUE_REORDERING_GEORGE[4][16] = {
+	{ 3, 0, 7, 2, 1, 4, 5, 6 },
+	{ 1, 4, 3, 0, 5, 6, 7, 2 },
+	{ 5, 2, 1, 6, 7, 0, 3, 4 }
+};
+
+static const int8_t MODE_CHANGE_START_SEQUENCE[5] = { 0x99, 0x02, 0x05, 0x02, 0x03 };
+static const int8_t MODE_CHANGE_END_SEQUENCE[5] = { 0x99, 0x03, 0x62, 0x02, 0x56 };
+
+// A portion of the initialisation routine that gets copied into RAM - Always seems to be present at 0x15C in VFame game ROM
+static const char INIT_SEQUENCE[16] = { 0xB4, 0x00, 0x9F, 0xE5, 0x99, 0x10, 0xA0, 0xE3, 0x00, 0x10, 0xC0, 0xE5, 0xAC, 0x00, 0x9F, 0xE5 };
+
+static bool _isInMirroredArea(uint32_t address, size_t romSize);
+static uint32_t _getPatternValue(uint32_t addr);
+static uint32_t _patternRightShift2(uint32_t addr);
+static int8_t _modifySramValue(enum GBAVFameCartType type, uint8_t value, int mode);
+static uint32_t _modifySramAddress(enum GBAVFameCartType type, uint32_t address, int mode);
+static int _reorderBits(uint32_t value, const uint8_t* reordering, int reorderLength);
+
+void GBAVFameInit(struct GBAVFameCart* cart) {
+	cart->cartType = VFAME_NO;
+	cart->sramMode = -1;
+	cart->romMode = -1;
+	cart->acceptingModeChange = false;
+}
+
+void GBAVFameDetect(struct GBAVFameCart* cart, uint32_t* rom, size_t romSize) {
+	cart->cartType = VFAME_NO;
+
+	// The initialisation code is also present & run in the dumps of Digimon Ruby & Sapphire from hacked/deprotected reprint carts,
+	// which would break if run in "proper" VFame mode so we need to exclude those..
+	if (romSize == 0x2000000) { // the deprotected dumps are 32MB but no real VF games are this size
+		return;
+	}
+
+	if (memcmp(INIT_SEQUENCE, &rom[0x57], sizeof(INIT_SEQUENCE)) == 0) {
+		cart->cartType = VFAME_STANDARD;
+		mLOG(GBA_MEM, INFO, "Vast Fame game detected");
+	}
+
+	// This game additionally operates with a different set of SRAM modes
+	// Its initialisation seems to be identical so the difference must be in the cart HW itself
+	// Other undumped games may have similar differences
+	if (memcmp("George Sango", &((struct GBACartridge*) rom)->title, 12) == 0) {
+		cart->cartType = VFAME_GEORGE;
+		mLOG(GBA_MEM, INFO, "George mode");
+	}
+}
+
+// This is not currently being used but would be called on ROM reads
+// Emulates mirroring used by real VF carts, but no games seem to rely on this behaviour
+uint32_t GBAVFameModifyRomAddress(struct GBAVFameCart* cart, uint32_t address, size_t romSize) {
+	if (cart->romMode == -1 && (address & 0x01000000) == 0) {
+		// When ROM mode is uninitialised, it just mirrors the first 0x80000 bytes
+		// All known games set the ROM mode to 00 which enables full range of reads, it's currently unknown what other values do
+		address &= 0x7FFFF;
+	} else if (_isInMirroredArea(address, romSize)) {
+		address -= 0x800000;
+	}
+	return address;
+}
+
+static bool _isInMirroredArea(uint32_t address, size_t romSize) {
+	address &= 0x01FFFFFF;
+	// For some reason known 4m games e.g. Zook, Sango repeat the game at 800000 but the 8m Digimon R. does not
+	if (romSize != 0x400000) {
+		return false;
+	}
+	if (address < 0x800000) {
+		return false;
+	}
+	if (address >= 0x800000 + romSize) {
+		return false;
+	}
+	return true;
+}
+
+// Looks like only 16-bit reads are done by games but others are possible...
+uint32_t GBAVFameGetPatternValue(uint32_t address, int bits) {
+	switch (bits) {
+	case 8:
+		if (address & 1) {
+			return _getPatternValue(address) & 0xFF;
+		} else {
+			return (_getPatternValue(address) & 0xFF00) >> 8;
+		}
+	case 16:
+		return _getPatternValue(address);
+	case 32:
+		return (_getPatternValue(address) << 2) + _getPatternValue(address + 2);
+	}
+	return 0;
+}
+
+// when you read from a ROM location outside the actual ROM data or its mirror, it returns a value based on some 16-bit transformation of the address
+// which the game relies on to run
+static uint32_t _getPatternValue(uint32_t addr) {
+	addr &= 0x1FFFFF;
+	uint32_t value = 0;
+	switch (addr & 0x1F0000) {
+	case 0x000000:
+	case 0x010000:
+		value = (addr >> 1) & 0xFFFF;
+		break;
+	case 0x020000:
+		value = addr & 0xFFFF;
+		break;
+	case 0x030000:
+		value = (addr & 0xFFFF) + 1;
+		break;
+	case 0x040000:
+		value = 0xFFFF - (addr & 0xFFFF);
+		break;
+	case 0x050000:
+		value = (0xFFFF - (addr & 0xFFFF)) - 1;
+		break;
+	case 0x060000:
+		value = (addr & 0xFFFF) ^ 0xAAAA;
+		break;
+	case 0x070000:
+		value = ((addr & 0xFFFF) ^ 0xAAAA) + 1;
+		break;
+	case 0x080000:
+		value = (addr & 0xFFFF) ^ 0x5555;
+		break;
+	case 0x090000:
+		value = ((addr & 0xFFFF) ^ 0x5555) - 1;
+		break;
+	case 0x0A0000:
+	case 0x0B0000:
+		value = _patternRightShift2(addr);
+		break;
+	case 0x0C0000:
+	case 0x0D0000:
+		value = 0xFFFF - _patternRightShift2(addr);
+		break;
+	case 0x0E0000:
+	case 0x0F0000:
+		value = _patternRightShift2(addr) ^ 0xAAAA;
+		break;
+	case 0x100000:
+	case 0x110000:
+		value = _patternRightShift2(addr) ^ 0x5555;
+		break;
+	case 0x120000:
+		value = 0xFFFF - ((addr & 0xFFFF) >> 1);
+		break;
+	case 0x130000:
+		value = 0xFFFF - ((addr & 0xFFFF) >> 1) - 0x8000;
+		break;
+	case 0x140000:
+	case 0x150000:
+		value = ((addr >> 1) & 0xFFFF) ^ 0xAAAA;
+		break;
+	case 0x160000:
+	case 0x170000:
+		value = ((addr >> 1) & 0xFFFF) ^ 0x5555;
+		break;
+	case 0x180000:
+	case 0x190000:
+		value = ((addr >> 1) & 0xFFFF) ^ 0xF0F0;
+		break;
+	case 0x1A0000:
+	case 0x1B0000:
+		value = ((addr >> 1) & 0xFFFF) ^ 0x0F0F;
+		break;
+	case 0x1C0000:
+	case 0x1D0000:
+		value = ((addr >> 1) & 0xFFFF) ^ 0xFF00;
+		break;
+	case 0x1E0000:
+	case 0x1F0000:
+		value = ((addr >> 1) & 0xFFFF) ^ 0x00FF;
+		break;
+	}
+
+	return value & 0xFFFF;
+}
+
+static uint32_t _patternRightShift2(uint32_t addr) {
+	uint32_t value = addr & 0xFFFF;
+	value >>= 2;
+	value += (addr & 3) == 2 ? 0x8000 : 0;
+	value += (addr & 0x10000) ? 0x4000 : 0;
+	return value;
+}
+
+void GBAVFameSramWrite(struct GBAVFameCart* cart, uint32_t address, uint8_t value, uint8_t* sramData) {
+	address &= 0x00FFFFFF;
+	// A certain sequence of writes to SRAM FFF8->FFFC can enable or disable "mode change" mode
+	// Currently unknown if these writes have to be sequential, or what happens if you write different values, if anything
+	if (address >= 0xFFF8 && address <= 0xFFFC) {
+		cart->writeSequence[address - 0xFFF8] = value;
+		if (address == 0xFFFC) {
+			if (memcmp(MODE_CHANGE_START_SEQUENCE, cart->writeSequence, sizeof(MODE_CHANGE_START_SEQUENCE)) == 0) {
+				cart->acceptingModeChange = true;
+			}
+			if (memcmp(MODE_CHANGE_END_SEQUENCE, cart->writeSequence, sizeof(MODE_CHANGE_END_SEQUENCE)) == 0) {
+				cart->acceptingModeChange = false;
+			}
+		}
+	}
+
+	// If we are in "mode change mode" we can change either SRAM or ROM modes
+	// Currently unknown if other SRAM writes in this mode should have any effect
+	if (cart->acceptingModeChange) {
+		if (address == 0xFFFE) {
+			cart->sramMode = value;
+		} else if (address == 0xFFFD) {
+			cart->romMode = value;
+		}
+	}
+
+	if (cart->sramMode == -1) {
+		// when SRAM mode is uninitialised you can't write to it
+		return;
+	}
+
+	// if mode has been set - the address and value of the SRAM write will be modified
+	address = _modifySramAddress(cart->cartType, address, cart->sramMode);
+	value = _modifySramValue(cart->cartType, value, cart->sramMode);
+	// these writes are mirrored
+	address &= 0x7FFF;
+	sramData[address] = value;
+	sramData[address + 0x8000] = value;
+}
+
+static uint32_t _modifySramAddress(enum GBAVFameCartType type, uint32_t address, int mode) {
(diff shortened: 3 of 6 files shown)


endrift:
I have a bunch of minor nitpicks with this patch that are easy to fix, but my one real complaint is that it puts code in the front of the ROM access memory routines, which are hot regions of code, so this has performance concerns. This is the same reason that I didn't want to implement the 64MB Video cartridges. Perhaps if a fast path can be devised (such as faking the ROM size) so that this isn't a performance concern anymore, I can merge this and add 64MB Video support, but for right now I'm not going to be able to merge this. I'll think about a way to do a fast path when I'm a little less busy, hopefully later today.

src/gba/memory.c
```diff
@@ -379,10 +381,15 @@ static void GBASetActiveRegion(struct ARMCore* cpu, uint32_t address) {
...
 #define LOAD_CART \
 	wait += waitstatesRegion[address >> BASE_OFFSET]; \
+	if ( memory->vfame.cartType ) { \
+		address = GBAVFameModifyRomAddress(&memory->vfame, address, memory->romSize); \
+	} \
 	if ((address & (SIZE_CART0 - 1)) < memory->romSize) { \
 		LOAD_32(value, address & (SIZE_CART0 - 4), memory->rom); \
 	} else if (memory->mirroring && (address & memory->romMask) < memory->romSize) { \
 		LOAD_32(value, address & memory->romMask, memory->rom); \
+	} else if (memory->vfame.cartType && address <= 0x09FFFFFF) { \
```
endrift: Should be `address < BASE_CART1`

src/gba/memory.c
```diff
@@ -511,10 +518,15 @@ uint32_t GBALoad16(struct ARMCore* cpu, uint32_t address, int* cycleCounter) {
...
 	case REGION_CART2:
 		wait = memory->waitstatesNonseq16[address >> BASE_OFFSET];
+		if (memory->vfame.cartType) {
+			address = GBAVFameModifyRomAddress(&memory->vfame, address, memory->romSize);
+		}
 		if ((address & (SIZE_CART0 - 1)) < memory->romSize) {
 			LOAD_16(value, address & (SIZE_CART0 - 2), memory->rom);
 		} else if (memory->mirroring && (address & memory->romMask) < memory->romSize) {
 			LOAD_16(value, address & memory->romMask, memory->rom);
+		} else if (memory->vfame.cartType && address <= 0x09FFFFFF) {
```
endrift: Ditto.

src/gba/memory.c
```diff
@@ -379,10 +381,15 @@ static void GBASetActiveRegion(struct ARMCore* cpu, uint32_t address) {
 
 #define LOAD_CART \
 	wait += waitstatesRegion[address >> BASE_OFFSET]; \
+	if ( memory->vfame.cartType ) { \
```
endrift: Please remove the extra spaces.

src/gba/memory.c
```diff
@@ -609,10 +621,15 @@ uint32_t GBALoad8(struct ARMCore* cpu, uint32_t address, int* cycleCounter) {
 	case REGION_CART2:
 	case REGION_CART2_EX:
 		wait = memory->waitstatesNonseq16[address >> BASE_OFFSET];
+		if ( memory->vfame.cartType ) {
```
endrift: Ditto.

src/gba/memory.c
```diff
@@ -609,10 +621,15 @@ uint32_t GBALoad8(struct ARMCore* cpu, uint32_t address, int* cycleCounter) {
...
+			address = GBAVFameModifyRomAddress(&memory->vfame, address, memory->romSize);
+		}
 		if ((address & (SIZE_CART0 - 1)) < memory->romSize) {
 			value = ((uint8_t*) memory->rom)[address & (SIZE_CART0 - 1)];
 		} else if (memory->mirroring && (address & memory->romMask) < memory->romSize) {
 			value = ((uint8_t*) memory->rom)[address & memory->romMask];
+		} else if (memory->vfame.cartType && address <= 0x09FFFFFF) {
```
endrift: Ditto.

src/gba/memory.c
```diff
@@ -639,6 +656,7 @@ uint32_t GBALoad8(struct ARMCore* cpu, uint32_t address, int* cycleCounter) {
 			value = 0xFF;
 		}
 		value &= 0xFF;
+
```
endrift: Unneeded new line.

src/gba/memory.h
```diff
@@ -143,6 +144,8 @@ struct GBAMemory {
 	int32_t eventDiff;
 
 	bool mirroring;
+
+	struct GBAVFameCart vfame;
```
endrift: Minor nitpick, but this should be put with the other stuff closer to the top, e.g. right under `savedata`.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+void GBAVFameDetect(struct GBAVFameCart* cart, uint32_t* rom, size_t romSize) {
+	cart->cartType = VFAME_NO;
+
+	// The initialisation code is also present & run in the dumps of Digimon Ruby & Sapphire from hacked/deprotected reprint carts,
+	// which would break if run in "proper" VFame mode so we need to exclude those..
+	if (romSize == 0x2000000) { // the deprotected dumps are 32MB but no real VF games are this size
+		return;
+	}
+
+	if (memcmp(INIT_SEQUENCE, &((char*)rom)[0x15C], 16) == 0) {
```
endrift: `memcmp(INIT_SEQUENCE, &rom[0x57], sizeof(INIT_SEQUENCE)) == 0` is a bit cleaner.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+		cart->cartType = VFAME_STANDARD;
+		mLOG(GBA_MEM, INFO, "Vast Fame game detected");
+	}
+
+	// This game additionally operates with a different set of SRAM modes
+	// Its initialisation seems to be identical so the difference must be in the cart HW itself
+	// Other undumped games may have similar differences
+	if (memcmp("George Sango", &((char*)rom)[0xA0], 12) == 0) {
```
endrift: There's the GBACartridge struct and also GBAGetTitle for this usecase.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+		cart->cartType = VFAME_GEORGE;
+		mLOG(GBA_MEM, INFO, "George mode");
+	}
+}
+
+uint32_t GBAVFameModifyRomAddress(struct GBAVFameCart* cart,uint32_t address, size_t romSize) {
```
endrift: Missing space.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+	if (cart->romMode == -1 && address <= 0x08FFFFFF) {
```
endrift: `address < BASE_ROM1`

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+		// When ROM mode is uninitialised, only the first 0x80000 bytes are readable
+		// All known games set the ROM mode to 00 which enables full range of reads, it's currently unknown what other values do
+		address = address & 0x7FFFF;
```
endrift: Is this meant to be mirrored? You say only the first 0x80000 bytes are readable, which implies to me the rest are some unitialized value. Might want to make the comment clearer if that's the case.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+	} else if (_isInMirroredArea(address, romSize)) {
+		address -= 0x800000;
+	}
+	return address;
+}
+
+static bool _isInMirroredArea(uint32_t address, size_t romSize) {
+	// For some reason known 4m games e.g. Zook, Sango repeat the game at 800000 but the 8m Digimon R. does not
+	return (address >= 0x08800000 && address < 0x8800000 + romSize && romSize == 0x400000);
```
endrift: These parens probably aren't necessary. You might want to break up the check into multiple if (...) { return false; } statements for clarity, though.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+}
+
+// Looks like only 16-bit reads are done by games but others are possible...
+uint32_t GBAVFameGetPatternValue(uint32_t address, int bits) {
+	switch(bits) {
+		case 8:
```
endrift: Extra indentation level here.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+			if (address % 2 == 1) {
```
endrift: `if (address & 1)`

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+	}
+	return 0;
+}
+
+// when you read from a ROM location outside the actual ROM data or its mirror, it returns a value based on some 16-bit transformation of the address
+// which the game relies on to run
+static uint32_t _getPatternValue(uint32_t addr) {
+	addr = addr & 0x1FFFFF;
+	uint32_t value = 0;
+	switch(addr & 0x1F0000) {
```
endrift: Extra indentation level

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+		case 0x000000:
+		case 0x010000:
+			value = _patternRightShift1(addr);
+		break;
```
endrift: `break` indentation should match the code, not the case.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	return value & 0xFFFF;
+}
+
+static uint32_t _patternRightShift1(uint32_t addr) {
+	addr = addr & 0x1FFFF;
+	addr = addr >> 1;
+	addr = addr & 0xFFFF;
+	return addr;
```
endrift: Since addr is unsigned, isn't this just `return (addr >> 1) & 0xFFFF;` ?

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+}
+static uint32_t _patternRightShift2(uint32_t addr) {
+	uint32_t value = addr & 0xFFFF;
+	value = value >> 2;
+	value += (addr % 4 == 2) ? 0x8000 : 0;
```
endrift: `(addr & 3) == 2`

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+	value += (addr & 0x10000) ? 0x4000 : 0;
+	return value;
+}
+
+void GBAVFameSramWrite(struct GBAVFameCart* cart,uint32_t address, int8_t value, uint8_t* sramData) {
+	// A certain sequence of writes to SRAM FFF8->FFFC can enable or disable "mode change" mode
+	// Currently unknown if these writes have to be sequential, or what happens if you write different values, if anything
+	if (address >= 0x0E00FFF8 && address <= 0xE00FFFC) {
+		cart->writeSequence[address - 0x0E00FFF8] = value;
+		if ( address == 0xE00FFFC ) {
```
endrift: Extra spaces.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+			if (memcmp(MODE_CHANGE_START_SEQUENCE, cart->writeSequence, 5 * sizeof(int8_t))==0){
```
endrift: `if (memcmp(MODE_CHANGE_START_SEQUENCE, cart->writeSequence, sizeof(MODE_CHANGE_START_SEQUENCE)) == 0) {`

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+				cart->acceptingModeChange = true;
+			}
+			if (memcmp(MODE_CHANGE_END_SEQUENCE, cart->writeSequence, 5 * sizeof(int8_t))==0){
```
endrift: Ditto.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+				return _reorderBits(address, ADDRESS_REORDERING_3, 16);
+			}
+		break;
+	}
+	return address;
+}
+
+static int8_t _modifySramValue(enum GBAVFameCartType type, int8_t value, int mode) {
+	switch(mode & 0xF) {
+		case 0x00: case 0x01: case 0x02: case 0x03:
```
endrift: Extra indentation level.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+			return value;
+		break;
```
endrift: Previous comment about break indentation.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+		case 0x04: case 0x05: case 0x06: case 0x07:
```
endrift: Please put each `case` on a separate line.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+			if (type == VFAME_GEORGE) {
+				return _reorderBits(value,VALUE_REORDERING_GEORGE_1, 8);
+			} else {
+				return _reorderBits(value,VALUE_REORDERING_1, 8);
+			}
+		break;
+		case 0x08: case 0x09: case 0x0A: case 0x0B:
```
endrift: Ditto.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+			if (type == VFAME_GEORGE) {
+				return _reorderBits(value,VALUE_REORDERING_GEORGE_2, 8);
+			} else {
+				return _reorderBits(value,VALUE_REORDERING_2, 8);
+			}
+		break;
+		case 0x0C: case 0x0D: case 0x0E: case 0x0F:
```
endrift: Ditto.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+		break;
+	}
+	return value;
+}
+
+// Reorder bits in a byte according to the reordering given
+static int _reorderBits(uint32_t value, const char* reordering, int reorderLength) {
+	uint32_t retval = value;
+
+	for(int x = reorderLength; x > 0; x--)
```
endrift: Space between `for` and the paren, and brace on the same line.

endrift: Also, please put the `int x` outside of the for loop. I've had some compilers complain about this even though it should be allowed.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+	{
+		char reorderPlace = reordering[reorderLength-x]; // get the reorder position
```
endrift: Spaces around the -

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+
+		uint32_t mask = 1 << reorderPlace; // move the bit to the position we want
+		uint32_t val = value & mask; // AND it with the original value
+		val = val >> reorderPlace; // move the bit back, so we have the correct 0 or 1
+
+		int destinationPlace = x-1;
```
endrift: Ditto.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+
+		uint32_t newMask = 1 << destinationPlace;
+		if (val == 1) {
+			retval = retval | newMask;
```
endrift: |=

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+		} else {
+			retval = retval & ~newMask;
```
endrift: &=

src/gba/vfame.h
```diff
@@ -0,0 +1,34 @@
+/* Copyright (c) 2016 taizou
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ // Support for copy protected unlicensed games from the company Vast Fame
```
endrift: Extra space before the comment

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+		}
+	}
+
+	return retval;
+}
+
```
endrift: Extra new line.

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
+/* Copyright (c) 2016 taizou
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "vfame.h"
+#include "memory.h"
```
endrift: `#include "gba/memory.h"`

src/gba/vfame.c
```diff
@@ -0,0 +1,324 @@
...
+	// these writes are mirrored
+	if (address >= 0xE008000) {
+		address -= 0x8000;
+	}
+	sramData[address & (SIZE_CART_SRAM - 1)] = value;
+	sramData[(address + 0x8000) & (SIZE_CART_SRAM - 1)] = value;
+}
+
+static uint32_t _modifySramAddress(enum GBAVFameCartType type, uint32_t address, int mode) {
+	switch(mode & 0xF) {
```
endrift: This can probably be done with a lookup table instead of a switch statement, but that's not a big deal.

src/gba/vfame.c
tzlion: Yep, it is supposed to be mirrored, I'll update the comment to reflect that.

tzlion:
Thanks - I'll work through your comments when I have some time, probably tomorrow evening.

As for the performance issue: the calls to GBAVFameModifyRomAddress in the ROM access routines could actually be removed for now, as so far I haven't observed any games relying on the mirroring behaviour it implements; it's just there for accuracy's sake, and to cover for the possibility of some undumped game making use of it.

So if you'd be otherwise happy to merge this PR without that code in place (after I've addressed your other comments), I could remove it for now, and then restore it on a separate branch later?

endrift:
Sounds good!

profi200:
"Digimon Ruby & Sapphire" u wot m8

I have seen this in multiple comments. It is "Pokémon" ;)

endrift:
It's a bootleg cart. It's not actually Pokémon.

profi200:
Oh, interesting. That's like all these R4 clones with their strange names ¯\_(ツ)_/¯

tzlion:
Okay, I've made a bunch of changes, hopefully that covers everything!

tzlion:
And yeah @profi200 they're actually original games (aside from the ripped off Digimon stuff):
Ruby https://www.youtube.com/watch?v=jqqjA-5r_2c
Sapphire https://www.youtube.com/watch?v=wNG4k0h8qTg

profi200:
lol looks a bit like a Pokémon clone. Anyway we better stop with the offtopic.

src/gba/memory.c
```diff
@@ -383,6 +385,8 @@ static void GBASetActiveRegion(struct ARMCore* cpu, uint32_t address) {
 		LOAD_32(value, address & (SIZE_CART0 - 4), memory->rom); \
 	} else if (memory->mirroring && (address & memory->romMask) < memory->romSize) { \
 		LOAD_32(value, address & memory->romMask, memory->rom); \
+	} else if (memory->vfame.cartType && address < BASE_CART1) { \
```
endrift: Come to think of it, does this not work with the cart mirrors, regardless of the WAITCNT setting? That seems kinda weird to me.

src/gba/vfame.c
```diff
@@ -0,0 +1,296 @@
...
+		address = address & 0x7FFFF;
+	} else if (_isInMirroredArea(address, romSize)) {
+		address -= 0x800000;
+	}
+	return address;
+}
+
+static bool _isInMirroredArea(uint32_t address, size_t romSize) {
+	// For some reason known 4m games e.g. Zook, Sango repeat the game at 800000 but the 8m Digimon R. does not
+	if (romSize != 0x400000) { return false; }
```
endrift: Sorry if I was unclear previously, but the return statement should be on its own line.

src/gba/vfame.c
```diff
@@ -0,0 +1,296 @@
...
+	case 8:
+		if (address & 1) {
+			return _getPatternValue(address) & 0xFF;
+		} else {
+			return (_getPatternValue(address) & 0xFF00) >> 8;
+		}
+	case 16:
+		return _getPatternValue(address);
+	case 32:
+		return (_getPatternValue(address) << 2) + _getPatternValue(address+2);
```
endrift: Spaces around +

src/gba/vfame.c
```diff
@@ -0,0 +1,296 @@
...
+		value = _patternRightShift1(addr) ^ 0x00FF;
+		break;
+	}
+
+	return value & 0xFFFF;
+}
+
+static uint32_t _patternRightShift1(uint32_t addr) {
+	return (addr >> 1) & 0xFFFF;
+}
```
endrift: Missing new line between the functions.

src/gba/vfame.c
```diff
@@ -0,0 +1,296 @@
...
+	return value;
+}
+
+void GBAVFameSramWrite(struct GBAVFameCart* cart,uint32_t address, int8_t value, uint8_t* sramData) {
+	// A certain sequence of writes to SRAM FFF8->FFFC can enable or disable "mode change" mode
+	// Currently unknown if these writes have to be sequential, or what happens if you write different values, if anything
+	if (address >= 0x0E00FFF8 && address <= 0xE00FFFC) {
+		cart->writeSequence[address - 0x0E00FFF8] = value;
+		if (address == 0xE00FFFC) {
+			if (memcmp(MODE_CHANGE_START_SEQUENCE, cart->writeSequence, sizeof(MODE_CHANGE_START_SEQUENCE))==0){
```
endrift: Spaces around ==

src/gba/vfame.c
```diff
@@ -0,0 +1,296 @@
...
+				cart->acceptingModeChange = true;
+			}
+			if (memcmp(MODE_CHANGE_END_SEQUENCE, cart->writeSequence, sizeof(MODE_CHANGE_END_SEQUENCE))==0){
```
endrift: Ditto.

src/gba/vfame.c
```diff
@@ -0,0 +1,296 @@
...
+	// these writes are mirrored
+	if (address >= 0xE008000) {
+		address -= 0x8000;
+	}
+	sramData[address & (SIZE_CART_SRAM - 1)] = value;
+	sramData[(address + 0x8000) & (SIZE_CART_SRAM - 1)] = value;
+}
+
+static uint32_t _modifySramAddress(enum GBAVFameCartType type, uint32_t address, int mode) {
+	mode = (mode & 0x3);
```
endrift: `mode &= 0x3`

src/gba/vfame.c
```diff
@@ -0,0 +1,296 @@
...
+		return value;
+	} else if (type == VFAME_GEORGE) {
+		return _reorderBits(value, VALUE_REORDERING_GEORGE[mode-1], 8);
+	} else {
+		return _reorderBits(value, VALUE_REORDERING[mode-1], 8);
+	}
+}
+
+// Reorder bits in a byte according to the reordering given
+static int _reorderBits(uint32_t value, const char* reordering, int reorderLength) {
```
endrift: Although I know my wait state code does this, it's kinda old and should be fixed: this should be const uint8_t*, since you're shifting on it. The possibility of negative shifts really bothers some compilers.

src/gba/vfame.h
```diff
@@ -0,0 +1,34 @@
...
+	VFAME_GEORGE = 2
+};
+
+struct GBAVFameCart {
+	enum GBAVFameCartType cartType;
+	int sramMode;
+	int romMode;
+	int8_t writeSequence[5];
+	char acceptingModeChange;
+};
```
endrift: `bool`

Also at some point, these might need to go into save states, especially for TAS reasons.

endrift:
Left a few more comments, but overall it looks pretty good now. Once these are taken care of, I think it'll be good to merge.

src/gba/vfame.h
tzlion: ah yeah, good point about save states, I didn't think about that.  For the most part save states should work without these being saved, at least as long as the ROM mirroring stuff is disabled; the games always seem to change SRAM mode immediately before doing a write, so the state of these modes beforehand doesn't usually matter. But definitely something I'll look at later.

endrift: It's worth noting that if I end up getting a way for the cart mirroring to work, I'll want to do the mapper for the GBA Video carts, too, so VFameCartType might turn into CartType at that point, to reduce redundancy.

tzlion: Yeah that would make sense - there is at least one other unlicensed company, Sintax, who have their own unique copy-protected cart hardware, so that's another potential cart type out there. (but I haven't been able to dump any of those yet)

src/gba/memory.c
tzlion: hmm, I just tested it and it does seem to work - would it be sufficient to replace this condition with something like
`if (memory->vfame.cartType && (address < BASE_CART1 || memory->mirroring && address < BASE_CART_SRAM))`
instead?

endrift: I'm not sure why either of those checks are necessary? It'll always be less than BASE_CART_SRAM; that's what the switch is for, and it'll only ever be above BASE_CART1 if it's in one of the auxiliary cart regions, which are accessed identically, but have different wait states. Furthermore, memory->mirroring is really only for the Classic NES cartridge carts, so wouldn't be applicable here. The former condition makes this only apply if it's above the ROM bounds and a vfame cart, which is the real condition here, isn't it?

src/gba/vfame.c
```diff
@@ -0,0 +1,303 @@
...
+	case 0x1F0000:
+		value = _patternRightShift1(addr) ^ 0x00FF;
+		break;
+	}
+
+	return value & 0xFFFF;
+}
+
+static uint32_t _patternRightShift1(uint32_t addr) {
+	return (addr >> 1) & 0xFFFF;
```
endrift: It might make sense to have these be inline now that it's a one-line function.

src/gba/vfame.c
```diff
@@ -0,0 +1,303 @@
...
+	value = value >> 2;
+	value += (addr & 3) == 2 ? 0x8000 : 0;
+	value += (addr & 0x10000) ? 0x4000 : 0;
+	return value;
+}
+
+void GBAVFameSramWrite(struct GBAVFameCart* cart, uint32_t address, int8_t value, uint8_t* sramData) {
+	// A certain sequence of writes to SRAM FFF8->FFFC can enable or disable "mode change" mode
+	// Currently unknown if these writes have to be sequential, or what happens if you write different values, if anything
+	if (address >= 0x0E00FFF8 && address <= 0xE00FFFC) {
```
endrift: I'm generally in favor of masking off the base address bits at the top (i.e. `address &= 0x00FFFFFF`) and then only using the offset for the rest of the function; much closer to what hardware does, and allows the SRAM bus to be relocated if needed, for whatever reason.

src/gba/vfame.c
```diff
@@ -0,0 +1,303 @@
...
+	if (cart->sramMode == -1) {
+		// when SRAM mode is uninitialised you can't write to it
+		return;
+	}
+
+	// if mode has been set - the address and value of the SRAM write will be modified
+	address = _modifySramAddress(cart->cartType, address, cart->sramMode);
+	value = _modifySramValue(cart->cartType, value, cart->sramMode);
+	// these writes are mirrored
+	if (address >= 0xE008000) {
```
endrift: `address &= 0x7FFF` works without needing a check, and removes the need to do the masking below.

src/gba/vfame.c
```diff
@@ -0,0 +1,303 @@
...
+	}
+}
+
+// Reorder bits in a byte according to the reordering given
+static int _reorderBits(uint32_t value, const uint8_t* reordering, int reorderLength) {
+	uint32_t retval = value;
+
+	int x;
+	for (x = reorderLength; x > 0; x--) {
+		char reorderPlace = reordering[reorderLength - x]; // get the reorder position
```
endrift: `char` -> `uint8_t`

src/gba/vfame.c
endrift: Missing a space between `switch` and the paren.

src/gba/vfame.c
```diff
@@ -0,0 +1,303 @@
...
+	return value & 0xFFFF;
+}
+
+static uint32_t _patternRightShift1(uint32_t addr) {
+	return (addr >> 1) & 0xFFFF;
+}
+
+static uint32_t _patternRightShift2(uint32_t addr) {
+	uint32_t value = addr & 0xFFFF;
+	value = value >> 2;
```
endrift: `value >>= 2`

endrift:
One last thing: You should add a small summary (i.e. the commit message minus the region of code affected) to the bottom of the feature list for 0.5 in CHANGES, and if you want, you can put your name in parens or brackets to show you wrote it. There isn't a style guide for which one to use, since it hasn't come up before, so feel free to do either.

src/gba/memory.c
tzlion: Ah I see. Makes sense! Sorry this is the first time I've worked with the GBA at all, forgive me if I ask stupid questions sometimes

tzlion:
okay, done! i've also made those last few changes you suggested, a couple other similar ones, and also fixed a bug I caught when testing George Sango (a mistake in its address reordering)

tzlion:
actually it occurs to me that I should specify GBA in the change log, as you're adding GB/C support and Vast Fame made games for those systems too :cat:

src/gba/vfame.c
```diff
@@ -0,0 +1,298 @@
...
+	sramData[address] = value;
+	sramData[address + 0x8000] = value;
+}
+
+static uint32_t _modifySramAddress(enum GBAVFameCartType type, uint32_t address, int mode) {
+	mode &= 0x3;
+	if (mode == 0) {
+		return address;
+	} else if (type == VFAME_GEORGE) {
+		return _reorderBits(address, ADDRESS_REORDERING_GEORGE[mode-1], 16);
```
endrift: Spaces around -

src/gba/vfame.c
```diff
@@ -0,0 +1,298 @@
...
+	} else {
+		return _reorderBits(address, ADDRESS_REORDERING[mode-1], 16);
```
endrift: Ditto.

src/gba/vfame.c
```diff
@@ -0,0 +1,298 @@
...
+	}
+}
+
+static int8_t _modifySramValue(enum GBAVFameCartType type, int8_t value, int mode) {
+	mode = (mode & 0xF) >> 2;
+	if (mode == 0) {
+		return value;
+	} else if (type == VFAME_GEORGE) {
+		return _reorderBits(value, VALUE_REORDERING_GEORGE[mode-1], 8);
```
endrift: Ditto.

src/gba/vfame.c
```diff
@@ -0,0 +1,298 @@
...
+	} else {
+		return _reorderBits(value, VALUE_REORDERING[mode-1], 8);
```
endrift: Ditto.

src/gba/vfame.c
```diff
@@ -0,0 +1,298 @@
...
+
+	int x;
+	for (x = reorderLength; x > 0; x--) {
+		uint8_t reorderPlace = reordering[reorderLength - x]; // get the reorder position
+
+		uint32_t mask = 1 << reorderPlace; // move the bit to the position we want
+		uint32_t val = value & mask; // AND it with the original value
+		val = val >> reorderPlace; // move the bit back, so we have the correct 0 or 1
+
+		int destinationPlace = x - 1;
```
endrift: `int` -> `unsigned`

src/gba/vfame.c
```diff
@@ -0,0 +1,298 @@
...
+static int _reorderBits(uint32_t value, const uint8_t* reordering, int reorderLength) {
+	uint32_t retval = value;
+
+	int x;
+	for (x = reorderLength; x > 0; x--) {
+		uint8_t reorderPlace = reordering[reorderLength - x]; // get the reorder position
+
+		uint32_t mask = 1 << reorderPlace; // move the bit to the position we want
+		uint32_t val = value & mask; // AND it with the original value
+		val = val >> reorderPlace; // move the bit back, so we have the correct 0 or 1
```
endrift: `val >>= reorderPlace`

endrift:
Sorry I keep finding nitpicks :P There are fewer and fewer each time though.

tzlion:
haha don't worry about it. done!

src/gba/vfame.c
```diff
@@ -0,0 +1,298 @@
...
+	case 32:
+		return (_getPatternValue(address) << 2) + _getPatternValue(address + 2);
+	}
+	return 0;
+}
+
+// when you read from a ROM location outside the actual ROM data or its mirror, it returns a value based on some 16-bit transformation of the address
+// which the game relies on to run
+static uint32_t _getPatternValue(uint32_t addr) {
+	addr = addr & 0x1FFFFF;
```
endrift: &=

src/gba/vfame.c
```diff
@@ -0,0 +1,298 @@
...
+}
+
+void GBAVFameSramWrite(struct GBAVFameCart* cart, uint32_t address, int8_t value, uint8_t* sramData) {
+	address &= 0x00FFFFFF;
+	// A certain sequence of writes to SRAM FFF8->FFFC can enable or disable "mode change" mode
+	// Currently unknown if these writes have to be sequential, or what happens if you write different values, if anything
+	if (address >= 0xFFF8 && address <= 0xFFFC) {
+		cart->writeSequence[address - 0xFFF8] = value;
+		if (address == 0xFFFC) {
+			if (memcmp(MODE_CHANGE_START_SEQUENCE, cart->writeSequence, sizeof(MODE_CHANGE_START_SEQUENCE)) == 0){
```
endrift: Missing space between paren and brace.

src/gba/vfame.c
```diff
@@ -0,0 +1,298 @@
...
+				cart->acceptingModeChange = true;
+			}
+			if (memcmp(MODE_CHANGE_END_SEQUENCE, cart->writeSequence, sizeof(MODE_CHANGE_END_SEQUENCE)) == 0){
```
endrift: Ditto.

endrift:
These seem to be the last nitpicks.

tzlion:
done and done :cat:

src/gba/vfame.c
```diff
@@ -0,0 +1,298 @@
...
+
+static uint32_t _patternRightShift2(uint32_t addr) {
+	uint32_t value = addr & 0xFFFF;
+	value >>= 2;
+	value += (addr & 3) == 2 ? 0x8000 : 0;
+	value += (addr & 0x10000) ? 0x4000 : 0;
+	return value;
+}
+
+void GBAVFameSramWrite(struct GBAVFameCart* cart, uint32_t address, int8_t value, uint8_t* sramData) {
```
endrift: It looks like this won't come up in any Vast Fame games, but value should be uint8_t, or else a game can set the mode to -1, which is your reserved value.

tzlion: Ah yeah, good point

endrift:
Whew! That took longer than expected.

PR #296, opened 2016-07-12
Don't include internal zip library when USE_MINIZIP is defined.

Polynomial-C:
According to CmakeLists.txt file USE_MINIZIP enables linking against system
minizip library but the code in src/util/vfs/vfs-zip.c doesn't reflect that.


endrift:
Thanks.

PR #297, opened 2016-07-12
Debugger: Support additional GDB stub packets

Touched:
Add support for GDB memory and register writing commands.

### DIFF
--- a/src/debugger/gdb-stub.c
+++ b/src/debugger/gdb-stub.c
@@ -6,6 +6,7 @@
 #include "gdb-stub.h"
 
 #include "core/core.h"
+#include "gba/memory.h"
 
 #include <signal.h>
 
@@ -149,7 +150,7 @@ static void _int2hex32(uint32_t value, char* out) {
 static uint32_t _readHex(const char* in, unsigned* out) {
 	unsigned i;
 	for (i = 0; i < 8; ++i) {
-		if (in[i] == ',') {
+		if (in[i] == ',' || in[i] == ':' || in[i] == '=') {
 			break;
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/debugger/gdb-stub.c
```diff
@@ -210,6 +211,62 @@ static void _step(struct GDBStub* stub, const char* message) {
...
+
+	if (size > 512) {
+		_error(stub, GDB_BAD_ARGUMENTS);
+		return;
+	}
+
+	struct ARMCore* cpu = stub->d.core->cpu;
+	for (i = 0; i < size; i++) {
+		// Parse escape char
+		uint8_t byte;
```
endrift: Please do the assignment here and don't increment the variable in the middle of an expression, for readability.

src/debugger/gdb-stub.c
```diff
@@ -210,6 +211,62 @@ static void _step(struct GDBStub* stub, const char* message) {
...
+		if ((byte = *readAddress++) == 0x7D) {
+			byte = *readAddress++ ^ 0x20;
```
endrift: Is this safe? The boundary condition here makes me a bit wary, especially given that we're incrementing with a fixed size that we don't update.

src/debugger/gdb-stub.c
```diff
@@ -243,6 +314,29 @@ static void _readGPRs(struct GDBStub* stub, const char* message) {
...
 }
 
+static void _writeRegister(struct GDBStub* stub, const char* message) {
+	struct ARMCore* cpu = stub->d.core->cpu;
+	const char* readAddress = message;
+
+	unsigned i = 0;
+	uint32_t reg = _readHex(readAddress, &i);
+	readAddress += i + 1;
+	uint32_t value = __builtin_bswap32(_readHex(readAddress, &i));
```
endrift: This isn't compatible with MSVC. Please ifdef this and use _byteswap_ulong for MSVC.

endrift:
Looks pretty good so far, but a few minor nitpicks. Please also add this to the Misc section in CHANGES.

src/debugger/gdb-stub.c
Touched: The size specified as a parameter in the packet is the size in addressable memory units, not the amount of bytes transmitted in the packet. Therefore it's not really possible to check when the packet is over without first reading it. It shouldn't overrun the buffer though, due to the earlier check for `size > 512`. 
Should I make the buffer overrun check more explicit by comparing `size` to `GDB_STUB_MAX_LINE`?

endrift: Hmm, I suppose since size <= 512, and the max line size is > 1024, it shouldn't be a problem.

endrift:
Looks good, will merge when I get home.

PR #322, opened 2016-08-06
Qt: Fix software renderer scaling

snickerbockers:
When the aspect ratio is locked and the window's aspect ratio is wider than the game boy's, the Qt software renderer divides the width by 2 instead of the game boy's height.  As a result, the video is way too wide.

Here's what it looks like without this fix:

mgba_without_my_aspect_ratio_fix

Here's what it looks like with this fix:

mgba_with_my_aspect_ratio_fix

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,7 @@ Bugfixes:
  - GBA Serialize: Savestates now properly store prefetch
  - PSP2: Fix accelerometer range
  - PSP2: Actually load screen mode setting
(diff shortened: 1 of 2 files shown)


endrift:
Please add this to the Bugfixes section of the change log.

snickerbockers:
Okay, I added it to the changelog.

endrift:
Please squash this down to one commit.

endrift:
Thanks!

PR #330, opened 2016-08-08
Qt: Set OpenGL version to 2.0

snickerbockers:
Prior to commit f95be3071abe4393d940cfb14c2aa4f7dd0f1398, the OpenGL renderer was using QGLFormat's default version number, which is 2.0 (see http://code.qt.io/cgit/qt/qtbase.git/tree/src/opengl/qgl_p.h#n79 ).  f95be3071abe4393d940cfb14c2aa4f7dd0f1398 changes this to 3.0.  This causes the OpenGL renderer to display an empty screen when a rom is loaded.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,7 @@ Bugfixes:
  - PSP2: Actually load screen mode setting
  - Qt: Fix bug in software renderer scaling
  - Debugger: Fix GDB breakpoints
+ - Qt: Set OpenGL version to 2.0
 Misc:
  - 3DS: Use blip_add_delta_fast for a small speed improvement
  - OpenGL: Log shader compilation failure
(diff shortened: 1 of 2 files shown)


endrift:
This isn't the right way to fix this. Moreover, I use features not available in 2.0 (FBOs).

taisel:
Not able to use EXT_Framebuffer_object for FBO access on older GPUs that can only do OGL 2? IIRC, old nvidia drivers stretching back to GeForce FX from '03 support this.

endrift:
I'm not sure Qt has a way of asking for that.

MoochMcGee:
Maybe you could try using OpenGL 3.1 with a compatibility context. That
way, you can get the best of both worlds.

PR #334, opened 2016-08-09
Qt: Reset button status after turning off autofire

nagitsu:
When disabling the autofire option the key is sometimes left in the pressed status, causing it to ignore further button presses. This change forces the key to be released in those cases.


endrift:
Looks good, please add this to the change log under Bugfixes.

nagitsu:
Done!

PR #352, opened 2016-08-19
[Wii] Fix garbage flash at startup, and other minor things

SuperrSonic:
I'd like to have the pixel accurate image stretched like that because it's supposed to be as close as square pixels you can get on GC/Wii, but it causes some problems, the interface is no longer overscan safe and I don't know the proper way to fix it. Without this change output looks too thin in PA mode.

The retraceCount change, I don't understand it but changing it to just be > makes games slightly faster, in FFVI I would notice audio crackling without this change.


endrift:
Please don't lump three changes into one PR, especially when I have no interest in merging two of them, or at least one of them and possibly another (needs more investigation).

PR #378, opened 2016-09-13
[WIP] ARMv7 Dynarec

MerryMage:
Still not in a state I'm quite happy with. This PR is more of a "progress report" more than anything and is not intended to be merged.

First things first: I kind of went against the original design that used the decoder and did something that more resembles the interpreter. This was because I was having hard to track down bugs and wanted to start fresh by first reproducing what the interpreter does as closely as possible before doing anything on top of that. If you wish for the decoder to be used, I would like to see what plans you have for updating the cycle counting code first before I refactor.

---

The "reproduce what the interpreter does exactly" phase was completed with commit 6c4949f3. Cycle accuracy is preserved throughout the implementation thereof.

At this point, the guest instruction `adds r0, #3` would compile to the following host instruction sequence:

```
movw r1, <low word of pc>
movt r1, <high word of pc>
str r1, [r0, <gprs[15]>]
movw r1, <value of prefech[0]>
movt r1, <value of prefech[0]>
str r1, [r0, <prefetch[0]>]
movw r1, <value of prefech[1]>
movt r1, <value of prefech[1]>
str r1, [r0, <prefetch[1]>]
ldr r1, [r0, <cycles>]
ldr r2, [r0, <nextEvent>]
cmp r1, r2
bge <epilogue>

ldr r1, [r0, <gpr[0]>]
adds r1, r1, #3
str r1, [r0, <gpr[0]>]
mrs cpsr, r1
lsr r1, r1, #24
strb r1, [r0, <MSB of ARMCore::cpsr>]
```

---

The first optimization was to have a register cache. This would remove unnecessary loads and stores to and from host registers and `ARMCore::gprs`. The cache also stores NZCV in a scratch register whenever possible. This was done in 8314c4a and 269e5a2.

The static register allocation decided upon follows:

| Host Register | Use |  |
| --- | --- | --- |
| R0 | ARMCore pointer | Saved before function call |
| R1 | Scratch register |  |
| R2 | Scratch register |  |
| R3 | Scratch register / NZCV temporary | Materalized as necessary |
| R4 | Guest R0 | Callee-saved register |
| R5 | Guest R1 | Callee-saved register |
| R6 | Guest R2 | Callee-saved register |
| R7 | Guest R3 | Callee-saved register |
| R8 | Guest R4 | Callee-saved register |
| R9 | - |  |
| R10 | Guest R5 | Callee-saved register |
| R11 | Guest R6 | Callee-saved register |
| R12 | Guest R7 | Volatile register; Saved before function call |
| R13 | - |  |
| R14 | Remaining cycle count | Remateralized as necessary |
| R15 | - |  |

The `ARMCore` pointer is in r0 because it is always the first argument in a function call and keeping it there saves us having to move it. `push {r0, r12}` and `pop {r0, r12}` respectively are done before and after a function call. ARM-EABI requires the stack to be 8 byte aligned in any case.

---

The second optimization was moving PC and Prefetch updating to the epilogue. This was performed in 3850941. At this point disassembly looks like this:

```
ldr r1, [r0, <cycles>]
ldr r2, [r0, <nextEvent>]
cmp r1, r2
movwge r1, <low word of pc>
movtge r1, <high word of pc>
bge <epilogue>

adds r1, r1, #3
mrs cpsr, r3
```

---

The third optimization was changing the way cycles were tracked internal to the dynarec. This was performed in 38bbb7e.

`r14` was chosen to be used as the cycle register. This is a volatile register, but that is okay because its values need to be written back to `ARMCore` before a function call in any case.

The value stored in `r14` is `cpu->cycles - cpu->nextEvent + 1`. Being offset by 1 means that when `cpu->cycles >= cpu->nextEvent`, the value in `r14` would be positive. This allows us to remove two instructions from typical instruction emission as follows:

```
adds r14, r14, #<cycles that happened last instruction>
movwpl r1, <low word of pc>
movtpl r1, <high word of pc>
bpl <epilogue>

adds r1, r1, #3
mrs cpsr, r3
```

At this point there is one optimization left that would still preserve cycle accuracy: The removal of `movtpl`. However I do not feel the additional complexity is worth saving one host cycle per guest instruction.

**TODO:** Completely eliminating event updating on instructions that do not perform a memory operation, trap or branch doesn't fully preserve cycle accuracy, but is also something to consider.

---

The fourth and most recent optimization is prediction of what the next trace would be at branches. This was performed in 59393a9f. While this eliminates hash-table lookups at branches and is a nice saving, this doesn't fully have the performance proper block linking would have. One would likely have to reimplement idle loop detection in the dynarec once block-linking is properly implemented.

---

I not properly tested the performance of the dynarec in its current state.

TODOs:
- [ ] Implement block-linking.
- [ ] Implement idle loop detection.
- [ ] Implement `STM`, `PUSH`, `LDM`, `POP` in terms of the `stm` and `ldm` host instructions. This would eliminate the need to flush the register cache for these instructions. (I would prefer to wait and see what you're doing wrt cycle counting before doing this.)
- [ ] Implement a return stack buffer, or some other kind of local PC cache for returns and unpredictable jumps.
- [ ] Compile instead of interpret `BX` (maybe).
- [ ] Compile instead of interpret `MUL` (maybe).


endrift:
As far as I'm aware, the cycle counting in the decoder is accurate except for MUL. If there are any exceptions, I'd need to track them down (a difficult process). The decoder was originally written to be used with a dynarec (and it shows) as a mini-IR. However, I don't have any major objections to using the emitter style; it kind of obviates a bunch of the stuff in the decoder from being necessary, though, which might be nice.

endrift:
Please also note that this is not (yet) compatible with ARMv6--several instructions are used here (e.g. MOVT and MOVW) that do not exist on ARMv6. Immediate loading would need to be replaced with a literal pool, which complicates design.

src/arm/dynarec-arm/dynarec-impl.c
```diff
@@ -0,0 +1,1225 @@
...
+		loadRegValue(ctx, rn, 1); \
+		EMIT_IMM(ctx, AL, 2, rs); \
+		PUSH_REGLIST_SAVE; /* needs to be here due to stack manipulation */  \
+		EMIT(ctx, SUBI, AL, ARM_SP, ARM_SP, 8); \
+		/* here we load the &cycles argument: */ \
+		EMIT(ctx, ADDI, AL, 3, REG_ARMCore, offsetof(struct ARMCore, cycles)); \
+		EMIT(ctx, STRI, AL, 3, ARM_SP, 0); \
+		/* arguments: (cpu, address, rs, LSM_ ## DIRECTION, &cycles) */ \
+		EMIT_IMM(ctx, AL, 3, LSM_ ## DIRECTION); \
+		EMIT(ctx, BL, AL, ctx->code, cpu->memory.CONCAT2(LS, Multiple)); \
```
endrift: Why is CONCAT2 needed? See `src/arm/debugger/memory-shim.c` for how I do this without a macro.

src/arm/dynarec-arm/dynarec-impl.c
```diff
@@ -0,0 +1,1225 @@
...
+
+#define FLUSH_HOST_NZCV \
+	flushNZCVToScratch(ctx);
+
+#define PREPARE_FOR_BL \
+	flushCycles(ctx); \
+	flushNZCVFully(ctx);
+
+#define PUSH_REGLIST_SAVE \
+	do { \
```
endrift: I don't believe the `do { } while (0)` trick is needed with C99, unless you care about scoping.

MerryMage:
I'm probably going to work on tests next, because debugging this without tests was painful.

src/arm/dynarec-arm/dynarec-impl.c
MerryMage: Noted. Will double check.

MerryMage:
I'll revisit this when I find more time.

PR #450, opened 2016-10-21
GBA Cheats: Fix GameShark ROM patches

waedt:
The address for GameShark ROM patch codes was being incorrectly calculated.  It appears that ActionReplay ROM patch codes were fixed in 3a3b7dffdb9e69763a9567119e45312d4b1453df, but no such fix was made for GameShark codes.


endrift:
Please add an entry to the CHANGES file, then I'll gladly merge. Thanks!

PR #453, opened 2016-10-27
GBA Memory: Support for Mo Jie Qi Bing by Vast Fame

tzlion:
Adds support for a feature of Vast Fame protected carts used by a newly dumped game, Mo Jie Qi Bing (aka Lord of the Rings). 

Also adds detection for this game based on its title, because it's based on Kiki KaiKai Advance and doesn't share the usual initialisation routine used to detect other Vast Fame games.

### DIFF
--- a/src/gba/vfame.c
+++ b/src/gba/vfame.c
@@ -58,7 +58,9 @@ void GBAVFameDetect(struct GBAVFameCart* cart, uint32_t* rom, size_t romSize) {
 		return;
 	}
 
-	if (memcmp(INIT_SEQUENCE, &rom[0x57], sizeof(INIT_SEQUENCE)) == 0) {
+	// Most games have the same init sequence in the same place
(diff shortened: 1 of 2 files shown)


src/gba/vfame.c
```diff
@@ -262,14 +264,18 @@ static uint32_t _modifySramAddress(enum GBAVFameCartType type, uint32_t address,
 }
 
 static int8_t _modifySramValue(enum GBAVFameCartType type, uint8_t value, int mode) {
-	mode = (mode & 0xF) >> 2;
-	if (mode == 0) {
-		return value;
+	int reorderType = (mode & 0xF) >> 2;
+	if (reorderType == 0) {
+		value = value;
```
endrift: `value = value`?

endrift:
Thanks! At some point it might make sense to hook into the overrides system instead of doing the game name check manually, but that would require a bit of finagling with how it works.

PR #456, opened 2016-11-03
Fix missing termination in call to SET_INPUT_DESCRIPTORS

garbear:
From https://github.com/libretro/mgba/pull/45


endrift:
Please add an entry to the Bugfixes section of the changelog, then I'll merge.

garbear:
like that?

endrift:
Generally I add them at the bottom, but I sort them before a release so it's not really a big deal where it goes.

PR #469, opened 2016-11-19
VBA Console Print

Touched:
libgba implements various ways of printing to the console, but the only one that seems to work in VBA is the [mappy print](https://github.com/devkitPro/libgba/blob/master/src/mappy_print.c). This works by checking for a `and r0, r0` where `r0` is `0xC0DED00D` before printing the null-terminated string at the address in `r2`. While this is not ideal, it seems to be the only way to get debug console printing working in both VBA and mGBA.

### DIFF
--- a/src/arm/arm.c
+++ b/src/arm/arm.c
@@ -315,3 +315,17 @@ void ARMRunFake(struct ARMCore* cpu, uint32_t opcode) {
 	cpu->prefetch[1] = cpu->prefetch[0];
 	cpu->prefetch[0] = opcode;
 }
+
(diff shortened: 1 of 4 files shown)


endrift:
mGBA already has a way of doing debug printing that is far less invasive to the ARM core. Furthermore, if libgba weren't so poorly designed around debug printing, it would be possible to detect mGBA and use its debug print routines, and fall back to mappy if that fails. I can try to add code to https://github.com/mgba-emu/mgba/blob/master/opt/libgba/mgba.c to do that on my end, and then possibly upstream it to libgba, but mappy print is far too invasive to the ARM core for me to merge it, as the ARM core is intended to be agnostic of the frontend.

PR #540, opened 2017-03-16
(Mostly) fix the build on MSVC (again).

waddlesplash:


src/util/vfs/vfs-zip.c
```diff
@@ -597,10 +597,12 @@ struct VFile* _vdzOpenFile(struct VDir* vd, const char* path, int mode) {
 	UNUSED(mode);
 	struct VDirZip* vdz = (struct VDirZip*) vd;
 
+#ifndef _MSC_VER
 	if ((mode & O_ACCMODE) != O_RDONLY) {
```
endrift: This check should be preserved, but I'm not sure how windows handles O_* flags.

waddlesplash: Hmm. Maybe this is the solution? https://github.com/ArtanisCV/STLport-Visual-Studio/blob/master/src/details/fstream_win32io.cpp#L39

waddlesplash: Yeah, looks like KDE's MSVC wrappers have something mostly identical to that. Where should I put that define? Along with the other MSVC defines in common.h?

PR #543, opened 2017-03-17
CMakeLists: Fix SDL-less build when there is no SDL on the system.

waddlesplash:


src/platform/sdl/CMakeLists.txt
```diff
@@ -24,6 +24,8 @@ if (NOT SDL2_FOUND AND NOT SDL_FOUND)
 	return()
 endif()
 
+add_definitions(-DBUILD_SDL)
```
endrift: Is this needed in both places?

waddlesplash: Indeed it is, see https://cmake.org/cmake/help/v3.0/command/add_definitions.html:

> Adds definitions to the compiler command line for sources in the current directory and below.

PR #544, opened 2017-03-17
.travis.yml: Use an explicitly defined build matrix.

waddlesplash:


endrift:
Please squash the PR.

.travis-deps.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 if [ $TRAVIS_OS_NAME = "osx" ]; then
 	brew update
-	brew install qt5 ffmpeg imagemagick sdl2 libzip libpng
+	brew install qt5 ffmpeg imagemagick sdl2 libzip libpng gcc@4.8
```
endrift: It would make sense to only install gcc on the matrix entry where it's needed, otherwise it's just wasting time.

endrift:
Didn't work. The travis-deps script doesn't export the variables past its scope.

PR #549, opened 2017-03-29
New library system.

waddlesplash:
~~This is so WIP it's absolutely ridiculous. Can't add new folders, can't switch list modes, games not in the NoIntro db appear blank, random commented out code still from Dolphin, etc. But it looks like I'll make rapid progress on it (when I have time...) so I'm putting it here.~~

**TODO list:**
 - [x] Wire up ability to switch between tree/list/grid/icon (icon view is kinda pointless for mGBA I guess?) modes
 - [x] Fix the treed mode showing entire pathnames
 - [x] Replace and remove old library system in last few places
 - [x] Threading
 - [x] Fix all TODO markers and stubbed code
 - [x] Fix all FIXMEs that are probably just more stubbed code
 - [x] Handle "Add folder to library"
 - [x] Right align size column
 - [x] Rework setup for icons/screenshots
- [x] Fix size column sorting
~~[ ] Enable the gamelist by default~~


src/platform/qt/CMakeLists.txt
```diff
@@ -22,13 +22,13 @@ if(BUILD_SDL)
 	if(SDL2_FOUND)
 		link_directories(${SDL2_LIBDIR})
 	endif()
-	list(APPEND PLATFORM_LIBRARY ${SDL_LIBRARY} ${SDLMAIN_LIBRARY})
+	list(APPEND PLATFORM_LIBRARY ${SDL_LIBRARY})
```
waddlesplash: Oh yeah, I should probably break these changes out into another PR. Without them, SDL + Qt build fails on MSVC.

src/platform/qt/Window.cpp
```diff
@@ -123,9 +125,10 @@ Window::Window(ConfigController* config, int playerId, QWidget* parent)
 	}, this);
 	m_config->updateOption("showLibrary");
 
-	connect(m_libraryView, &LibraryView::accepted, [this]() {
+	connect(m_libraryView, &LibraryController::startGame, [this]() {
 		VFile* output = m_libraryView->selectedVFile();
-		QPair<QString, QString> path = m_libraryView->selectedPath();
+		mLibraryEntry* entry = m_libraryView->selectedGame();
+		QPair<QString, QString> path = qMakePair(entry->filename, entry->base);
```
waddlesplash: paging @endrift: Should there even be a `selectedVFile` method? Or should that be taken care of here, since it fetches the Entry anyway?

Also what terminology should I use? "selectedGame" was the name in Dolphin, maybe "selectedEntry" is more appropriate here?

src/platform/qt/Window.h
```diff
@@ -195,7 +195,7 @@ private slots:
 #endif
 
 #ifdef USE_SQLITE3
-	LibraryView* m_libraryView;
+	LibraryController* m_libraryView;
```
waddlesplash: Wasn't sure what to call it, if you have a better name suggestion I'm all ears

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,158 @@
...
+	m_grid_widget->RemoveGames(removedGames);
+
+	for (mLibraryEntry* file : removedGames)
+		delete file;
+		*/
+
+	setDisabled(false);
+}
+
+void LibraryController::selectLastBootedGame()
```
waddlesplash: Dolphin had this feature where on startup it would select the last booted game in the gamelist. Implement or remove?

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,98 @@
...
+		size /= 1024.0;
+		unit = "kiB";
+	}
+	if (size >= 1024.0) {
+		size /= 1024.0;
+		unit = "MiB";
+	}
+	return QString("%0 %1").arg(size, 0, 'f', 1).arg(unit);
+}
+inline QString NicePlatformFormat(mPlatform platform) {
```
waddlesplash: This and the above names come straight from Dolphin, suggestions accepted

src/platform/qt/library/LibraryTree.cpp
```diff
@@ -0,0 +1,163 @@
...
+void LibraryTree::SetViewStyle(GameListStyle newStyle)
+{
+	if (newStyle == GameListStyle::STYLE_LIST) {
+		m_current_style = GameListStyle::STYLE_LIST;
+		setIndentation(0);
+		RebuildTree();
+	} else {
+		m_current_style = GameListStyle::STYLE_TREE;
+		setIndentation(20);
+		RebuildTree();
```
waddlesplash: Needs more camelCase

src/platform/qt/library/LibraryController.cpp
endrift: Implement, but not needed all at once.

src/platform/qt/CMakeLists.txt
```diff
@@ -22,13 +22,13 @@ if(BUILD_SDL)
...
 	list(APPEND PLATFORM_SRC ${PLATFORM_SRC} ${CMAKE_SOURCE_DIR}/src/platform/sdl/sdl-events.c ${CMAKE_SOURCE_DIR}/src/platform/sdl/sdl-audio.c)
 	include_directories(${SDL_INCLUDE_DIR} ${CMAKE_SOURCE_DIR}/src/platform/sdl)
 endif()
 
+include_directories("${CMAKE_CURRENT_BINARY_DIR}")
 set(CMAKE_AUTOMOC ON)
-set(CMAKE_INCLUDE_CURRENT_DIR ON)
```
waddlesplash: needs revert

src/platform/qt/ArchiveInspector.cpp
```diff
@@ -13,10 +13,11 @@ ArchiveInspector::ArchiveInspector(const QString& filename, QWidget* parent)
...
 {
 	m_ui.setupUi(this);
-	connect(m_ui.archiveView, &LibraryView::doneLoading, [this]() {
-		m_ui.loading->hide();
-	});
-	connect(m_ui.archiveView, SIGNAL(accepted()), this, SIGNAL(accepted()));
+	//connect(m_ui.archiveView, &LibraryController::doneLoading, [this]() {
+	//	m_ui.loading->hide();
+	//});
+	connect(m_ui.archiveView, &LibraryController::startGame, this, &ArchiveInspector::accepted);
```
endrift: Why not use the SIGNAL/SLOT notation here, to match the rest of the code?

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,173 @@
...
+		addGame(o);
+	}
+}
+void AbstractGameList::removeGames(QList<LibraryEntry*> items) {
+	for (LibraryEntry* o : items) {
+		removeGame(o);
+	}
+}
+
+LibraryController::LibraryController(QWidget* parent_widget, const QString& path, ConfigController* config)
```
endrift: Why is parent not the last argument, like it is everywhere else?

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,173 @@
...
+	: QStackedWidget(parent_widget),
```
endrift: Comma goes at the beginning of the next line.

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,131 @@
...
+		return QObject::tr("GB");
+#endif
+	default:
+		return QObject::tr("?");
+	}
+}
+
+class LibraryEntry {
+public:
+	LibraryEntry(const mLibraryEntry* entry) : base(entry->base), filename(entry->filename),
```
endrift: Implementation belongs in the cpp file.

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,131 @@
...
+class ConfigController;
+
+enum class LibraryStyle {
+	STYLE_LIST,
+	STYLE_TREE,
+	STYLE_GRID,
+	STYLE_ICON
+};
+
+inline QString NiceSizeFormat(size_t filesize)
```
endrift: Why are these inlines and not static methods?

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+	refresh();
+	selectLastBootedGame(); // TODO: doesn't belong here
+
+	setViewStyle(LibraryStyle::STYLE_TREE);
+}
+
+LibraryController::~LibraryController() {
+	for (LibraryEntry* file : m_games.values()) {
+		delete file;
```
endrift: Hugely unfond of the fact that the list contains pointers that need to be deleted one by one. shared_ptr or just storing the raw data is probably a much better idea.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,173 @@
...
+		m_libraryTree->setViewStyle(newStyle);
+	} else {
+		m_libraryGrid->selectEntry(selectedEntry());
+		setCurrentWidget(m_libraryGrid);
+		m_libraryGrid->setViewStyle(newStyle);
+	}
+}
+
+void LibraryController::selectEntry(LibraryEntry* entry) {
+	if (currentWidget() == m_libraryGrid) {
```
endrift: Um, can't you just use `currentWidget()->selectEntry(entry);`?

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+void LibraryController::setViewStyle(LibraryStyle newStyle) {
+	m_currentStyle = newStyle;
+
+	if (newStyle == LibraryStyle::STYLE_LIST || newStyle == LibraryStyle::STYLE_TREE) {
+		m_libraryTree->selectEntry(selectedEntry());
+		setCurrentWidget(m_libraryTree);
+		m_libraryTree->setViewStyle(newStyle);
+	} else {
+		m_libraryGrid->selectEntry(selectedEntry());
```
endrift: This code is duplicated unnecessarily. Assuming they're the same base class, you can just set a variable to the member based on the style type, then call the functions in only one place.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,173 @@
...
+void LibraryController::selectEntry(LibraryEntry* entry) {
+	if (currentWidget() == m_libraryGrid) {
+		m_libraryGrid->selectEntry(entry);
+	} else {
+		m_libraryTree->selectEntry(entry);
+	}
+}
+
+LibraryEntry* LibraryController::selectedEntry() {
+	if (currentWidget() == m_libraryGrid) {
```
endrift: Ditto.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,173 @@
...
+		return m_libraryGrid->selectedEntry();
+	} else {
+		return m_libraryTree->selectedEntry();
+	}
+}
+
+VFile* LibraryController::selectedVFile() {
+	LibraryEntry* entry = selectedEntry();
+	if (entry) {
+		mLibraryEntry bogus_entry; // TODO: this is probably a bad idea?
```
endrift: Yes, this is a bad idea. `LibraryEntry` should be a thin wrapper around `mLibraryEntry`. Just store the mLibraryEntry and query data off of it instead of duplicating all of the fields.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	m_libraryTree->addGames(newItems);
+	m_libraryGrid->addGames(newItems);
+
+	m_libraryTree->removeGames(removedGames);
+	m_libraryGrid->removeGames(removedGames);
+
+	for (LibraryEntry* file : removedGames) {
+		delete file;
```
endrift: Ditto.

src/platform/qt/library/LibraryController.cpp
waddlesplash: Because this class (just like the LibraryView) are used inside a .ui file, and uic passes `parent` as the first argument to the constructor.

src/platform/qt/library/LibraryController.cpp
waddlesplash: Hmm. Don't think QList has a pointer getter, and there need to be uniquely identifiable pointers. Will see what I can do.

src/platform/qt/library/LibraryController.cpp
waddlesplash: I'd have to `static_cast` or `dynamic_cast` it, but, sure.

src/platform/qt/library/LibraryController.cpp
waddlesplash: Hmm. I suppose I could rig up something that would be able to handle having a `mLibraryEntry` pointer that changed.

src/platform/qt/ArchiveInspector.cpp
waddlesplash: So, about that. The SIGNAL/SLOT notation is stringly typed, which means it'll compile no matter what happens, and if the signals/slots don't exist you'll get only a warning printed to the command line (which I think I saw some of... might need to look into that.)

If you use the `&method` syntax (C++11 and up only), it does type and function checking at compile time, so if the functions are incompatible or don't exist at all, it'll fail to compile. (Plus it looks cleaner imo.)

So for now it doesn't match the rest of the code, but it would make more sense to make the rest of the code match this rather than the other way round.

waddlesplash:
All comments (except first two) addressed.

src/platform/qt/Window.cpp
```diff
@@ -108,7 +110,7 @@ Window::Window(ConfigController* config, int playerId, QWidget* parent)
 		i = m_savedScale;
 	}
 #ifdef USE_SQLITE3
-	m_libraryView = new LibraryView();
+	m_libraryView = new LibraryController(nullptr, m_config->configDir() + "/library.sqlite3", m_config);
```
endrift: configDir() is a static method. No object needed.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+	AbstractGameList* newCurrentWidget = nullptr;
+	if (newStyle == LibraryStyle::STYLE_LIST || newStyle == LibraryStyle::STYLE_TREE) {
+		newCurrentWidget = m_libraryTree;
+	} else {
+		newCurrentWidget = m_libraryGrid;
+	}
+	newCurrentWidget->selectEntry(selectedEntry());
+	newCurrentWidget->setViewStyle(newStyle);
+	setCurrentWidget(dynamic_cast<QWidget*>(newCurrentWidget));
```
endrift: I'm not too familiar with diamond inheritance, but inheriting QWidget on AbstractGameList would avoid all of these dynamic casts. That being said, composition is better than inheritance, and it's probably entirely possible to do this without MI. Hell, a bunch of the functions are copy/paste implementations over slightly different types, so templates may even work.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,162 @@
...
+	QList<LibraryEntry> removedEntries;
+	for (QString& path : m_entries.keys()) {
+		if (!allEntries.contains(path)) {
+			removedEntries.append(m_entries.value(path));
+			m_entries.remove(path);
+		}
+	}
+
+	m_libraryTree->addEntries(newEntries);
+	m_libraryGrid->addEntries(newEntries);
```
endrift: I'm pretty unfond of the "let's have two copies of the model just slightly different" approach. Seems clunky and probably not the right way to go about this

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,135 @@
...
+class ConfigController;
+
+enum class LibraryStyle {
+	STYLE_LIST,
+	STYLE_TREE,
+	STYLE_GRID,
+	STYLE_ICON
+};
+
+static QString NiceSizeFormat(size_t filesize)
```
endrift: No initial caps on methods. Brace on same line. Why is this just floating here instead of a static class method, or in a better place?

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,135 @@
...
+		size /= 1024.0;
+		unit = "kiB";
+	}
+	if (size >= 1024.0) {
+		size /= 1024.0;
+		unit = "MiB";
+	}
+	return QString("%0 %1").arg(size, 0, 'f', 1).arg(unit);
+}
+static QString NicePlatformFormat(mPlatform platform)
```
endrift: Ditto.

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,135 @@
...
+	QString fullpath() const { return m_fullpath; }
+	QString title() const { return QString(entry->title); }
+	QByteArray internalTitle() const { return QByteArray(entry->internalTitle); }
+	QByteArray internalCode() const { return QByteArray(entry->internalCode); }
+	mPlatform platform() const { return entry->platform; }
+	size_t filesize() const { return entry->filesize; }
+	uint32_t crc32() const { return entry->crc32; }
+
+private:
+	friend class LibraryController;
```
endrift: friends 😢

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,135 @@
...
+	mLibraryEntry* entry;
+	const QString m_fullpath;
+};
+typedef QSharedPointer<CLibraryEntry> LibraryEntry;
```
endrift: QSharedPointer is deprecated. Use std::shared_ptr instead. Not fond of the name; it's opaque and doesn't imply that it's a pointer object, which leads to confusing code. Try `CLibraryEntry` to `LibraryEntry` and `LibraryEntry` to `LibraryEntryRef` or `LibraryEntryPtr`

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,135 @@
...
+	virtual void setViewStyle(LibraryStyle newStyle) = 0;
+
+	virtual void addEntry(LibraryEntry item) = 0;
+	virtual void addEntries(QList<LibraryEntry> items);
+
+	virtual void removeEntry(LibraryEntry item) = 0;
+	virtual void removeEntries(QList<LibraryEntry> items);
+};
+
+class LibraryController final : public QStackedWidget {
```
endrift: Why final? Nothing else in the codebase is final.

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,135 @@
...
+Q_OBJECT
+
+public:
+	LibraryController(QWidget* parent_widget = nullptr, const QString& path = "",
```
endrift: I still don't understand why the parent pointer is the first argument when exactly 0 Qt classes have it as the first argument.

src/platform/qt/library/LibraryGrid.cpp
```diff
@@ -0,0 +1,93 @@
...
+	m_ui->setupUi(this);
+	setViewStyle(LibraryStyle::STYLE_GRID);
+
+	connect(this, &QListWidget::itemActivated, this, &LibraryGrid::StartGame);
+}
+LibraryGrid::~LibraryGrid() {
+	for (QListWidgetItem* i : m_items.keys()) { // TODO: probably not needed
+		delete i;
+	}
+	delete m_ui;
```
endrift: Is there no way to automatically parent m_ui to this object so it gets automatically deleted? unique_ptr maybe?

src/platform/qt/library/LibraryGrid.cpp
```diff
@@ -0,0 +1,93 @@
...
+		setIconSize(QSize(ICON_BANNER_WIDTH, ICON_BANNER_HEIGHT));
+		setViewMode(QListView::ListMode);
+	}
+
+	// QListView resets this when you change the view mode, so let's set it again
+	setDragEnabled(false);
+}
+
+void LibraryGrid::addEntry(LibraryEntry item) {
+	if (m_items.values().contains(item))
```
endrift: Braces missing.

src/platform/qt/library/LibraryGrid.cpp
```diff
@@ -0,0 +1,93 @@
...
+	i->setText(item->displayTitle());
+	//if (gameItem->compressed) // TODO
+	//	i->setTextColor(QColor("#00F"));
+
+	addItem(i);
+	m_items.insert(i, item);
+}
+
+void LibraryGrid::removeEntry(LibraryEntry gameItem) {
+	if (!m_items.values().contains(gameItem))
```
endrift: Ditto.

src/platform/qt/library/LibraryGrid.cpp
```diff
@@ -0,0 +1,93 @@
...
+		return;
+
+	QListWidgetItem* i = m_items.key(gameItem);
+	m_items.remove(i);
```
endrift: This is what the `take` method is for.

src/platform/qt/library/LibraryTree.cpp
```diff
@@ -0,0 +1,168 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include "LibraryTree.h"
+
+#include "ui_LibraryTree.h"
+
+#include <QDir>
+
+namespace QGBA {
+
+LibraryTree::LibraryTree(QWidget* parent_widget)
```
endrift: `parent`, not `parent_widget`

src/platform/qt/library/LibraryTree.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+LibraryTree::~LibraryTree() {
+	int count = topLevelItemCount();
+	for (int a = 0; a < count; a++) {
+		takeTopLevelItem(0);
+	}
+
+	for (QTreeWidgetItem* i : m_path_nodes.values()) {
+		count = i->childCount();
+		for (int a = 0; a < count; a++)
```
endrift: Missing braces

src/platform/qt/library/LibraryTree.h
```diff
@@ -0,0 +1,60 @@
...
+		COL_SIZE = 3,
+		COL_CRC32 = 4,
+	};
+
+	Ui::LibraryTree* m_ui;
+	LibraryStyle m_currentStyle;
+
+	bool m_deferredTreeRebuild = false;
+	QMap<QTreeWidgetItem*, LibraryEntry> m_items;
+	QMap<QString, QTreeWidgetItem*> m_path_nodes;
```
endrift: `m_pathNodes`. Any chance of these being managed pointers so you don't have to manually delete each one?

src/platform/qt/library/LibraryTree.h
```diff
@@ -0,0 +1,60 @@
...
+
+	virtual void addEntries(QList<LibraryEntry> items) override;
+	virtual void addEntry(LibraryEntry item) override;
+	virtual void removeEntry(LibraryEntry item) override;
+
+signals:
+	void StartGame();
+
+private slots:
+	void ItemActivated(QTreeWidgetItem* item);
```
endrift: `itemActivated`

src/platform/qt/library/LibraryTree.h
```diff
@@ -0,0 +1,60 @@
...
+	virtual void selectEntry(LibraryEntry game) override;
+
+	virtual void setViewStyle(LibraryStyle newStyle) override;
+
+	virtual void addEntries(QList<LibraryEntry> items) override;
+	virtual void addEntry(LibraryEntry item) override;
+	virtual void removeEntry(LibraryEntry item) override;
+
+signals:
+	void StartGame();
```
endrift: `startGame`

src/platform/qt/library/LibraryTree.h
```diff
@@ -0,0 +1,60 @@
...
+
+private slots:
+	void ItemActivated(QTreeWidgetItem* item);
+
+private:
+	enum Columns {
```
endrift: Why explicitly instantiate these values? There will surely be more columns later, and their values aren't particularly important.

src/platform/qt/library/LibraryTree.cpp
```diff
@@ -0,0 +1,168 @@
...
+	LibraryEntry currentGame = selectedEntry();
+
+	int count = topLevelItemCount();
+	for (int a = 0; a < count; a++) {
+		takeTopLevelItem(0);
+	}
+
+	for (QTreeWidgetItem* i : m_path_nodes.values()) {
+		count = i->childCount();
+		for (int a = 0; a < count; a++)
```
endrift: Missing braces.

src/platform/qt/library/LibraryTree.h
```diff
@@ -0,0 +1,60 @@
...
+	};
+
+	Ui::LibraryTree* m_ui;
+	LibraryStyle m_currentStyle;
+
+	bool m_deferredTreeRebuild = false;
+	QMap<QTreeWidgetItem*, LibraryEntry> m_items;
+	QMap<QString, QTreeWidgetItem*> m_path_nodes;
+
+	void RebuildTree();
```
endrift: `rebuildTree`

src/platform/qt/library/LibraryTree.h
```diff
@@ -0,0 +1,60 @@
...
+	void ResizeAllCols();
```
endrift: `resizeAllCols`

src/platform/qt/library/LibraryController.cpp
waddlesplash: They probably don't even need to be dynamic_casts, I could just make them static_casts. I haven't ever used the `class A : virtual public B` method of solving diamond inheritance, but I could give it a shot. Personally, I think MI is the best and cleanest solution here; there's a lot of social stigma around it, but this is one case where it absolutely is better than the alternatives.

src/platform/qt/library/LibraryController.cpp
waddlesplash: What would you suggest? Giving the GameList widgets access to m_entries wouldn't solve anything, because then they'd have to do the `for ...` to find the items added. "Seems" clunky, but there's not really any way this could fall over that I can see...

Additionally, this mechanism makes it possible to have the treed gamelist, which was quite well liked when Dolphin had it.

src/platform/qt/library/LibraryController.h
waddlesplash: I know, I know, everyone hates friends. I think they make sense here though.

endrift: I don't even see anything in LibraryController reaching into the struct. Why is it being declared at all?

waddlesplash: LibraryController changes the pointer to the mLibraryEntry every time it re-queries the library?

src/platform/qt/library/LibraryController.h
waddlesplash: It was just general practice over on Dolphin (as well as my own general habit) and I didn't remove it. I can if you don't want it, or I can also just put "add final to classes" on my TODO list for refactoring mgba-qt...

src/platform/qt/library/LibraryController.h
endrift: Ah, that didn't show up on my search the first time.

Setters. This is what setters are for. Or even make it public, seeing as it's a struct that we can basically fully reach into anyway. Friends are an ugly hack.

src/platform/qt/library/LibraryController.h
endrift: I don't think it makes a difference if everything is final or not, I just want to keep it consistent.

src/platform/qt/library/LibraryController.h
waddlesplash: Where should it go? It's just a size formatter (copied out of your library code). I suppose it could go on just the tree. In DolphinQt, this was in a Utils.h.

endrift: There's no reason that this PR can't introduce a utils.h into the mGBA codebase. (Lowercase because it's not a class name.)

src/platform/qt/library/LibraryController.cpp
endrift: I'd have to take a closer look at this to make assertions about how bad MI is here, but the fact that you need to cast your own classes to QWidget when they all inherit QWidget shows that you need a better inheritance chain. Diamond inheritance solves that for MI, but diamond inheritance is a huge fucking pain any way you slice it.

src/platform/qt/library/LibraryTree.h
waddlesplash: These are the indexes of the columns as they are in the QTreeWidget. So if we insert/add/remove/etc. columns, the numbers will change, and rather than changing a bunch of magic numbers all over the .cpp, instead you just change this enum. Keeps everything nice and tidy.

endrift: The question is not "why is there an enum". The question is "why are you explicitly assigning the values to the enum fields when they're all sequential?"

src/platform/qt/library/LibraryController.h
waddlesplash: Huh? Zero? Basically all QWidget-derived classes have it as the first argument. I can't think of any that don't. If you reverse the order, the compile fails because the uic-generated files expect it to be the first argument.

endrift: From my quick survey of a bunch of classes, a handful start with it, and most of those only have a window flag parameter after it. A few have it first, but most have a constructor that is just the parent, and then a constructor that is everything with the parent last.

Also they ALL call it `parent` not `parent_widget`, a custom my codebase inherits.

src/platform/qt/library/LibraryController.h
waddlesplash: Well, you can fully reach into the struct, but nobody besides the LibraryController is supposed to *change* the pointer to the struct.

I suppose I can just make it public anyway.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include "LibraryTree.h"
+
+namespace QGBA {
+
+LibraryEntry::LibraryEntry(mLibraryEntry* entry)
+    : entry(entry)
+    , m_fullpath(QString("%1/%2").arg(entry->base, entry->filename))
+{
+}
+LibraryEntry::~LibraryEntry() {
```
endrift: Missing newline before.

endrift: Why is there an explicit destructor if it's empty?

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,162 @@
...
+	}
+}
+void AbstractGameList::removeEntries(QList<LibraryEntryRef> items) {
+	for (LibraryEntryRef o : items) {
+		addEntry(o);
+	}
+}
+
+LibraryController::LibraryController(QWidget* parent, const QString& path, ConfigController* config)
+    : QStackedWidget(parent)
```
endrift: Mixed indentation.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,162 @@
...
+	} else {
+		newCurrentWidget = m_libraryGrid;
+	}
+	newCurrentWidget->selectEntry(selectedEntry());
+	newCurrentWidget->setViewStyle(newStyle);
+	setCurrentWidget(dynamic_cast<QWidget*>(newCurrentWidget));
+}
+
+void LibraryController::selectEntry(LibraryEntryRef entry) {
+	dynamic_cast<AbstractGameList*>(currentWidget())->selectEntry(entry);
```
endrift: Still super un-fond of using `dynamic_cast` literally ever.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+	mLibraryListingClear(&m_listing);
+	mLibraryGetEntries(m_library, &m_listing, 0, 0, nullptr);
+	for (size_t i = 0; i < mLibraryListingSize(&m_listing); i++) {
+		mLibraryEntry* entry = mLibraryListingGetPointer(&m_listing, i);
+		QString fullpath = QString("%1/%2").arg(entry->base, entry->filename);
+		if (m_entries.contains(fullpath)) {
+			m_entries.value(fullpath)->entry = entry;
+		} else {
+			LibraryEntryRef libentry = std::shared_ptr<LibraryEntry>(new LibraryEntry(entry));
```
endrift: `std::make_shared`

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,105 @@
+/* Copyright (c) 2014-2017 waddlesplash
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#pragma once
```
endrift: `#pragma once` isn't used anywhere in this codebase. I'm not opposed to revamping the entire Qt frontend to use it, but maintaining consistency until such time is a good idea. Cf. my stance of signals/slots.

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,105 @@
...
+	virtual void setViewStyle(LibraryStyle newStyle) = 0;
+
+	virtual void addEntry(LibraryEntryRef item) = 0;
+	virtual void addEntries(QList<LibraryEntryRef> items);
+
+	virtual void removeEntry(LibraryEntryRef item) = 0;
+	virtual void removeEntries(QList<LibraryEntryRef> items);
+};
+
+class LibraryController final : public QStackedWidget {
```
endrift: See previous statements about use of `final`.

src/platform/qt/library/LibraryController.h
```diff
@@ -0,0 +1,105 @@
...
+Q_OBJECT
+
+public:
+	LibraryController(QWidget* parent = nullptr, const QString& path = "",
```
endrift: Shouldn't this be const `const QString& path = QString()`?

src/platform/qt/library/LibraryGrid.cpp
```diff
@@ -0,0 +1,88 @@
...
+// Game banner image size
+static const quint32 GRID_BANNER_WIDTH = 144;
+static const quint32 GRID_BANNER_HEIGHT = 48;
+
+static const quint32 ICON_BANNER_WIDTH = 64;
+static const quint32 ICON_BANNER_HEIGHT = 64;
+
+LibraryGrid::LibraryGrid(QWidget* parent)
+    : QListWidget(parent)
+    , m_ui(std::make_unique<Ui::LibraryGrid>())
```
endrift: `std::make_unique` is a C++14 feature. This codebase is C++11 (for now).

src/platform/qt/library/LibraryGrid.cpp
```diff
@@ -0,0 +1,88 @@
...
+{
+	m_ui->setupUi(this);
+	setViewStyle(LibraryStyle::STYLE_GRID);
+
+	connect(this, &QListWidget::itemActivated, this, &LibraryGrid::startGame);
+}
+LibraryGrid::~LibraryGrid() {
```
endrift: Missing newline and questionable empty destructor.

src/platform/qt/library/LibraryTree.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include "ui_LibraryTree.h"
+
+#include <QDir>
+
+namespace QGBA {
+
+class TreeWidgetItem : public QTreeWidgetItem {
+public:
+	TreeWidgetItem(QTreeWidget* parent = nullptr) : QTreeWidgetItem(parent) {}
+	void setFilesize(size_t size) {
```
endrift: If it's more than one line, put it in the .cpp file.

src/platform/qt/library/LibraryTree.cpp
```diff
@@ -0,0 +1,170 @@
...
+		m_size = size;
+		setText(LibraryTree::COL_SIZE, niceSizeFormat(size));
+	}
+
+	virtual bool operator<(const QTreeWidgetItem& other) const override {
```
endrift: Ditto.

src/platform/qt/library/LibraryTree.cpp
```diff
@@ -0,0 +1,170 @@
...
+		return ((column == LibraryTree::COL_SIZE) ?
+		    m_size < dynamic_cast<const TreeWidgetItem*>(&other)->m_size :
+		    QTreeWidgetItem::operator<(other));
+	}
+protected:
+	size_t m_size = 0;
+};
+
+LibraryTree::LibraryTree(QWidget* parent)
+    : QTreeWidget(parent)
```
endrift: Tabs.

src/platform/qt/library/LibraryTree.cpp
```diff
@@ -0,0 +1,170 @@
...
+	m_ui->setupUi(this);
+
+	setViewStyle(LibraryStyle::STYLE_TREE);
+	//setIconSize(QSize(BANNER_WIDTH, BANNER_HEIGHT));
+	sortByColumn(COL_NAME, Qt::AscendingOrder);
+
+	connect(this, &QTreeWidget::itemActivated, this, &LibraryTree::itemActivated);
+}
+
+LibraryTree::~LibraryTree() {
```
endrift: See previous comment about empty destructors.

src/platform/qt/utils.h
```diff
@@ -0,0 +1,17 @@
+/* Copyright (c) 2013-2017 Jeffrey Pfau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#pragma once
```
endrift: See previous statement about `#pragma once`

src/platform/qt/library/LibraryTree.cpp
waddlesplash: This is the .cpp file.

src/platform/qt/library/LibraryTree.cpp
waddlesplash: Dang Qt Creator, thinking it's smarter than me re indentation.

src/platform/qt/library/LibraryTree.cpp
endrift: Ok sorry, I meant not inline in the class definition.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+QPair<QString, QString> LibraryController::selectedPath() {
+	LibraryEntryRef e = selectedEntry();
+	return e ? qMakePair(e->base(), e->filename()) : qMakePair<QString, QString>("", "");
+}
+
+void LibraryController::addDirectory(const QString& dir) {
+	m_loaderThread.m_directory = dir;
+	m_loaderThread.m_library = m_library;
+	// m_loaderThread temporarily owns the library
```
endrift: The backend is threadsafe, afaik, so this might not be necessary.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -0,0 +1,185 @@
...
+	connect(&m_loaderThread, &QThread::finished, this, &LibraryController::refresh, Qt::QueuedConnection);
+
+	refresh();
+
+	setViewStyle(LibraryStyle::STYLE_TREE);
+}
+
+LibraryController::~LibraryController() {
+	mLibraryListingDeinit(&m_listing);
+	if (m_library) {
```
endrift: If the loader thread is active, make sure to wait on it finishing so you don't leak the m_library.

src/platform/qt/library/LibraryGrid.cpp
```diff
@@ -0,0 +1,86 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include "LibraryGrid.h"
+
+#include "ui_LibraryGrid.h"
+
+namespace QGBA {
+
+// Game banner image size
+static const quint32 GRID_BANNER_WIDTH = 144;
```
endrift: Not really important, but is there any reason these aren't class constants?

src/platform/qt/library/LibraryGrid.cpp
```diff
@@ -0,0 +1,86 @@
...
+	setDragEnabled(false);
+}
+
+void LibraryGrid::addEntry(LibraryEntryRef item) {
+	if (m_items.contains(item)) {
+		return;
+	}
+
+	QListWidgetItem* i = new QListWidgetItem;
+	//i->setIcon(QIcon(gameItem->GetBitmap() // FIXME
```
endrift: Please clean up commented out code.

src/platform/qt/library/LibraryController.cpp
waddlesplash: Ah, right, I thought of that and then promptly forgot. Will do.

src/platform/qt/library/LibraryController.cpp
waddlesplash: It'd be nice to have a definite on that and not an "afaik", but any reason not to have the additional safety anyway?

src/platform/qt/library/LibraryGrid.cpp
waddlesplash: No, and I should probably remove them anyway because they're totally wrong (I think these are the numbers from Dolphin.)

src/platform/qt/library/LibraryTree.cpp
```diff
@@ -35,33 +34,42 @@ bool TreeWidgetItem::operator<(const QTreeWidgetItem& other) const {
...
-	m_ui->setupUi(this);
+	m_widget->setObjectName("LibraryTree");
+	m_widget->setSortingEnabled(true);
+
+	QTreeWidgetItem* header = new QTreeWidgetItem({
+		QApplication::translate("LibraryTree", "Name", nullptr),
+		QApplication::translate("LibraryTree", "Location", nullptr),
+		QApplication::translate("LibraryTree", "Platform", nullptr),
+		QApplication::translate("LibraryTree", "Size", nullptr),
+		QApplication::translate("LibraryTree", "CRC32", nullptr),
```
endrift: Is this the correct way to do the equivalent of `QObject::tr` outside of a QObject?

src/platform/qt/library/LibraryGrid.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+void LibraryGrid::addEntry(LibraryEntryRef item) {
+	if (m_items.contains(item)) {
+		return;
+	}
+
+	QListWidgetItem* i = new QListWidgetItem;
+	//i->setIcon(QIcon(item.bitmap
+	//	.scaled(GRID_BANNER_WIDTH, GRID_BANNER_HEIGHT, Qt::KeepAspectRatio, Qt::SmoothTransformation)));
```
endrift: Please remove commented out code.

src/platform/qt/library/LibraryGrid.ui
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
endrift: This file isn't needed anymore.

src/platform/qt/library/LibraryTree.ui
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
endrift: This file isn't needed anymore.

src/platform/qt/library/LibraryTree.h
```diff
@@ -0,0 +1,57 @@
...
+	virtual LibraryEntryRef selectedEntry() override;
+	virtual void selectEntry(LibraryEntryRef game) override;
+
+	virtual void setViewStyle(LibraryStyle newStyle) override;
+
+	virtual void addEntries(QList<LibraryEntryRef> items) override;
+	virtual void addEntry(LibraryEntryRef item) override;
+	virtual void removeEntry(LibraryEntryRef item) override;
+
+	virtual QWidget* widget() { return m_widget; }
```
endrift: Needs `override`.

src/platform/qt/library/LibraryTree.cpp
lioncash: Little late, but in a non-Qt class, you can use `Q_DECLARE_TR_FUNCTIONS(ClassName);` in the class body definition to allow for the regular `tr()` shorthand (see [Writing Source Code For Translation](https://doc.qt.io/qt-5/i18n-source-translation.html))

PR #551, opened 2017-03-29
qt: Don't link against SDLMAIN, Qt already handles that.

waddlesplash:
Fixes SDL+MSVC builds.

Broken out from #549. Just did it on GitHub, so you'll have to squash-merge instead of rebase-merge.


endrift:
Is it not necessary to call SDL_SetMainReady?

waddlesplash:
Hmm. SDL audio worked like this, but it appears that I do need to be doing something more. Will investigate.

endrift:
So why exactly is this needed? How is MSVC+SDL broken?

waddlesplash:
GCC on Windows allows main() as an entry point for GUI apps, and so qtmain doesn't do anything, and SDLMain is ok. MSVC does not allow this, and so qtmain exists and is the WinMain entry point, but that fails to link because SDL renamed the main function. This stops SDL from renaming the main function and makes the entry point qtmain again.

PR #554, opened 2017-03-30
 	Qt: Show native directory separators in the GUI.

### LINKED ISSUE #552: '&'s in "Recent Files" menu are not escaped
If you have a game file with a `&` in it (e.g. "Mario & Luigi: Superstar Saga"), it won't appear in the Recent Files menu, because Qt interprets it as a shortcut marker.

(I'll fix this after the Library pr is done.)

waddlesplash:
The core still gets '/'s always (it chokes on '\'s), but the Qt
interface always uses the native separators.

In the process of doing this, also removed the custom FileDialog
subclass and made everything use GBAApp::get*FileDialog instead.

Also fixes #552, because I had to change that code anyway.

### DIFF
--- a/src/platform/qt/GBAApp.cpp
+++ b/src/platform/qt/GBAApp.cpp
@@ -163,7 +163,7 @@ QString GBAApp::getOpenFileName(QWidget* owner, const QString& title, const QStr
 	QString filename = QFileDialog::getOpenFileName(owner, title, m_configController.getOption("lastDirectory"), filter);
 	continueAll(paused);
 	if (!filename.isEmpty()) {
-		m_configController.setOption("lastDirectory", QFileInfo(filename).dir().path());
+		m_configController.setOption("lastDirectory", QFileInfo(filename).dir().canonicalPath());
 	}
 	return filename;
 }
@@ -174,7 +174,7 @@ QString GBAApp::getSaveFileName(QWidget* owner, const QString& title, const QStr
 	QString filename = QFileDialog::getSaveFileName(owner, title, m_configController.getOption("lastDirectory"), filter);
 	continueAll(paused);
 	if (!filename.isEmpty()) {
-		m_configController.setOption("lastDirectory", QFileInfo(filename).dir().path());
+		m_configController.setOption("lastDirectory", QFileInfo(filename).dir().canonicalPath());
 	}
 	return filename;
 }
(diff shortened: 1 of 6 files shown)


endrift:
Have you tested this to make sure that savestates, screenshots, loading from the command line, etc all still work? There are a bunch of edge cases.

waddlesplash:
Fixed some issues and retested. Savestates, screenshots, loading from command line, loading from Open menu, and loading from gamelist all confirmed to work.

src/platform/qt/GBAApp.cpp
```diff
@@ -163,9 +163,9 @@ QString GBAApp::getOpenFileName(QWidget* owner, const QString& title, const QStr
 	QString filename = QFileDialog::getOpenFileName(owner, title, m_configController.getOption("lastDirectory"), filter);
 	continueAll(paused);
 	if (!filename.isEmpty()) {
-		m_configController.setOption("lastDirectory", QFileInfo(filename).dir().path());
+		m_configController.setOption("lastDirectory", QFileInfo(filename).dir().canonicalPath());
 	}
-	return filename;
+	return QDir::toNativeSeparators(QFileInfo(filename).canonicalFilePath());
```
endrift: Why are you returning a native separator filename to the program if this is used for internal purposes, not user feedback?

src/platform/qt/GBAApp.cpp
```diff
@@ -174,9 +174,9 @@ QString GBAApp::getSaveFileName(QWidget* owner, const QString& title, const QStr
 	QString filename = QFileDialog::getSaveFileName(owner, title, m_configController.getOption("lastDirectory"), filter);
 	continueAll(paused);
 	if (!filename.isEmpty()) {
-		m_configController.setOption("lastDirectory", QFileInfo(filename).dir().path());
+		m_configController.setOption("lastDirectory", QFileInfo(filename).dir().canonicalPath());
 	}
-	return filename;
+	return QDir::toNativeSeparators(QFileInfo(filename).canonicalFilePath());
```
endrift: Ditto

src/platform/qt/GBAApp.cpp
waddlesplash: Hm. I suppose at this point I should change these back.

Issue #550, opened 2017-03-29
Software breakpoints don't do anything if there aren't any hardware breakpoints


jeremyherbert:
Hi,

`ARMDebuggerCheckBreakpoints`@`/src/arm/debugger/debugger.c:35` checks to see if there are any hardware breakpoints pending, and then returns immediately if there are none. This causes it to not check if any software breakpoints are pending. 

I'm trying to work out a patch to fix this, but I'm struggling to understand the difference between hardware and software breakpoints in this emulator. Why are both needed? What is the difference between them? In my understanding, the only reason software breakpoints exist in debuggers is because on real hardware you can't have an unlimited number of breakpoints. But in the case of an emulator, this isn't an issue, because we are running a simulation of the CPU anyway, so we can break it at any point (PC address/memory access/register change, etc).

Thanks

endrift:
Have you actually verified that this is broken by testing it instead of looking at the source? Because software breakpoints go through an entirely different codepath. They insert BKPT instructions into memory, which then cause the emulation to trap when they're hit. This bubbles up into GBABreakpoint, which does appear to enter the debugger. This codepath never even hits ARMDebuggerCheckBreakpoints.

The reason these are separate is because hardware breakpoints slow down the emulation. It has to check EVERY instruction if it's hit a breakpoint. There are ways to optimize this that I haven't implemented (e.g. a bloom filter), but it will never be as fast as simply not checking every instruction. That's where the software breakpoints come in: we don't have to check for the breakpoints unless one actually gets hit. But there are disadvantages too: the memory can be overwritten or read, so it isn't a pure simulation anymore. Quite often, however, it can suffice.

jeremyherbert:
Yes, I have verified this at runtime. My software break points do not appear to trigger.

jeremyherbert:
Also thank you for the explanation about the difference between the two; that makes sense.

endrift:
Weird, they work fine for me with GDB. There's a bug that can cause it to crash on ROM addresses that I just fixed (but haven't pushed yet), but they're definitely getting hit normally.

jeremyherbert:
Ok thanks for the confirmation. I'll keep digging on this after work today.

jeremyherbert:
Well I've been playing with this for a bit, but I still can't get it to go. Might be some issue with my setup, but I will report my progress here anyway as sort of notes to myself. I'm using IDA 6.8 as the debugger.

[...]

PR #555, opened 2017-04-01
Add warning if the user is trying to set a software breakpoint in the BIOS region

jeremyherbert:
Related to https://github.com/mgba-emu/mgba/issues/550

This adds a warning message indicating that you can't have a software breakpoint in the BIOS region. It also fixes a related issue: there appears to be a complete implementation of a log level flag, and so I added it to the help text which prints out when running mgba from the command line.

### DIFF
--- a/src/arm/debugger/debugger.c
+++ b/src/arm/debugger/debugger.c
@@ -126,6 +126,10 @@ bool ARMDebuggerSetSoftwareBreakpoint(struct mDebuggerPlatform* d, uint32_t addr
 		return false;
 	}
 
+	if (address >= 0x3FFF) {
+		mLOG(DEBUGGER, WARN, "Software breakpoints are not supported in the BIOS region. Ignoring software breakpoint at 0x%X.", address);
+	}
+
 	struct ARMDebugBreakpoint* breakpoint = ARMDebugBreakpointListAppend(&debugger->swBreakpoints);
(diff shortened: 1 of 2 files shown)


src/arm/debugger/debugger.c
```diff
@@ -126,6 +126,10 @@ bool ARMDebuggerSetSoftwareBreakpoint(struct mDebuggerPlatform* d, uint32_t addr
 		return false;
 	}
 
+	if (address >= 0x3FFF) {
```
endrift: That's GBA specific, but the ARM core is supposed to be generic. This shouldn't be done here. There isn't really a good way to do this, though, without changing a bunch of stuff to pipe it up into the GBA-specific code.

src/feature/commandline.c
```diff
@@ -31,6 +31,7 @@ static const struct option _options[] = {
 	{ "bios",      required_argument, 0, 'b' },
 	{ "cheats",    required_argument, 0, 'c' },
 	{ "frameskip", required_argument, 0, 's' },
+	{ "loglevel",  required_argument, 0, 'l' },
```
endrift: This should be in a separate PR.

endrift:
Closing due to lack of updates. Will reopen if issues are addressed, and the patch is rebased, but I highly doubt there's a non-invasive way to address this.

PR #558, opened 2017-04-02
SDL: Fix build on MSVC.

waddlesplash:
Crashes on exit (but @endrift said that she knew why that was ...)


src/platform/sdl/main.c
```diff
@@ -27,7 +27,9 @@
 
 #include <errno.h>
 #include <signal.h>
+#ifndef _MSC_VER
 #include <sys/time.h>
```
endrift: It's fine to remove this include entirely. The right one should be in common.h now anyway

src/platform/sdl/CMakeLists.txt
```diff
@@ -6,7 +6,6 @@ if (SDL_VERSION EQUAL "2")
 	if (SDL2_FOUND)
 		set(SDL_INCLUDE_DIR ${SDL2_INCLUDE_DIRS})
 		set(SDL_LIBRARY ${SDL2_LIBRARIES})
-		set(SDLMAIN_LIBRARY "")
```
endrift: I think this might break on some OSes.

waddlesplash: Which OSes are those? Looks like the Travis build succeeded.

waddlesplash:
Addressed comment.

src/platform/sdl/CMakeLists.txt
endrift: I'm...not sure offhand. Why is removing it needed?

waddlesplash: Because SDLMain is the equivalent of QtMain for handling WinMain() on Windows and forwarding it to the actual main(). Until #551, Qt was linked to both (which works everywhere but Windows, because while SDLmain is sometimes used elsewhere, Qtmain is only used on Win32+MSVC). So by removing this line, the SDL backend will be linked to it again. 

You can see the platforms for which this is the case in SDL_main.h: https://github.com/spurious/SDL-mirror/blob/13666b37a08a78bb209b5c584f74846f7e7b5cf2/include/SDL_main.h#L33

PR #660, opened 2017-04-14
Fixed viewport and translucent polygon rendering

Gericom:
I was not able to fix blending of 3d with 2d though


endrift:
Please don't put two distinct fixes in the same commit, much less the same PR.

endrift:
I committed the first half of this patch as 9dfcef3f45efe580e252ec8bc3852257e9de9fd6.

PR #719, opened 2017-04-28
Qt: Fix minimal build

Anty-Lemon:


src/platform/qt/GameController.cpp
```diff
@@ -622,8 +622,9 @@ void GameController::closeGame() {
 	if (!m_gameOpen) {
 		return;
 	}
-
+#ifdef USE_GDB_STUB
 	setDebugger(nullptr);
```
endrift: If `setDebugger` isn't present when `USE_GDB_STUB` isn't defined, that's the real problem; the CLI debugger should still function without the GDB stub.

src/platform/qt/ObjView.cpp
```diff
@@ -51,7 +51,11 @@ ObjView::ObjView(GameController* controller, QWidget* parent)
 	connect(m_ui.magnification, static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged), [this]() {
 		updateTiles(true);
 	});
+#ifdef USE_PNG
 	connect(m_ui.exportButton, SIGNAL(clicked()), this, SLOT(exportObj()));
+#else
+	m_ui.exportButton->setEnabled(false);
```
endrift: Probably a better idea to hide the button instead so people aren't confused as to why it can't be clicked in their build.

PR #729, opened 2017-05-06
Fix sprite window crash; fix poor handling of multiple gamepads in Qt

eevee:
Increasingly verbose explanations included in the individual commits.

### DIFF
--- a/src/core/input.c
+++ b/src/core/input.c
@@ -11,7 +11,7 @@
 
 #include <inttypes.h>
 
-#define SECTION_NAME_MAX 128
+#define SECTION_NAME_MAX 50
 #define KEY_NAME_MAX 32
 #define KEY_VALUE_MAX 16
 #define AXIS_INFO_MAX 12
--- a/src/platform/qt/GBAKeyEditor.cpp
+++ b/src/platform/qt/GBAKeyEditor.cpp
@@ -396,14 +396,14 @@ void GBAKeyEditor::updateJoysticks() {
 	m_controller->updateJoysticks();
 	m_controller->recalibrateAxes();
 
+	// Block the currentIndexChanged signal while rearranging the combo box
+	auto wasBlocked = m_profileSelect->blockSignals(true);
 	m_profileSelect->clear();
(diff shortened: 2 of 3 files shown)


src/core/input.c
```diff
@@ -11,7 +11,7 @@
 
 #include <inttypes.h>
 
-#define SECTION_NAME_MAX 128
+#define SECTION_NAME_MAX 50
```
endrift: Does inih export a preprocessor define for this? If so, it might be a good idea to use that. Might pose an issue with includes though.

src/platform/qt/GBAKeyEditor.cpp
```diff
@@ -396,14 +396,14 @@ void GBAKeyEditor::updateJoysticks() {
...
 	m_profileSelect->addItems(m_controller->connectedGamepads(m_type));
 	int activeGamepad = m_controller->gamepad(m_type);
+	m_profileSelect->setCurrentIndex(activeGamepad);
+	m_profileSelect->blockSignals(wasBlocked);
+
 	selectGamepad(activeGamepad);
-	if (activeGamepad > 0) {
-		m_profileSelect->setCurrentIndex(activeGamepad);
-	}
-	lookupAxes(m_controller->map());
```
endrift: Can you actually remove these? This is needed to update the axis and hat mapping, so it might get out of date if the mapping changes (e.g. due to switching which controller you're using)

src/core/input.c
eevee: Alas, `MAX_SECTION` is only defined in `ini.c`.

src/platform/qt/GBAKeyEditor.cpp
eevee: The last thing `selectGamepad` does is call `refresh`, and the last thing `refresh` does is call `lookupAxes` and `lookupHats`, so these should be 100% redundant.

endrift:
So, some of these changes will be clobbered by my input revamp branch, but I can worry about that on my end.

PR #736, opened 2017-05-15
Core Input: Correct a memset size within mInputUnbindHat()

lioncash:
Previously the binding struct wouldn't be fully invalidated.

Do you want me to add an entry to the CHANGES doc for this? (guidelines don't say whether or not to update them).


endrift:
CHANGES only applies to fixes since the previous stable. This code hasn't reached a stable yet, so don't update it.

I really need to update the guidelines.

PR #740, opened 2017-05-16
Qt: Various cleanups and bugfixes.

waddlesplash:
See individual commits.

### DIFF
--- a/src/platform/qt/DisplayGL.cpp
+++ b/src/platform/qt/DisplayGL.cpp
@@ -5,6 +5,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 #include "DisplayGL.h"
 
+#if defined(BUILD_GL) || defined(BUILD_GLES)
+
 #include <QApplication>
 #include <QResizeEvent>
 #include <QTimer>
@@ -483,3 +485,5 @@ void PainterGL::clearShaders() {
 VideoShader* PainterGL::shaders() {
 	return &m_shader;
 }
+
+#endif
--- a/CHANGES
+++ b/CHANGES
@@ -55,6 +55,7 @@ Bugfixes:
  - Test: Fix crash when loading invalid file
  - GBA Hardware: Fix crash if a savestate lies about game hardware
  - Test: Fix crash when fuzzing fails to load a file
+ - Qt: Disable "New multiplayer window" when MAX_GBAS is reached (fixes mgba.io/i/107)
 Misc:
  - SDL: Remove scancode key input
  - GBA Video: Clean up unused timers
@@ -113,6 +114,7 @@ Misc:
  - VFS: Add VFileFIFO for operating on circle buffers
  - Core: Move rewind diffing to its own thread
  - Util: Tune patch-fast extent sizes
+ - Qt: Relax hard dependency on OpenGL
 
 0.5.2: (2016-12-31)
 Bugfixes:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
-cmake_minimum_required(VERSION 2.6)
-project(mGBA C)
+cmake_minimum_required(VERSION 2.8.11)
+project(mGBA)
 set(BINARY_NAME mgba CACHE INTERNAL "Name of output binaries")
 if(NOT MSVC)
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wno-missing-field-initializers -std=c99")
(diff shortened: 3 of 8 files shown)


src/platform/qt/AssetView.cpp
```diff
@@ -21,7 +21,7 @@ AssetView::AssetView(GameController* controller, QWidget* parent)
 	connect(&m_updateTimer, SIGNAL(timeout()), this, SLOT(updateTiles()));
 
 	connect(m_controller, &GameController::frameAvailable, &m_updateTimer,
-	        static_cast<void(QTimer::*)()>(&QTimer::start));
+	        static_cast<void(QTimer::*)()>(&QTimer::start), Qt::QueuedConnection);
```
endrift: I highly doubt this will matter. Since it's a cross-thread call the default is already QueuedConnection.

src/platform/qt/GameController.cpp
```diff
@@ -144,7 +144,7 @@ GameController::GameController(QObject* parent)
 		unsigned width, height;
 		controller->m_threadContext.core->desiredVideoDimensions(controller->m_threadContext.core, &width, &height);
 		memset(controller->m_frontBuffer, 0xFF, width * height * BYTES_PER_PIXEL);
-		QMetaObject::invokeMethod(controller, "frameAvailable", Q_ARG(const uint32_t*, controller->m_frontBuffer));
+		emit controller->frameAvailable(controller->m_frontBuffer);
```
endrift: Can't do this. The object is owned by a different thread.

src/platform/qt/GameController.cpp
```diff
@@ -172,7 +172,7 @@ GameController::GameController(QObject* parent)
 		unsigned width, height;
 		controller->m_threadContext.core->desiredVideoDimensions(controller->m_threadContext.core, &width, &height);
 		memcpy(controller->m_frontBuffer, controller->m_drawContext, width * height * BYTES_PER_PIXEL);
-		QMetaObject::invokeMethod(controller, "frameAvailable", Q_ARG(const uint32_t*, controller->m_frontBuffer));
+		emit controller->frameAvailable(controller->m_frontBuffer);
```
endrift: Ditto.

src/platform/qt/GameController.cpp
```diff
@@ -975,8 +975,8 @@ void GameController::loadBackupState() {
 		controller->m_backupLoadState->seek(controller->m_backupLoadState, 0, SEEK_SET);
 		if (mCoreLoadStateNamed(context->core, controller->m_backupLoadState, controller->m_loadStateFlags)) {
 			mLOG(STATUS, INFO, "Undid state load");
-			controller->frameAvailable(controller->m_drawContext);
-			controller->stateLoaded(context);
+			emit controller->frameAvailable(controller->m_drawContext);
+			emit controller->stateLoaded(context);
```
endrift: Ditto.

src/platform/qt/DisplayGL.cpp
```diff
@@ -5,6 +5,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 #include "DisplayGL.h"
 
+#if defined(BUILD_GL) || defined(BUILD_GLES)
```
endrift: Are you sure these are sufficient?

waddlesplash: Yes, I did a GL-less build and it worked.

src/platform/qt/GameController.cpp
waddlesplash: I'll just get rid of that commit altogether I guess.

endrift:
Also make sure to update CHANGES as needed.

waddlesplash:
@endrift Done.

endrift:
Please stop lumping more and more unrelated changes into the same PR.

Also that new change makes it so you can't compile mGBA without a C++ compiler installed. Probably not an issue, but...it is a behavior change.

waddlesplash:
> Please stop lumping more and more unrelated changes into the same PR.

You know that you can review one commit at a time on GitHub, right? I figured as these are all small changes that this was the easiest way to do it, and then just rebase-merge at the end.

And there is "Various" in the PR title. :-P

> Also that new change makes it so you can't compile mGBA without a C++ compiler installed. Probably not an issue, but...it is a behavior change.

Wait, really? I didn't know that's what CMake did in the absence of a C++ compiler. I don't know of any systems where there isn't one, though... I'll look into it.

waddlesplash:
It looks like you can put a variable reference in the `project` call, e.g.
```
project(mGBA ${C_LANGUAGE_ONLY})
```
and then allow users to specify `-DC_LANGUAGE_ONLY=C` on the command line. Is that OK?

endrift:
Oh god no that's totally wrong. The right way would be to check if BUILD_QT is set, but that's more effort than not caring, which is where I'm leaning right now.

waddlesplash:
OK sounds fine then.

PR #741, opened 2017-05-17
Qt: Resolve -Wreorder warnings

lioncash:
Resolves some trivial -Wreorder warnings. I can also expand the scope of this PR to use C++11 in-class initialization where (reasonably) possible in the Qt code for cases where members are initialized with constant literal values, if that's more preferable (which has a benefit of getting rid of ifdefs in cpp file initializer lists in some cases).


endrift:
Wait, is -Wreorder not in -Werror -Wextra?

waddlesplash:
Using C++11 in-class initialization for anything set to a literal sounds like a good idea to me.

endrift:
I didn't realize C++11 added that; I am very strongly in favor of using it when possible.

lioncash:
> Wait, is -Wreorder not in -Werror -Wextra?

Now that you say that I was curious, I checked GCC's [diagnostics](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) page and it turns out -Wall enables it, but in clang's case it's only enabled with [-Wmost](https://github.com/llvm-mirror/clang/blob/master/include/clang/Basic/DiagnosticGroups.td#L684) (or explicitly declaring it).

> I didn't realize C++11 added that; I am very strongly in favor of using it when possible.

Alright, cool, I'll update the PR in a bit then.

endrift:
s/-Werror/-Wall/ that is, but uh, it looks like -Wall includes -Wmost. https://github.com/llvm-mirror/clang/blob/master/include/clang/Basic/DiagnosticGroups.td#L725 Weird.

lioncash:
> but uh, it looks like -Wall includes -Wmost. 

Whoops, I missed that, my bad. That's even weirder, since I don't get warnings about it when building on macOS.

lioncash:
Alright, I did a pass over the Qt stuff, let me know if there's anything out of place or if you want other members to have initializers in the class body that didn't have them in their class' initializer list whenever you have the time to review it.

I kept the reorder change commit and the in-class initialization change commit separate for the time being to make it nicer to review, but I can squash them if that's preferable after addressing anything that may come up.

src/platform/qt/GameController.cpp
```diff
@@ -39,41 +39,11 @@ using namespace std;
...
-	, m_wasPaused(false)
-	, m_audioChannels()
-	, m_videoLayers()
-	, m_autofire{}
-	, m_autofireStatus{}
-	, m_stateSlot(1)
-	, m_backupLoadState(nullptr)
-	, m_backupSaveState(nullptr)
 	, m_saveStateFlags(SAVESTATE_SCREENSHOT | SAVESTATE_SAVEDATA | SAVESTATE_CHEATS | SAVESTATE_RTC)
 	, m_loadStateFlags(SAVESTATE_SCREENSHOT | SAVESTATE_RTC)
```
endrift: Any reason these weren't moved? Those are C enum values, so I'd imagine you can use them as-is.

src/platform/qt/GameController.cpp
```diff
@@ -39,41 +39,11 @@ using namespace std;
...
-	, m_logLevels(0)
-	, m_gameOpen(false)
-	, m_vf(nullptr)
-	, m_useBios(false)
-	, m_override(nullptr)
 	, m_audioProcessor(AudioProcessor::create())
-	, m_pauseAfterFrame(false)
-	, m_sync(true)
 	, m_videoSync(VIDEO_SYNC)
 	, m_audioSync(AUDIO_SYNC)
```
endrift: Any reason these weren't moved? Those are constexpr (IIRC).

src/platform/qt/KeyEditor.cpp
```diff
@@ -16,10 +16,6 @@ using namespace QGBA;
 
 KeyEditor::KeyEditor(QWidget* parent)
 	: QLineEdit(parent)
-	, m_key(-1)
-	, m_axis(-1)
-	, m_hat(-1)
-	, m_button(false)
 	, m_direction(GamepadAxisEvent::NEUTRAL)
 	, m_hatDirection(GamepadHatEvent::CENTER)
```
endrift: Enums again, IIRC, although in a different h file, so that might cause include woes.

src/platform/qt/ShortcutController.cpp
```diff
@@ -555,8 +552,8 @@ ShortcutController::ShortcutItem::ShortcutItem(QAction* action, const QString& n
 ShortcutController::ShortcutItem::ShortcutItem(ShortcutController::ShortcutItem::Functions functions, int shortcut, const QString& visibleName, const QString& name, ShortcutItem* parent)
 	: m_action(nullptr)
```
lioncash: I'm a dingus and missed these (will address alongside everything else that gets flagged down, if any).

src/platform/qt/GameController.cpp
lioncash: These would require including the serialization header in the header file and wasn't sure if you would want that (more files need to recompile if a change is made in the serialization header), rather than one cpp file. This was also my reasoning for the ones below as well. Do you still want me to move them there anyway?

src/platform/qt/GameController.cpp
lioncash: Good catch, I'll move these over.

src/platform/qt/GameController.cpp
endrift: Yeah, as I said in the next comment, I realized your reasoning. Let's leave this one be.

lioncash:
Whoops, forgot to mention that I've resolved everything that was pointed out; sorry about that. (I forget if Github even sends notifications via email on new force pushes, in which case ignore this if it does).

PR #745, opened 2017-06-01
New shaders

Iniquitatis:
Added simplest possible configurable shaders.

(I think it will be more useful in this repository than just on my HDD. :))

### DIFF
--- a/res/shaders/lcd.shader/lcd.fs
+++ b/res/shaders/lcd.shader/lcd.fs
@@ -1,0 +1,40 @@
+/*
+   LCD Shader
+
+   Copyright (C) 2017 Dominus Iniquitatis - zerosaiko@gmail.com
+
+   Permission is hereby granted, free of charge, to any person obtaining a copy
+   of this software and associated documentation files (the "Software"), to deal
+   in the Software without restriction, including without limitation the rights
+   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+   copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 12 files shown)


endrift:
Please use e.g. -2 for dimensions to be 2x input. The syntax needs some refreshing, but this avoids the problem of having the shaders be broken for GB. I see I forgot to do this in my shaders though.

Iniquitatis:
Thanks for review.

Iniquitatis:
(Closed again by accident, sorry. I'm writing from mobile browser.)

endrift:
For height/width. Integer scaling just constrains the dimensions of the frame buffer to any integer multiple instead of a specific size, IIRC. You may need to experiment a bit.

Iniquitatis:
Done and tested. Do I need to fix anything else?

endrift:
I'll take a closer look when I get a chance but you're probably fine. Won't be able to merge for a few hours though.

Iniquitatis:
Fixed capital letters in several names and squashed all existing commits.

Iniquitatis:
Fixed scanlines scaling and squashed last commits again. Sorry for inconvenience.

endrift:
Sorry for the delay; I've been really busy the past few days. I'll try to get to this tonight or tomorrow. I don't expect there to be any further hiccups that would require revisions on your end though.

Iniquitatis:
Ah, it's okay. I'm away from my PC (actually, from stable internet connection) for another week, so it's highly unlikely I can fix anything here for now.

PR #771, opened 2017-06-30
Update CHANGES

theheroGAC:
Qt: Italian translation (by theheroGAC)


endrift:
What are you doing?? This is still the wrong PR.

PR #785, opened 2017-07-08
Qt: Actually load the libraryStyle setting.

waddlesplash:
No idea how I missed this in the library PR.


src/platform/qt/SettingsView.cpp
```diff
@@ -322,6 +322,7 @@ void SettingsView::reloadConfig() {
 	loadSetting("savestatePath", m_ui.savestatePath);
 	loadSetting("screenshotPath", m_ui.screenshotPath);
 	loadSetting("patchPath", m_ui.patchPath);
+	m_ui.libraryStyle->setCurrentIndex(loadSetting("libraryStyle").toInt());
```
endrift: Please move this to the bottom of the function (or at least outside of this block).

PR #818, opened 2017-07-21
Qt: Redo GameController into multiple classes

endrift:
Some big changes in this PR:
- GameController is gone
- CoreManager spawns cores as CoreController which are then given to the Window
- CoreControllers are now shared_ptrs preventing any additional windows that are open and have references to it will prevent the shutdown until they can all clean up
- Window creates the shared_ptr but doesn't necessarily have a CoreController at all times
- Display driver reloading now works without restarting the emulator

Some big caveats:
- Due to the use of variadic templates I bumped the build to C++14, so I'm using make_unique now too. I will likely be dropping support for Trusty and g++ <5 as a result.
- ~~I haven't tested multiplayer yet, it could be super broken~~
- ~~I haven't tested that all of the dialogs still work yet~~
  - ~~I think debuggers are broken right now~~
  - ~~Games that crash may not work currently~~
  - ~~Haven't tested unimplemented BIOS stuff~~
- ~~I think I forgot to test BIOS booting~~
- ~~DisplayQt seems to be a bit buggy but I'm not sure that's a regression or not~~ (it was)
- ~~Saving configs is very slow at the moment~~
- ~~Rewind seems to be broken~~
- ~~Replace ROM seems to be broken~~
- ~~Sync settings are messed up on first load~~
- ~~Loading a patch before a game has loaded is broken~~
- ~~It sometimes crashes on first game load~~
- ~~I need to put back log filtering~~

~~It's not hugely well-tested at the moment but it seems to be working so far.~~


src/platform/qt/CoreManager.cpp
```diff
@@ -0,0 +1,130 @@
...
+	}
+	mCoreConfigSetOverrideIntValue(&core->config, "useBios", 1);
+	mCoreConfigSetOverrideIntValue(&core->config, "skipBios", 0);
+
+	CoreController* cc = new CoreController(core);
+	if (m_multiplayer) {
+		cc->setMultiplayerController(m_multiplayer);
+	}
+	emit coreLoaded(cc);
+	return cc;}
```
waddlesplash: wardrobe ... er, style malfunction?

waddlesplash:
I did a rapid skim-through of the code; the changes make sense, I suppose. I'm not entirely clear on how the CoreManager operates or who owns it (but again I didn't read through that).

src/platform/qt/CoreManager.cpp
endrift: Whoops, not sure what happened there.

endrift:
CoreManager is owned by the GBAApp currently, although there's nothing stopping it from being somewhere else. Once the Window split is done it might move somewhere else.

endrift:
PR is ready for a full review.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.11)
+cmake_minimum_required(VERSION 3.1)
```
waddlesplash: 👍

src/platform/qt/AudioProcessorSDL.cpp
```diff
@@ -16,18 +16,19 @@ AudioProcessorSDL::AudioProcessorSDL(QObject* parent)
...
-}
-
-void AudioProcessorSDL::setInput(mCoreThread* input) {
+void AudioProcessorSDL::setInput(std::shared_ptr<CoreController> input) {
 	AudioProcessor::setInput(input);
-	if (m_audio.core && input->core != m_audio.core) {
+	if (m_audio.core && input->thread()->core != m_audio.core) {
 		mSDLDeinitAudio(&m_audio);
-		mSDLInitAudio(&m_audio, input);
+		mSDLInitAudio(&m_audio, input->thread());
```
waddlesplash: Sometimes you initialize with `input()` and sometimes with `input->thread()`. Am I missing something, or is that a mistake? (If not, maybe deserves a comment?)

src/platform/qt/CoreController.cpp
```diff
@@ -0,0 +1,725 @@
...
+	}
+
+	mCoreThreadJoin(&m_threadContext);
+
+	mCoreConfigDeinit(&m_threadContext.core->config);
+	m_threadContext.core->deinit(m_threadContext.core);
+}
+
+color_t* CoreController::drawContext() {
+	QMutexLocker locker(&m_mutex);
```
waddlesplash: If the mutex here is protecting `m_completeBuffer`, perhaps you could use a `QReadWriteLocker`? Or would that be overkill, as there's never going to be more than one reader?

src/platform/qt/AudioProcessorSDL.cpp
endrift: `input` here is a variable shadowing the `input()` method, which returns the `->thread()` everywhere else. I should probably rename it so it doesn't shadow anymore.

src/platform/qt/CoreController.cpp
endrift: Yeah, so what's going on here is that `m_completeBuffer` is a rudimentary double-buffer--it's swapped between two objects in a different thread so here we just want one of them. The `data()` should remain valid since they are never resized after the initial sizing (in the constructor). Although that may change in the future (e.g. when I add SGB support, hm). That said, there are only two threads it'll ever be accessed in, so I think that's overkill.

src/platform/qt/AudioProcessorSDL.cpp
endrift: Fixed

probonopd:
> Due to the use of variadic templates I bumped the build to C++14, so I'm using make_unique now too. I will likely be dropping support for Trusty and g++ <5 as a result.

Bad decision. As a result, I am unable to build on trusty.

https://travis-ci.org/probonopd/mgba/jobs/423678980#L1147

As a result, the application won't run on all still-supported versions of Ubuntu.

As a result, I won't be making an AppImage.

PR #848, opened 2017-08-05
GBA: Fix MSR when execution mode changes to Thumb

MerryMage:
Testcases available.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -29,6 +29,7 @@ Bugfixes:
  - GB, GBA Savedata: Fix savestate-related save overwriting (fixes mgba.io/i/834)
  - Qt: Fix timezone issues with time overrides
  - Qt: Fix sprite export pausing game indefinitely (fixes mgba.io/i/841)
+ - ARM: Fix MSR when T bit is set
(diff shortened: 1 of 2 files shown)


endrift:
I'm confused by the presence of the nop. How does that affect timing?

MerryMage:
Here are a couple of testcases: https://gist.github.com/MerryMage/529d888834963d9128bab6156fe33b97

When CPSR.T is set while you're in ARM state, the following happens: (Take $ to be the address of the MSR instruction)

1. The instruction currently being decoded ($+4) is discarded. This wastes a cycle.
2. The low-word of the instruction already fetched ($+8) is interpreted as normal.
3. Execution continues as normal from ($+12).

Note that this means that any thumb instructions that may reside at ($+4), ($+6), ($+10) do not get executed.

endrift:
Is there a source on that info? I can look into it a bit but merging this isn't dependent on that.

endrift:
Can you add an entry to the changelog?

MerryMage:
Added.

PR #852, opened 2017-08-09
Qt: Update German GUI translation

rootfather:
Some Game Boy Printer related stuff is missing, I need more reference on that. What exactly does "Hurry up!" and "Tear off" means? Are those button labels on the printer itself or some software commands?


vaguerant:
The Game Boy Printer emulation attempts to maintain some of the "analog" feel of a real Game Boy Printer by having the printing take a while ("Hurry up!" skips the cosmetic 'printing' process and displays the finished print) and allowing the user to tear the printed paper away ("Tear off" removes the paper after printing). These aren't functions that existed on the real printer, they're just user interface features.

rootfather:
Thank you @vaguerant for your help!

@endrift, the translation is now complete besides the IOViewer stuff (as always). I really have to dig to the internals to understand the IOViewer stuff and then decide which strings should be left original and which ones should be translated.

Issue #909, opened 2017-10-16
Linking with static build enabled fails on Linux


rootfather:
I'm currently setting up an Arch Linux based build environment for the projects I'm working on. I'm on the very latest mGBA master code and using the latest Arch packages.

When I set -DBUILD_SHARED=ON -DBUILD_STATIC=OFF, linking works fine and produces a working mgba-qt executable. However, if I use -DBUILD_SHARED=OFF -DBUILD_STATIC=ON, linking fails with the following message:

```[ 86%] Linking CXX executable mgba-qt
/usr/bin/ld: ../libmgba.a(rewind.c.o): undefined reference to symbol 'pthread_setname_np@@GLIBC_2.12'
/usr/lib/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: Fehler: ld gab 1 als Ende-Status zurück [ld returned error code 1]
make[2]: *** [qt/CMakeFiles/mgba-qt.dir/build.make:2088: qt/mgba-qt] Fehler 1 [Error 1]
make[1]: *** [CMakeFiles/Makefile2:217: qt/CMakeFiles/mgba-qt.dir/all] Fehler 2 [Error 2]
make: *** [Makefile:152: all] Fehler 2 [Error 2]
```
Seems somewhat weird for me... I don't think it's a missing dependency because otherwise the shared build would be broken too? A quick google search indicated that adding `-lpthread` to the linker arguments would solve this issue.

Unfortunately, I am not familiar enough with CMake to add this LD argument properly.

Please let me know if you need any additional information, e.g. regarding package versions.

rootfather:
I think I found a solution - working on it.

PR #911, opened 2017-10-16
CMake/Linux: Enforce linking with -pthreads for static builds

rootfather:
This should fix https://github.com/mgba-emu/mgba/issues/909.

With this change, I can build a static and shared build of mGBA Qt on my Linux system. I also compiled it successfully with MSYS2. However, I don't know if there is a more elegant solution than this... I think this needs some more testing, especially on the *nix systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,6 +206,10 @@ elseif(UNIX)
 
 	if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
 		add_definitions(-D_GNU_SOURCE)
+
+		if(BUILD_STATIC)
+			link_libraries(-pthread)
+		endif()
 	endif()
 	if(NOT APPLE AND NOT HAIKU)
 		set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread")


endrift:
The very next if block contains code that SHOULD do the same thing, and if it DOESN'T then it should be fixed there,

rootfather:
@endrift Thanks for the hint, I see. Actually, this might be even easier to fix by simply modifying the block to
```
if(NOT APPLE AND NOT HAIKU)
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -lpthread")
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -lpthread")
```
I have to verify this tomorrow.

endrift:
No, that's not correct. -pthread makes sure the right headers and flags are present for compilation phase, and -lpthread for link phase. -lpthread would do nothing at compilation phase as you have there.

rootfather:
I must admit that my knowledge about Linux development is too limitied right now. I'll leave this PR open for reference, but I have to do further investigation before I'm able to fix this.

endrift:
Obsoleted by 4d2f721a8061815f746cfdf345e54a0f926543b4.

PR #914, opened 2017-10-19
Fix PARv3 If-AND code types (fixes #913)

### LINKED ISSUE #913: GBA Cheats: PARv3 If-AND not working correctly
Testing on 523aaf1.

I have the following PARv3 code:
```
38200000 00000004	if 8-bit [02000000] & 4 then 1-line
80200000 000000FC	\   8-bit [02000001] += -4
78200001 00000004	if 8-bit [02000001] & 4 then 2-line
80200001 000000FE	|   8-bit [02000001] += -2
80200001 000000FE	\   8-bit [02000001] += -2
B8200002 00000004	if 8-bit [02000001] & 4 then multi
80200002 000000FF	|   8-bit [02000002] += -1
80200002 000000FF	|   8-bit [02000002] += -1
80200002 000000FF	|   8-bit [02000002] += -1
80200002 000000FF	\   8-bit [02000002] += -1
00000000 40000000	endif
```
Which results in the following encrypted code:
```
12335464 5A5AF80F
C48F1536 905DBDB7
2F3BB1C0 8D0A444B
3A0CDD86 90EA88F7
3A0CDD86 90EA88F7
71119C3F 006CF823
F763DEB2 B8060F53
F763DEB2 B8060F53
F763DEB2 B8060F53
F763DEB2 B8060F53
1E27D83B BBFB6923
```
Expected behavior: when bit 4 is set in the value at 0x02000000 etc., this code unsets it.
Actual behavior: when the value at 0x02000000 etc. is nonzero, this code decrements it by 4. So for instance, if I set the value at 0x02000000 to 01, mGBA will endlessly decrement it on every frame.

Basically it just seems like the AND with the right-hand value isn't being performed correctly.
I've tested the 16-bit and 32-bit code types too and they produce the same behavior.

~Furthermore, the 32-bit version only seems to check the lower 16 bits of the address. For example:~
```
3A200000 00040000	if 32-bit [02000000] & 0x40000 then 1-line
84200000 FFFC0000	\   32-bit [02000000] += -0x40000
```
~Nothing happens if [02000000] == 0xFFFF0000, for instance. (Even if the lower 16 bits of [02000000] are nonzero, strangely enough.)~

Prof9:
Here's a PR that fixes the following two problems with If-AND code types for PARv3 (#913):
* Operand width is not read correctly. 16-bit becomes 8-bit, 32-bit becomes 16-bit.
* AND condition is not checked correctly. Logical AND is used with the operand instead of bitwise AND.


endrift:
Where are you getting information about the operand width? It seems to directly contradict GBATEK and the GameShark Hacking guide.

Prof9:
Actually, I was getting it from [EnHacklopedia](http://doc.kodewerx.org/hacking_gba.html#artypes)... since it has the nicest layout, imo. Looking at it now, it does seem to be wrong, huh. I guess that explains why other emulators also had problems with it. I'll take out the part about the operand width.

PR #916, opened 2017-10-20
All: Add instructions for packaging all .DLLs with ntldd

rootfather:


endrift:
Instructions for using [windeployqt](http://doc.qt.io/qt-5/windows-deployment.html) are better since it's included with Qt already.

rootfather:
Thanks endrift, I wasn't aware of windeployqt. Let me run a few tests, then I'll update this PR accordingly.

rootfather:
Gave it a first shot - it doesn't work for me. It copies all the necessary Qt libraries as expected, but the necessary libraries from the MSYS2/MinGW toolchain are missing, e.g. it's not copying the zlib or the libexpoxy .dlls. I think this would only work if the libraries including the MinGW runtime files are compiled/linked statically.

waddlesplash:
> Gave it a first shot - it doesn't work for me. It copies all the necessary Qt libraries as expected, but the necessary libraries from the MSYS2/MinGW toolchain are missing, e.g. it's not copying the zlib or the libexpoxy .dlls. 

Are these libraries in your PATH? windeployqt actually scans the binary and looks for the libraries it wants, but if these are not in your PATH, it can't help you.

rootfather:
Endrift, have you had the time to look at this again?

endrift:
I'm actually going to subvert adding manual directions to instead add a script that parses ntldd output to do it automatically.

endrift:
btw it's not quite ready yet but it works with installing via cpack too, so you can just do cpack -G ZIP and it'll spit out a zip file with all the dependencies

PR #939, opened 2017-12-02
Doc: Add German translation of the README.md file

rootfather:
This is the German translation of the README.md file. I'll try to update
it as soon as there are any changes made the README file. One step
closer to a full German localization :)


endrift:
I'm not sure how I feel about having a German readme, especially keeping it synced (though you're pretty on the ball about that), and given the filename would have to be different per hypothetical language if we just translate README there could be collisions.

rootfather:
Keeping the German readme synced to upstream won't be an issue on my side. Besides the translation stuff for mGBA, I do the German translation and documentation for the ScummVM project and that's very up-to-date too ;)

My main intention is to ship the translated README file(s) with the upcoming (and hopefully soon merged) Inno installer if the language is set to German. It won't be installed if the language is to any other than German. And having an additional text file in the mGBA distribution archives shouldn't be a big deal?

Instead of just translate the filenames, we could go with something like README.DE or DE_README to avoid any collisions.

endrift:
I think README_DE.md might be the best name.

rootfather:
I renamed the file as you suggested.

rootfather:
Is it possible to convert the README_DE.md file to an .html and start distributing it with the mGBA packages?

Issue #906, opened 2017-10-12
Unintended change of the language interface.


ShalokShalom:
EDIT: The interface was in German on the initial start and changed to English on the second.
I was in the settings menu in between and I did not touch the language settings there.

I think there is a setting, which unintentionally touches the language setting.

rootfather:
What operating system and version of mGBA are you running?

When I use the latest mGBA codebase compiled with an up-to-date MSYS2 MinGW64 toolchain, at least on Windows 10 mGBA automatically detects that I am running the German version of Windows and thus applies the German locale right at the first start without any configuration.

ShalokShalom:
KaOS, current (0.6.1)

ShalokShalom:
To be honest, I think it was in German on the initial start and changed to English then.
I was in the settings menu in between, while I did not touch the language settings there.

endrift:
huh I thought I fixed that bug

ShalokShalom:
Oh, its well known. That is fine. 

I am not insane. 
Not here, at least. :smile:

rootfather:
I re-checked this again and I can actually reproduce this issue on the latest master code.

The language changes back to English as soon as you change *any* setting in the settings dialog. Please note that if you open the Interface tab, you can see that it always falls back to English in the language selection dropdown menu.

This bug doesn't seem to appear when you select the UI language manually in the Interface tab, but only if you don't touch the language settings and rely on the auto detect feature.

ShalokShalom:
Thanks a lot for testing.

rootfather:
Any news on this? :)

PR #941, opened 2017-12-07
Win32: Add Inno Setup script for building a Windows installer

lotharsm:
This PR adds an Inno Setup script to build a Windows installer (fixes [Bug 138](http://mgba.io/b/138)).

## Features
- Prevent installation on pre-Vista systems
- Option to add start menu and desktop shortcuts
- Option to register .gb/.gbc/.gba files with mGBA allowing the user to start those files directly
- ~~Attempt to work around Bug #906 by setting the correct language value in qt.ini depending on the selected language during the installation process instead of relying on the language auto detection. Seems to work pretty well for me, haven't experienced an unintentional language change since then. This workaround will be removed as soon as the bug is fixed.~~
- Automation friendly - no hardcoded paths, simply copy the installer files to the Win32 distribution directory.
- Release mode - when setting the "IsRelease" variable in the setup script to "Yes", the created setup file will be named mGBA-x.y.z-win32.exe while fetching the version information from the mGBA executable. When IsRelease is set to "No", then the filename mGBA-latest-win32.exe will be used.
- Tested on Windows Vista, 7 and 10
- Maintained by me :)

### Demo
As a small demo, here's an installer file for mGBA 0.6.1.
[mGBA-0.6.1-win32.exe.zip](https://github.com/mgba-emu/mgba/files/1540922/mGBA-0.6.1-win32.exe.zip)


endrift:
I'd like to fix the language issue for real before looking closer at this.

endrift:
Now that that's fixed you can remove the workaround 😀

endrift:
So one of the things that's holding me off from looking closer at this is that at some point I may start shipping stripped down builds with the extra features (e.g. GIF and video recording, scripting, etc; the ones most people don't use) as DLLs that can be optionally added. I'm not sure how I'm going to handle that yet though so I'm not sure if it makes sense to merge this yet.

rootfather:
Yeah, maybe you should decide that first. But in general, I see no problem in including all the "Plugin DLLs" inside the installer package and make those plugins optional during the installation process - that's fairly easy to do with Inno.

rootfather:
Hi endrift, is there anything I can do to move this forward and maybe can get this ready for an upcoming 0.7 release?

endrift:
I'll need to try it out first, which I keep forgetting to do.

rootfather:
> ...which I keep forgetting to do.

Friendly reminder... :P

endrift:
Before I forget: Is this updated for the new French translation?

rootfather:
Ah, nope, I have to add the french translation. Update will follow ASAP.

rootfather:
I should be more precise: The only thing what has to be done is to make French available during the installation process. For testing purposes, you can use the current code as-is.

tools/windows/mgba-setup-win32.iss
```diff
@@ -0,0 +1,182 @@
...
+Name: "desktopicon"; Description: "{cm:CreateDesktopIcon}"; GroupDescription: "{cm:AdditionalIcons}"
+Name: "gbfileassoc"; Description: "{cm:AssocFileExtension,mGBA,Game Boy}"; GroupDescription: "{cm:FileAssoc}"
+Name: "gbcfileassoc"; Description: "{cm:AssocFileExtension,mGBA,Game Boy Color}"; GroupDescription: "{cm:FileAssoc}"
+Name: "gbafileassoc"; Description: "{cm:AssocFileExtension,mGBA,Game Boy Advance}"; GroupDescription: "{cm:FileAssoc}"
+
+[Files]
+Source: "mGBA.exe"; DestDir: "{app}"; Flags: ignoreversion
+Source: "CHANGES.txt"; DestDir: "{app}\"; Flags: ignoreversion isreadme
+Source: "LICENSE.txt"; DestDir: "{app}\"; Flags: ignoreversion
+Source: "mGBA.exe"; DestDir: "{app}\"; Flags: ignoreversion
```
endrift: Why is this here twice?

tools/windows/mgba-setup-win32.iss
```diff
@@ -0,0 +1,182 @@
+;For automation purposes it is highly recommended to copy the files from
+;\tools\windows\ to the directory that contains the win32 distribution files!
+
+;Set CurrentReleaseVersion to the number of the latest stable mGBA build.
+#define CurrentReleaseVersion = '0.6.3'
```
endrift: It might make sense to make this a .in file that's processed by CMake so it can get the current version number from the version scripts--unless there's a way to run a command in the .iss folder to get the version from CMake.

tools/windows/mgba-setup-win32.iss
```diff
@@ -0,0 +1,182 @@
...
+  OutputBaseFilename=mGBA-{#AppVer}-win32
+#elif CurrentReleaseVersion != AppVer;
+  #define IsRelease = 'no'
+  AppVerName=mGBA (Development build)
+  OutputBaseFilename=mGBA-setup-latest-win32
+  #endif
+UsePreviousLanguage=False
+DisableWelcomePage=False
+VersionInfoDescription=mGBA is an open-source Game Boy Advance emulator
+VersionInfoCopyright=� 2013�2017 Jeffrey Pfau
```
endrift: 2018.

tools/windows/mgba-setup-win32.iss
```diff
@@ -0,0 +1,182 @@
...
+
+[Files]
+Source: "mGBA.exe"; DestDir: "{app}"; Flags: ignoreversion
+Source: "CHANGES.txt"; DestDir: "{app}\"; Flags: ignoreversion isreadme
+Source: "LICENSE.txt"; DestDir: "{app}\"; Flags: ignoreversion
+Source: "mGBA.exe"; DestDir: "{app}\"; Flags: ignoreversion
+Source: "nointro.dat"; DestDir: "{app}\"; Flags: ignoreversion
+Source: "README.html"; DestDir: "{app}\"; Flags: ignoreversion isreadme; Languages: english italian spanish
+Source: "README_DE.html"; DestDir: "{app}\"; DestName: "LIESMICH.html"; Flags: ignoreversion isreadme; Languages: german                        
+Source: "shaders\agb001.shader\agb001.fs"; DestDir: "{app}\shaders\agb001.shader\"; Flags: ignoreversion
```
endrift: Is there any way to copy a folder wholesale? This looks cumbersome.

endrift:
Can you please update this for the new CMake-generated file layout?

endrift:
The current file layout is pretty much identical to what is created by just building it, so if we could get the installer to be built by a CMake-generated .iss file that would be pretty ideal. Previously there was stuff that the buildbot tossed in after building but I wanted to streamline things a bit.

lotharsm:
I think we now have as much simplification of the script as possible.

endrift:
The version info is still wrong. Apart from that I haven't looked into how I'd integrate it with CMake yet and if it's possible to use CMake to keep the version number in sync.

lotharsm:
Seems that I finally should find some time to look deeper into how the CMake stuff works... might complicate things a bit unfortunately. :(

endrift:
Would it be possible to make an option to have it install the "portable" version that doesn't make the file associations and creates an empty "portable.ini" file instead?

Also .sgb should probably be added now.

lotharsm:
Yep, that's possible. I'll add .sgb to the list of the file associations.

Do I need to fill any values into portable.ini or is it really enough to simply create the file without any content?

PR #943, opened 2017-12-11
Qt: Fix build with Qt 5.10

ilovezfs:
Fixes "MemoryModel.cpp:102:15: error: no viable overloaded '='"


rootfather:
I can confirm that building on Qt 5.10 is currently broken and that this patch fixes this.

ilovezfs:
The failure on Travis's `osx` image is the same as the one I reported here: https://github.com/mgba-emu/mgba/issues/944

PR #988, opened 2018-02-08
Wii: Fix garbage during autoboot

SuperrSonic:
Specifically when autobooting from the system menu via a forwarder, while the ROM is loading a(usually pink) horizontal bar will show up at the bottom of the screen until the game starts.

If I boot the forwarder from the hbc, the horizontal bar never appears but there are still some noticeable green dots all over the screen.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -51,6 +51,7 @@ Bugfixes:
  - GB Timer: Minor accuracy improvements
  - GB Audio: Clock frame events on DIV
  - GBA: Fix SharkPort saves for EEPROM games
+ - Wii: Fix garbage during autoboot
 Misc:
  - GBA Timer: Use global cycles for timers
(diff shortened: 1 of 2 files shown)


endrift:
It look like there may be better ways to go about doing this, though it would involve moving around the VIDEO_SetBlack calls that already exist some.

endrift:
I'm working on a more comprehensive change that will obsolete this patch. It should be done this evening.

SuperrSonic:
The VIDEO_Flush at line 240 is causing my TV to react strangely, it displays a blue screen for a few seconds. Also noticed that in the file browser the text flickers for a moment, this only happens at startup.

endrift:
You mean line 244? That's super strange.

SuperrSonic:
For some reason I checked the line from https://github.com/mgba-emu/mgba/commit/c623970d0feeb61fdfa13c513468907dbd55c0f6 instead, thinking they were the same.

PR #1019, opened 2018-03-24
Libretro: Allow choosing the Game Boy model

dobyrch:
This adds a switch in RetroArch to force the Game Boy model to Game Boy, Game Boy Color, or Super Game Boy.  The default setting is "Autodetect," which preserves the current behavior of selecting the most appropriate model for the current game.

In order to support running DMG games with one of the Game Boy Color's built-in palettes, the RetroArch core will now also attempt to load the BIOS for the selected Game Boy model.

### DIFF
--- a/src/platform/libretro/libretro.c
+++ b/src/platform/libretro/libretro.c
@@ -455,21 +455,52 @@ bool retro_load_game(const struct retro_game_info* game) {
 	core->loadROM(core, rom);
 	core->loadSave(core, save);
 
+	const char* sysDir = 0;
+	const char* biosName = 0;
+	char biosPath[PATH_MAX];
+	environCallback(RETRO_ENVIRONMENT_GET_SYSTEM_DIRECTORY, &sysDir);
+
 #ifdef M_CORE_GBA
 	if (core->platform(core) == PLATFORM_GBA) {
 		core->setPeripheral(core, mPERIPH_GBA_LUMINANCE, &lux);
+		biosName = "gba_bios.bin";
 
-		const char* sysDir = 0;
-		if (core->opts.useBios && environCallback(RETRO_ENVIRONMENT_GET_SYSTEM_DIRECTORY, &sysDir)) {
-			char biosPath[PATH_MAX];
-			snprintf(biosPath, sizeof(biosPath), "%s%s%s", sysDir, PATH_SEP, "gba_bios.bin");
-			struct VFile* bios = VFileOpen(biosPath, O_RDONLY);
-			if (bios) {
-				core->loadBIOS(core, bios, 0);
-			}
-		}
 	}
 #endif
+
+#ifdef M_CORE_GB
+	if (core->platform(core) == PLATFORM_GB) {
+		const char* modelName = mCoreConfigGetValue(&core->config, "gb.model");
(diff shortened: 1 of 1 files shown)


src/platform/libretro/libretro.c
```diff
@@ -63,6 +63,27 @@ static void _reloadSettings(void) {
...
+		} else if (strcmp(var.value, "Super Game Boy") == 0) {
+			model = GB_MODEL_SGB;
+		} else if (strcmp(var.value, "Game Boy Color") == 0) {
+			model = GB_MODEL_CGB;
+		} else {
+			model = GB_MODEL_AUTODETECT;
+		}
+
+		modelName = GBModelToName(model);
+		mCoreConfigSetDefaultValue(&core->config, "gb.model", modelName);
```
endrift: These are three different values for a reason. Unifying them would cause more problems than it would fix.

src/platform/libretro/libretro.c
```diff
@@ -433,22 +455,53 @@ bool retro_load_game(const struct retro_game_info* game) {
...
+
+		switch (gb->model) {
+		case GB_MODEL_DMG:
+			biosName = "gb_bios.bin";
+			break;
+		case GB_MODEL_SGB:
+			biosName = "sgb_bios.bin";
+			break;
+		case GB_MODEL_CGB:
+		default:
```
endrift: Why would this be the default case?

src/platform/libretro/libretro.c
dobyrch: Can you elaborate more on what problems unifying them could cause? I don't want to clutter up the menu too much, so it seems nice having one switch to force the model regardless of whether the game is DMG, CGB, or SGB enhanced.

Since RetroArch supports per-game config files, users can select a model for a specific game and use autodetection for everything else.

src/platform/libretro/libretro.c
dobyrch: The default case should never be hit since only DMG, SGB, CGB, and Autodetect can be selected from the menu (and `GBDetectModel` also only chooses one of those three models).  I only put it there to silence warnings about not handling every case.

Maybe it would be clearer to just leave the default case empty or add an `assert(0)`?

endrift: I don't like the `assert(0)` (use `abort` if anything) but making it be DMG might be more clear.

src/platform/libretro/libretro.c
endrift: If it's per-game it's not a big deal. Mostly the problem lies in having the default be:

- DMG/SGB for CGB games will break a lot of things
- CGB/AGB for DMG games can cause incompatibilities at times (only a handful of games though)
- GCB/AGB for SGB games breaks SGB features

which is why I made it be separate categories. But basically there's no good way to control all three at once without something going wrong along the way.

dobyrch:
@endrift I changed the default case to load the DMG boot ROM, but everything else is the same.  

Considering that default option for the model is "Autodetect," and the model can be configured on a per-game basis, would you be willing to accept this patch in it's current form?

Thanks for the fast reply :)

endrift:
Yes. :)

PR #1021, opened 2018-03-25
GB Video: Fix SGB border transparency

dobyrch:
This fixes an issue with Pokemon Blue's SGB border:

bad good

According to the [Pan Docs](http://problemkaputt.de/pandocs.htm#sgbcolorpalettesoverview), "Color 0 of each of the eight palettes is transparent, causing the backdrop color to be displayed instead".  The output buffer needed to be initialized to white for this border to display correctly, which also has a nice side effect of preventing a border of garbage from briefly displaying before boot:

junk

I updated the libretro version to initialize the output buffer; the Qt version already initializes its buffer to all 1s.  I haven't tried any other platforms, so they may also need to be be updated.

Thanks for taking the time to look at these changes, and thank you for making this wonderful emulator!

### DIFF
--- a/src/gb/renderers/software.c
+++ b/src/gb/renderers/software.c
@@ -75,25 +75,25 @@ static void _regenerateSGBBorder(struct GBVideoSoftwareRenderer* renderer) {
 			tileData[3] = renderer->d.sgbCharRam[(SGBBgAttributesGetTile(mapData) * 16 + localY) * 2 + 0x11];
 
 			size_t base = y * renderer->outputBufferStride + x;
-			int p = SGBBgAttributesGetPalette(mapData) * 0x10;
+			int paletteBase = SGBBgAttributesGetPalette(mapData) * 0x10;
+			int colorSelector;
+
 			if (SGBBgAttributesIsXFlip(mapData)) {
-				renderer->outputBuffer[base + 0] = renderer->palette[p | ((tileData[0] >> 0) & 0x1) | ((tileData[1] << 1) & 0x2) | ((tileData[2] << 2) & 0x4) | ((tileData[3] << 3) & 0x8)];
-				renderer->outputBuffer[base + 1] = renderer->palette[p | ((tileData[0] >> 1) & 0x1) | ((tileData[1] >> 0) & 0x2) | ((tileData[2] << 1) & 0x4) | ((tileData[3] << 2) & 0x8)];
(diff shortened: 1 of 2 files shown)


src/gb/renderers/software.c
```diff
@@ -76,25 +76,18 @@ static void _regenerateSGBBorder(struct GBVideoSoftwareRenderer* renderer) {
...
-				renderer->outputBuffer[base + 2] = renderer->palette[p | ((tileData[0] >> 5) & 0x1) | ((tileData[1] >> 4) & 0x2) | ((tileData[2] >> 3) & 0x4) | ((tileData[3] >> 2) & 0x8)];
-				renderer->outputBuffer[base + 3] = renderer->palette[p | ((tileData[0] >> 4) & 0x1) | ((tileData[1] >> 3) & 0x2) | ((tileData[2] >> 2) & 0x4) | ((tileData[3] >> 1) & 0x8)];
-				renderer->outputBuffer[base + 4] = renderer->palette[p | ((tileData[0] >> 3) & 0x1) | ((tileData[1] >> 2) & 0x2) | ((tileData[2] >> 1) & 0x4) | ((tileData[3] >> 0) & 0x8)];
-				renderer->outputBuffer[base + 5] = renderer->palette[p | ((tileData[0] >> 2) & 0x1) | ((tileData[1] >> 1) & 0x2) | ((tileData[2] >> 0) & 0x4) | ((tileData[3] << 1) & 0x8)];
-				renderer->outputBuffer[base + 6] = renderer->palette[p | ((tileData[0] >> 1) & 0x1) | ((tileData[1] >> 0) & 0x2) | ((tileData[2] << 1) & 0x4) | ((tileData[3] << 2) & 0x8)];
-				renderer->outputBuffer[base + 7] = renderer->palette[p | ((tileData[0] >> 0) & 0x1) | ((tileData[1] << 1) & 0x2) | ((tileData[2] << 2) & 0x4) | ((tileData[3] << 3) & 0x8)];
-			}
+			int j, q;
```
endrift: Please name these a bit better.

src/gb/renderers/software.c
```diff
@@ -76,25 +76,18 @@ static void _regenerateSGBBorder(struct GBVideoSoftwareRenderer* renderer) {
...
+
+			for (i = 0; i < 8; ++i) {
+				j = SGBBgAttributesIsXFlip(mapData) ? i : 7 - i;
```
endrift: Probably want two different versions of this loop to avoid doing this check so often.

dobyrch:
Ok, I just pushed a new version with the changes you requested.

PR #1024, opened 2018-03-27
GB: Fix GBC games running on GBA with BIOS

dobyrch:
If the Game Boy Color model is set to Game Boy Advance, it only works as expected if the "Use BIOS" option is false or "Skip BIOS" is true.  `GBSkipBIOS` sets register B to 1 (which games can use to determine if they are running on a GBA), but register B never gets set when loading a real boot ROM.

Maybe there's a better approach (and I'd be happy to change this if there is), but this patches the GBC boot ROM to load 1 into B when the model is set to GBA.  There are two bytes of `NOOP` near the end of the main routine, which gives us exactly enough room to add the new instruction without compromising anything else.  It's almost too good to be a coincidence, and makes me wonder if the GBA contains a modified GBC boot ROM that works exactly like this...

Anyway, I tested this with Oracle of Ages, which uses a brighter color palette and opens a new shop when run on a GBA:

gbc gba

### DIFF
--- a/src/gb/gb.c
+++ b/src/gb/gb.c
@@ -555,6 +555,10 @@ void GBUnmapBIOS(struct GB* gb) {
 		free(gb->memory.romBase);
 		gb->memory.romBase = gb->memory.rom;
 	}
+	// XXX: Force AGB registers for AGB-mode
+	if (gb->model == GB_MODEL_AGB && gb->cpu->pc == 0x100) {
+		gb->cpu->b = 1;
+	}
 }
 
 void GBDetectModel(struct GB* gb) {
--- a/src/platform/libretro/libretro.c
+++ b/src/platform/libretro/libretro.c
@@ -66,7 +66,7 @@ static void _reloadSettings(void) {
 	enum GBModel model;
 	const char* modelName;
 
-	var.key = "mgba_model";
+	var.key = "mgba_gb_model";
 	var.value = 0;
 	if (environCallback(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value) {
 		if (strcmp(var.value, "Game Boy") == 0) {
@@ -75,6 +75,8 @@ static void _reloadSettings(void) {
 			model = GB_MODEL_SGB;
 		} else if (strcmp(var.value, "Game Boy Color") == 0) {
 			model = GB_MODEL_CGB;
+		} else if (strcmp(var.value, "Game Boy Advance") == 0) {
+			model = GB_MODEL_AGB;
 		} else {
 			model = GB_MODEL_AUTODETECT;
 		}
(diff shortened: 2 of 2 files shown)


src/gb/gb.c
```diff
@@ -420,6 +420,12 @@ void GBReset(struct LR35902Core* cpu) {
 				memcpy(&gb->memory.romBase[0x100], &gb->memory.rom[0x100], sizeof(struct GBCartridge));
 			}
 
+			if (gb->model == GB_MODEL_AGB) {
```
endrift: I'm not very fond of having a hack to do this, but I'll let you know if I come up with a better idea.

src/gb/gb.c
```diff
@@ -420,6 +420,12 @@ void GBReset(struct LR35902Core* cpu) {
...
+				/* Games can check register B to determine if they are running on GBA */
```
endrift: C++ style comments (`//`) please.

src/gb/gb.c
```diff
@@ -420,6 +420,12 @@ void GBReset(struct LR35902Core* cpu) {
...
+				uint8_t ldB1[3] = {0x00, 0x06, 0x01};
+				memcpy(gb->memory.romBase + 0xF3, ldB1, 3);
```
endrift: `memcpy(&gb->memory.romBase[0xF3], ldB1, sizeof(ldB1);`, possibly defining a constant to replace the magic 0xF3.

src/gb/gb.c
```diff
@@ -420,6 +420,12 @@ void GBReset(struct LR35902Core* cpu) {
 				memcpy(&gb->memory.romBase[0x100], &gb->memory.rom[0x100], sizeof(struct GBCartridge));
 			}
 
+			if (gb->model == GB_MODEL_AGB) {
+				/* Games can check register B to determine if they are running on GBA */
+				uint8_t ldB1[3] = {0x00, 0x06, 0x01};
```
endrift: Spaces around the braces (I think, check other parts of my code to see if I'm consistent about this).

src/platform/libretro/libretro.c
```diff
@@ -140,7 +142,7 @@ void retro_set_environment(retro_environment_t env) {
 	struct retro_variable vars[] = {
 		{ "mgba_solar_sensor_level", "Solar sensor level; 0|1|2|3|4|5|6|7|8|9|10" },
 		{ "mgba_allow_opposing_directions", "Allow opposing directional input; OFF|ON" },
-		{ "mgba_model", "Game Boy model (requires restart); Autodetect|Game Boy|Super Game Boy|Game Boy Color" },
+		{ "mgba_model", "Game Boy model (requires restart); Autodetect|Game Boy|Super Game Boy|Game Boy Color|Game Boy Advance" },
```
endrift: Since it's only been one commit would you mind changing this to `mgba_gb_model`?

dobyrch:
Done and done.  I left the magic number as it was because I can't think of a very meaningful name, but hopefully it's clear that its just a memory offset into the boot ROM.

endrift:
Here's an idea for an alternate, slightly less hacky approach: instead of patching the BIOS, when the BIOS is unmapped set the register

```diff
diff --git a/src/gb/gb.c b/src/gb/gb.c
index 4d404e83c..b932e9eda 100644
--- a/src/gb/gb.c
+++ b/src/gb/gb.c
@@ -555,6 +555,10 @@ void GBUnmapBIOS(struct GB* gb) {
 		free(gb->memory.romBase);
 		gb->memory.romBase = gb->memory.rom;
 	}
+	// XXX: Force AGB registers for AGB-mode
+	if (gb->model == GB_MODEL_AGB && gb->cpu->pc == 0x100) {
+		gb->cpu->b = 1;
+	}
 }
 
 void GBDetectModel(struct GB* gb) {
```

dobyrch:
Works for me, that's nicer than fiddling with machine code. :+1:

PR #1030, opened 2018-03-31
GB Serialize: Don't serialize SGB packet

dobyrch:
If the most recent SGB command consisted of multiple packets (like an `ATTR_BLK` command with three or more data sets), the serialized state only contains the last packet.  When it is replayed by loading a save state or rewinding, the packet is (incorrectly) interpreted as a command, causing miscolored graphics:

rewind

### DIFF
--- a/include/mgba/internal/gb/serialize.h
+++ b/include/mgba/internal/gb/serialize.h
@@ -395,8 +395,7 @@ struct GBSerializedState {
 		uint8_t command;
 		uint8_t bits;
 		GBSerializedSGBFlags flags;
-		uint8_t packet[16];
-		uint32_t reserved[4];
+		uint32_t reserved[8];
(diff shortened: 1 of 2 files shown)


endrift:
While this is a legitimate bug it's not the correct solution for it; if you load in the middle of a command sequence the later commands have the same issue.

dobyrch:
That's true, but since games use SGB commands pretty sparingly, "any time after a multi-packet command" is a much bigger window than "the time between the first and last packet of a command."  I've been playing with this patch for a while, and although I can still trigger the bug if I try (by rewinding or saving at _just_ the right time when colors are changing), I haven't experienced miscolored graphics with normal use.

Even if it doesn't completely fix the problem, at it least makes things a little better without adding any complexity.

I probably won't have time to come up with a more correct solution anytime soon, but I could close this pull request and open a new issue so this bug doesn't get forgotten.

PR #1035, opened 2018-04-04
Correct debug lib

uknys:
Fd changed from int to void* so a change is needed to make the debug library work :)


endrift:
When did the newlib change happen?

PR #1047, opened 2018-04-08
Doc: Add git and base-devel to MSYS2 dependencies

rootfather:
"git" is obviously needed for pulling the latest source code.

"base-devel" is needed to actually make the compiler toolchain work.

Both packages are not installed in current MSYS2 base installations.


endrift:
What does base-devel drag in that's needed? make?

rootfather:
Make, ar and I think various other related tools. It’s by far the most convenient way to get a build environment up and running.

PR #1058, opened 2018-04-20
doc: Add manual for the config file.

bentley:
Still very, very rough, but I’m looking for feedback.

* The title only makes sense if config.ini gets renamed to mgba.ini.
* Is there a complete list of settings?
* How many of these are qt only? They should go in the qt section.
* Are there any other sections besides [ports.qt] and inputs? What’s the difference between gba.input-profile and gba.input?
* Default values should probably be documented for most of these.


endrift:
fwiw I'm possibly going to be doing a revamp where I convert to YAML at some point, at which point I can easily rename the config file to `mgba.yaml`.

lyneca:
Just like to bump this, is there a complete list of settings available?

endrift:
Not at the moment. They're kinda added without much thought so such a list would rapidly become obsolete.

PR #1150, opened 2018-08-05
GBA: Zero DebugString buffer when handling messages

ketsuban:
At the moment mGBA doesn't zero the DebugString buffer in [`GBADebug`](https://github.com/mgba-emu/mgba/blob/master/src/gba/gba.c#L521-L532) after copying a message out into `oolBuf`; if I don't either, and happen to use a message which is shorter than the one which came before it, odd things happen.

### DIFF
--- a/src/gba/gba.c
+++ b/src/gba/gba.c
@@ -525,6 +525,7 @@ void GBADebug(struct GBA* gba, uint16_t flags) {
 		level &= 0x1F;
 		char oolBuf[0x101];
 		strncpy(oolBuf, gba->debugString, sizeof(gba->debugString));
+		memset(gba->debugString, 0, sizeof(gba->debugString));
 		oolBuf[0x100] = '\0';
 		mLog(_mLOG_CAT_GBA_DEBUG(), level, "%s", oolBuf);
 	}


endrift:
How are you writing to the buffer such that it winds up not null-terminated properly? The code in opt looks like it should do that. If you write to it directly and elide the null byte, well, yeah weird things are going to happen.

ketsuban:
I'm using [`ptr::copy_nonoverlapping`](https://doc.rust-lang.org/std/ptr/fn.copy_nonoverlapping.html).
```rust
/// Writes a message to the debugger at the given severity.
fn message(severity: u16, string: &str) {
    // UNSAFE: The string pointer is statically guaranteed to be valid, and we
    // cap the length so as not to overflow the buffer.
    unsafe {
        ptr::copy(
            string.as_ptr() as *const u8,
            REG_DEBUG_STRING.0 as *mut u8,
            cmp::min(string.len(), 255),
        );
    }
    REG_DEBUG_FLAGS.write(severity);
}
```
Rust strings aren't null-terminated, which had skipped my mind yesterday. It's the reason there's a `CStr` type, but it's intended for FFI so the only way to construct one is with a raw pointer.

Beyond compatibility with languages which don't null-terminate their strings, though, it makes semantic sense to me for `gba->debugString` to be cleared - you can think of it as the system consuming the message when you write to the flag register, so if you read it back afterwards you get blank memory.

endrift:
I'm not really fond of that logic but the use case is clear, whereas the use cases for keeping it in place is not.

PR #1154, opened 2018-08-24

kddlb:
I also updated the `ts` files for the other translations.


endrift:
Thanks! Will merge once Travis is done.

Issue #1142, opened 2018-07-27
Sound issue in PCM channel


retro-wertz:
Only tested with a GBC game, in particular Bio Hazard/Resident Evil Gaiden.

Problem is that the volume during the part where is says "Resident Evil" (of "Bio Hazard on the J version) is very loud compared to other channels. Using the Audio channels option to isolate which channel this comes from has no effect since even with all 4 channels off, this voice sample is still heard(doesn't seem to be hooked up to any channel on/off option).

the comparison of the volume is based on video samples on youtube etc.

manually compiled:
Commit: f888d9b
gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1)

endrift:
The GBC does not have a PCM channel, so I'm a bit confused.

endrift:
Here's a comparison with it on hardware: https://www.youtube.com/watch?v=vRTll7JqV2U

retro-wertz:
(those guys on the vid are so annoyinglu loud, so the voice sample during when it says "resident evil" is not heard.

not a gb /gbc expect here. so its probably wave channel or channel 3 (at least on some emu that can disable the 4 channels.)

here is a better sample (not sure if this is from an actual hardware) skip to 0:22 where it would say resident evil when showing the title screen.

https://www.youtube.com/watch?v=jIVwXBvMdNk&t=34s

endrift:
Yeah sorry it was the only obviously-hardware video I could find.

The trick it's doing is a little weird because it doesn't involve ANY of the sound channels--it's messing with audio artifacting caused by changing the volume rapidly, even when nothing is playing.

Issue #749, opened 2017-06-05
Donkey Kong Country (Game Boy Color) & Perfect Dark (Game Boy Color) - Sound samples don't play


GeekyFerret:
In the Game Boy Color games Donkey Kong Country and Perfect Dark, all sound samples don't play. 

Donkey Kong Country only uses them in the fishing bonus game. [Example](https://www.youtube.com/watch?v=NMuPBXNcUVE).

Perfect Dark uses sound samples many times throughout the game (Nintendo/Rare logo animations, voice acting, and most in game sound effects). [Example](https://www.youtube.com/watch?v=xeLNNsVPSCA).

endrift:
wrong number

PR #1164, opened 2018-09-07
GB Audio: Fix handling of the Game Boy's unsigned samples

eevee:
I _believe_ this to be correct.  It fixes some confusing behavior I saw while trying to get a long sample to play (by repeatedly editing the wavetable and doing a delicate dance to pause it just long enough).  It now matches SameBoy, at least for my purposes.  :)  And the waveform is no longer upside-down!

I listened to a couple songs in Oracle of Ages and Pokémon Crystal, and they sound _very slightly_ more crackly, but this is the Game Boy, so that's just as likely to be more correct.  mGBA with this patch is still not as crackly as SameBoy, which I'm told has very good APU emulation.

The one thing that confuses me is how this worked _at all_ before when playing multiple channels.  The sound should've been offset by 8 * (number of channels - 1), which seems...  significant?  I compared before/after recordings and the difference looks basically like white noise, so maybe it's imperceptible in most cases.

### DIFF
--- a/src/gb/audio.c
+++ b/src/gb/audio.c
@@ -576,46 +576,46 @@ void GBAudioUpdateFrame(struct GBAudio* audio, struct mTiming* timing) {
 
 void GBAudioSamplePSG(struct GBAudio* audio, int16_t* left, int16_t* right) {
 	int dcOffset = audio->style == GB_AUDIO_GBA ? 0 : 0x8;
-	int sampleLeft = dcOffset;
-	int sampleRight = dcOffset;
+	int sampleLeft = 0;
+	int sampleRight = 0;
 
 	if (audio->playingCh1 && !audio->forceDisableCh[0]) {
 		if (audio->ch1Left) {
-			sampleLeft -= audio->ch1.sample;
+			sampleLeft += audio->ch1.sample - dcOffset;
 		}
 
 		if (audio->ch1Right) {
-			sampleRight -= audio->ch1.sample;
+			sampleRight += audio->ch1.sample - dcOffset;
 		}
 	}
 
 	if (audio->playingCh2 && !audio->forceDisableCh[1]) {
 		if (audio->ch2Left) {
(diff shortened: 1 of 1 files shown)


endrift:
This almost certainly would reintroduce #749, and #1142, among other issues. See also https://github.com/LIJI32/SameBoy/issues/100, which is the same as #1142.

eevee:
Oh, no, you're right.  I fear I've opened a can of rabbit holes.  Summary thusfar:

 Behavior | Hardware | mGBA → this PR | SameBoy | gambatte | bgb
----------|-----------|------------------|-----------|----------|------
Spikes when toggling ch3 on and off | yes? | yes → yes | yes | **no** | **no**
Spikes when muting and unmuting ch3 | no? | **yes** → no | no | no | no
Plays RE/PD sounds | yes | yes → **no** | **no** | yes | **no**

Obviously we need to do a three-way merge between the open source codebases, and then everything will be fine.

If having all channels disabled doesn't play (signed) zero, which must be the case since volume still has an impact…  then all I can think of is that I'm wrong about the hardware behavior of briefly unplugging channel 3?  I'll have to put this on hold until I can confirm for sure.

eevee:
Tried my test ROMs on a real GBA SP, and the spiking is painfully obvious on _all three_ of them.  So it looks like this patch is completely wrong, and mGBA is just the only emulator that correctly sounds…  well, like a GBA sounds.  :)  Gambatte is closer to GBC speakers.

I've also discovered that Duke Nukem for GBC uses arbitrary sounds without spiking, with the same method I tried originally — simply disabling channel 3, reloading the wavetable, and re-enabling — but it sounds fine both in mGBA and on hardware.  For the life of me I can't figure out why it works.

endrift:
Should I close this then? Or do you have things you want to revamp?

eevee:
Nothing at the moment, so I'll close this.

PR #1203, opened 2018-10-03
Windows: Inno setup script

endrift:
Based on @lotharsm's installer, which I modified and integrated into CMake.

Replaces #941.


lotharsm:
Wow, this looks great! Thanks a lot for taking care of it and for the cmake integration! In my opinion, ready to get merged!

PR #1210, opened 2018-10-08
improve libretro

lacastri:
higher rate, lower latency, libretro memory

### DIFF
--- a/src/platform/libretro/libretro.c
+++ b/src/platform/libretro/libretro.c
@@ -26,7 +26,8 @@
 #include <mgba-util/memory.h>
 #include <mgba-util/vfs.h>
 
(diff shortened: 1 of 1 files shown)


lacastri:
From what I've read, 48000+ may work for purely psg (gb/c) but gba has 32768 (x1-x4) digital sampling rates. So 32768 => 48000 (blip) => 479xy.z (libretro sinc) could introduce more scratchy artifacts due to twice interpolation than a simple in-out at default 32768. Snes libretro notably outputs at ~32000/32040 since that's the native rate and frontend will sinc up to final rate.

Lowering the samples to 512 or 256 is unclear to me of side-effects so leaving that alone for official repository.

Better audio needs more investigation.

PR #1213, opened 2018-10-09
libretro: cheat search support

lacastri:


lacastri:
Retroarch 1.7.4+ has a menu option when game loaded called 'Cheats => Start or Continue Cheat Search => Start of Restart Cheat Search' which doesn't work with mgba. Exposing the memory allows access to the cheat searcher and you can look for memory value changes in gb/a . But I didn't realize the non-contiguous part. Oops. I'll just rely on Cheat Engine.

Thanks for explaining it carefully.

endrift:
That's on RetroArch though, since that memory API totally is part of libretro and does work.

Anyway I don't know how advanced their cheat search is but the standalone version of mGBA (at least the dev builds) has a basic version too.

lacastri:
Their cheat search engine is pretty decent and flexible so far. ```RETRO_ENVIRONMENT_SET_MEMORY_MAPS``` is currently for cheevos support and core-specific cheats. But reading api further, they could just edit the frontend to use that instead of ```retro_get_memory_data``` for memory searching.

PR #1263, opened 2019-01-13
Fixed pokemon & kirby link communication errors

kakashidinho:
Basically, let slaves wait until they use all added cycles from `addCycles` callback before signalling master to continue.

Also added macro switch to allow switching between using mutex vs using atomic ops for lockstep.

### DIFF
--- a/include/mgba-util/common.h
+++ b/include/mgba-util/common.h
@@ -77,6 +77,7 @@ typedef intptr_t ssize_t;
 #define ATOMIC_STORE(DST, SRC) __atomic_store_n(&DST, SRC, __ATOMIC_RELEASE)
 #define ATOMIC_LOAD(DST, SRC) DST = __atomic_load_n(&SRC, __ATOMIC_ACQUIRE)
 #define ATOMIC_ADD(DST, OP) __atomic_add_fetch(&DST, OP, __ATOMIC_RELEASE)
+#define ATOMIC_SUB(DST, OP) __atomic_sub_fetch(&DST, OP, __ATOMIC_RELEASE)
 #define ATOMIC_OR(DST, OP) __atomic_or_fetch(&DST, OP, __ATOMIC_RELEASE)
 #define ATOMIC_AND(DST, OP) __atomic_and_fetch(&DST, OP, __ATOMIC_RELEASE)
 #define ATOMIC_CMPXCHG(DST, EXPECTED, SRC) __atomic_compare_exchange_n(&DST, &EXPECTED, SRC, true,__ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE)
@@ -85,6 +86,7 @@ typedef intptr_t ssize_t;
 #define ATOMIC_STORE(DST, SRC) DST = SRC
 #define ATOMIC_LOAD(DST, SRC) DST = SRC
 #define ATOMIC_ADD(DST, OP) DST += OP
+#define ATOMIC_SUB(DST, OP) DST -= OP
 #define ATOMIC_OR(DST, OP) DST |= OP
 #define ATOMIC_AND(DST, OP) DST &= OP
 #define ATOMIC_CMPXCHG(DST, EXPECTED, OP) ((DST == EXPECTED) ? ((DST = OP), true) : false)
--- a/include/mgba-util/common.h
+++ b/include/mgba-util/common.h
@@ -81,6 +81,14 @@ typedef intptr_t ssize_t;
 #define ATOMIC_OR(DST, OP) __atomic_or_fetch(&DST, OP, __ATOMIC_RELEASE)
 #define ATOMIC_AND(DST, OP) __atomic_and_fetch(&DST, OP, __ATOMIC_RELEASE)
 #define ATOMIC_CMPXCHG(DST, EXPECTED, SRC) __atomic_compare_exchange_n(&DST, &EXPECTED, SRC, true,__ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE)
+#elif defined _MSC_VER
+#define ATOMIC_STORE(DST, SRC) InterlockedExchange(&DST, SRC)
+#define ATOMIC_LOAD(DST, SRC) DST = InterlockedOrAcquire(&SRC, 0)
+#define ATOMIC_ADD(DST, OP) InterlockedAddRelease(&DST, OP)
+#define ATOMIC_SUB(DST, OP) InterlockedAddRelease(&DST, -OP)
+#define ATOMIC_OR(DST, OP) InterlockedOrRelease(&DST, OP)
+#define ATOMIC_AND(DST, OP) InterlockedAndRelease(&DST, OP)
+#define ATOMIC_CMPXCHG(DST, EXPECTED, SRC) (InterlockedCompareExchange(&DST, SRC, EXPECTED) == EXPECTED)
 #else
 // TODO
 #define ATOMIC_STORE(DST, SRC) DST = SRC
--- a/include/mgba-util/platform/posix/threading.h
+++ b/include/mgba-util/platform/posix/threading.h
@@ -29,6 +29,18 @@ static inline int MutexInit(Mutex* mutex) {
 	return pthread_mutex_init(mutex, 0);
 }
 
+static inline int MutexInitRecursive(Mutex* mutex) {
+	pthread_mutexattr_t attr;
+	pthread_mutexattr_init(&attr);
+	pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
+
+	int re = pthread_mutex_init(mutex, &attr);
+
+	pthread_mutexattr_destroy(&attr);
+
+	return re;
+}
+
 static inline int MutexDeinit(Mutex* mutex) {
 	return pthread_mutex_destroy(mutex);
 }
--- a/include/mgba/core/lockstep.h
+++ b/include/mgba/core/lockstep.h
@@ -35,6 +35,7 @@ struct mLockstep {
 	bool (*wait)(struct mLockstep*, unsigned mask);
 	void (*addCycles)(struct mLockstep*, int id, int32_t cycles);
 	int32_t (*useCycles)(struct mLockstep*, int id, int32_t cycles);
+	int32_t (*unusedCycles)(struct mLockstep*, int id);
 	void (*unload)(struct mLockstep*, int id);
 	void* context;
 #ifndef NDEBUG
--- a/include/mgba/core/lockstep.h
+++ b/include/mgba/core/lockstep.h
@@ -7,6 +7,9 @@
 #define SIO_LOCKSTEP_H
 
 #include <mgba-util/common.h>
+#include <mgba-util/threading.h>
+
+#define MGBA_LOCK_STEP_USE_MUTEX 1
 
 CXX_GUARD_START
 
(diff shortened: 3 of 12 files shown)


include/mgba-util/common.h
```diff
@@ -77,14 +77,24 @@ typedef intptr_t ssize_t;
 #define ATOMIC_STORE(DST, SRC) __atomic_store_n(&DST, SRC, __ATOMIC_RELEASE)
 #define ATOMIC_LOAD(DST, SRC) DST = __atomic_load_n(&SRC, __ATOMIC_ACQUIRE)
 #define ATOMIC_ADD(DST, OP) __atomic_add_fetch(&DST, OP, __ATOMIC_RELEASE)
+#define ATOMIC_SUB(DST, OP) __atomic_sub_fetch(&DST, OP, __ATOMIC_RELEASE)
```
endrift: There was no `ATOMIC_SUB` for a reason: you can just negate the operand.

include/mgba-util/platform/posix/threading.h
```diff
@@ -29,6 +29,18 @@ static inline int MutexInit(Mutex* mutex) {
 	return pthread_mutex_init(mutex, 0);
 }
 
+static inline int MutexInitRecursive(Mutex* mutex) {
```
endrift: Why did you add this? You'd need to port it to all of the threading platforms and I don't think it's necessary.

include/mgba/core/lockstep.h
```diff
@@ -19,14 +22,20 @@ enum mLockstepPhase {
...
 
 struct mLockstep {
+#if MGBA_LOCK_STEP_USE_MUTEX
+	Mutex mutex;
+	Condition cond;
+#endif
+
 	int attached;
-	enum mLockstepPhase transferActive;
+	int32_t transferActive; // must be one of the values in mLockstepPhase
```
endrift: Why did you change this away from an enum?

include/mgba/internal/gb/sio/lockstep.h
```diff
@@ -38,6 +38,7 @@ struct GBSIOLockstepNode {
 };
 
 void GBSIOLockstepInit(struct GBSIOLockstep*);
+void GBSIOLockstepInit2(struct GBSIOLockstep*, bool initBase);
```
endrift: Why is this needed?

src/gb/sio/lockstep.c
```diff
@@ -236,6 +243,8 @@ static uint8_t GBSIOLockstepNodeWriteSC(struct GBSIODriver* driver, uint8_t valu
 			mTimingDeschedule(&driver->p->p->timing, &driver->p->event);
 			mTimingDeschedule(&driver->p->p->timing, &node->event);
 			mTimingSchedule(&driver->p->p->timing, &node->event, 0);
+		} else {
+			mLOG(GB_SIO, FATAL, "GBSIOLockstepNodeWriteSC() failed to write to masterClaimed\n");
```
endrift: This should absolutely not be `FATAL`.

src/gba/sio/lockstep.c
```diff
@@ -8,7 +8,23 @@
 #include <mgba/internal/gba/gba.h>
 #include <mgba/internal/gba/io.h>
 
-#define LOCKSTEP_INCREMENT 3000
+#define LOCKSTEP_INCREMENT 128
```
endrift: This sounds like it's going to be a massive overhead. This is way too low.

src/gba/sio/lockstep.c
```diff
@@ -8,7 +8,23 @@
...
+
+#if MGBA_LOCK_STEP_USE_MUTEX
+#	define LOCK_LOCK_STEP(node) MutexLock(&node.mutex)
```
endrift: Don't indent.

src/gba/sio/lockstep.c
```diff
@@ -19,7 +35,14 @@ static uint16_t GBASIOLockstepNodeNormalWriteRegister(struct GBASIODriver* drive
 static void _GBASIOLockstepNodeProcessEvents(struct mTiming* timing, void* driver, uint32_t cyclesLate);
 
 void GBASIOLockstepInit(struct GBASIOLockstep* lockstep) {
-	mLockstepInit(&lockstep->d);
+	GBASIOLockstepInit2(lockstep, true);
```
endrift: Why is this needed?

src/gba/sio/lockstep.c
endrift: Also lockstep is one word.

src/gba/sio/lockstep.c
```diff
@@ -148,13 +190,17 @@ static uint16_t GBASIOLockstepNodeMultiWriteRegister(struct GBASIODriver* driver
...
+	UNLOCK_LOCK_STEP(node->p->d);
+
 	return value;
 }
 
 static void _finishTransfer(struct GBASIOLockstepNode* node) {
 	if (node->transferFinished) {
 		return;
 	}
+
```
endrift: Unnecessary code churn.

src/gba/sio/lockstep.c
```diff
@@ -223,30 +279,30 @@ static int32_t _masterUpdate(struct GBASIOLockstepNode* node) {
 		node->p->multiRecv[2] = 0xFFFF;
 		node->p->multiRecv[3] = 0xFFFF;
 		needsToWait = true;
-		ATOMIC_STORE(node->p->d.transferActive, TRANSFER_STARTED);
-		node->nextEvent += 512;
+		SAFE_STORE(node->p->d.transferActive, TRANSFER_STARTED);
+		node->nextEvent += LOCKSTEP_INCREMENT;
```
endrift: This was a separate increment for a reason--it's related to the minimum transfer duration.

src/gba/sio/lockstep.c
```diff
@@ -269,16 +325,28 @@ static int32_t _masterUpdate(struct GBASIOLockstepNode* node) {
...
 	}
 	return node->nextEvent;
 }
 
 static uint32_t _slaveUpdate(struct GBASIOLockstepNode* node) {
-	node->d.p->multiplayerControl.ready = node->p->attachedMulti == node->p->d.attached;
+	LOCK_LOCK_STEP(node->p->d);
+
+	enum mLockstepPhase transferActive;
+	int attachedMulti, attached;
```
endrift: Separate lines.

src/gba/sio/lockstep.c
```diff
@@ -288,6 +356,8 @@ static uint32_t _slaveUpdate(struct GBASIOLockstepNode* node) {
 	case TRANSFER_FINISHING:
 		break;
 	case TRANSFER_STARTED:
+		if (node->p->d.unusedCycles(&node->p->d, node->id) > node->eventDiff)
```
endrift: Braces.

src/gba/sio/lockstep.c
```diff
@@ -315,6 +385,8 @@ static uint32_t _slaveUpdate(struct GBASIOLockstepNode* node) {
 		signal = true;
 		break;
 	case TRANSFER_FINISHED:
+		if (node->p->d.unusedCycles(&node->p->d, node->id) > node->eventDiff)
```
endrift: Braces.

src/platform/qt/MultiplayerController.cpp
```diff
@@ -16,43 +16,76 @@
 
 using namespace QGBA;
 
+#if MGBA_LOCK_STEP_USE_MUTEX
+#define ACQUIRE_CONTROLLER(controller, lockstep) MutexLock(&lockstep->mutex);
+#define RELEASE_CONTROLLER(controller, lockstep) MutexUnlock(&lockstep->mutex);
+#define CORE_THREAD_WAIT(player, lockstep) mCoreThreadWaitFromThread(player->controller->thread());
+#define CORE_THREAD_STOP_WAITING(player, lockstep) mCoreThreadStopWaiting(player->controller->thread());
+
```
endrift: Unnecessary newline.

src/platform/qt/MultiplayerController.cpp
```diff
@@ -16,43 +16,76 @@
...
+#else // MGBA_LOCK_STEP_USE_MUTEX
+#define ACQUIRE_CONTROLLER(controller, lockstep) controller->m_lock.lock()
+#define RELEASE_CONTROLLER(controller, lockstep) controller->m_lock.unlock()
+#define CORE_THREAD_WAIT(player, lockstep) mCoreThreadWaitFromThread(player->controller->thread());
```
endrift: Why #define this if it's the same on both sides of the conditional?

src/platform/qt/MultiplayerController.cpp
```diff
@@ -16,43 +16,76 @@
...
+#define CORE_THREAD_STOP_WAITING(player, lockstep) mCoreThreadStopWaiting(player->controller->thread());
+#endif // MGBA_LOCK_STEP_USE_MUTEX
+
+MultiplayerController::Player::Player(
```
endrift: Please match existing style.

src/platform/qt/MultiplayerController.h
```diff
@@ -42,6 +45,17 @@ Q_OBJECT
 
 private:
 	struct Player {
+		Player(
```
endrift: Match existing style.

src/util/ring-fifo.c
```diff
@@ -7,6 +7,18 @@
 
 #include <mgba-util/memory.h>
 
+#if defined _MSC_VER
```
endrift: Why is this needed? It should already include the file these are defined in.

include/mgba-util/platform/posix/threading.h
kakashidinho: In your callback code in `MultilplayerController` you have another lock. I noticed it is unnecessary if mutex is used with lockstep anw. So make it recursive. Anw, the lock call in callback might not be needed. So the recursive mutex is not really neccessary.
For the platform porting, right now i only see win32 and pthread ans they are already ported

include/mgba/core/lockstep.h
kakashidinho: Win32 atomic function does not support abitrary type

kakashidinho:
I tested pokemon ruby and even though it can play battle and trade now. After exitting the multiplayer mode, the link errors still occurred sometimes. Pokemon emerald seems fine.
So this might not be the final fix for this PR.

kakashidinho:
For the mutex approach. As mentioned, your old code used mutex in the callbacks and they are in turn invoked in every lockstep event processing, if there is any overhead, it already exists. So just by introducing mutex to the event processing, the atomic ops will not be needed anymore (and possibly removing mutex from callbacks).

src/util/ring-fifo.c
kakashidinho: Again, due to rigid nature of win32 interlocked functions. Pointers need different versions of atomic function, but in this module, the code assumes the atomic* version for integers can work for pointers. So I have to do this. It's not best way, but it the quick way without modifying the resting the code to use something like `ATOMIC_STORE_PTR` for example.

src/gba/sio/lockstep.c
kakashidinho: To allow caller to choose to avoid init the base struct again, in case he/she wants to defer the initialization of derived struct such as the code in `MultiplayerController` which calls `mLockstepInit ` in constructor and separately calls `GBASIOLockstepInit ` in `attachGame`

kakashidinho:
Close due to link error still occurs occasionally

endrift:
I merged your changes with some stylistic (mostly) fixes on my end.

PR #1271, opened 2019-01-18
SDL: Add option to drop privileges with unveil()/pledge()

bentley:
[`pledge()`](https://man.openbsd.org/pledge.2) and [`unveil()`](https://man.openbsd.org/unveil.2) are two OpenBSD APIs to restrict the capabilities of a program. `pledge` whitelists the set of syscalls the program can call, while `unveil` whitelists a view of the filesystem to the program.

`pledge` can be called repeatedly to reduce the list of allowed syscalls even further. This patch calls pledge twice, initially with a very broad list and finally with `"stdio rpath wpath cpath fattr"` (for savestates) `"drm"` (for graphics) `"audio"` (for audio). The debuggers additionally need `"tty"` (`-d`) and `"net"` (`-g`). The kernel kills a program that violates its promises. Obviously if that ever happens, the arguments to pledge need to be fixed.

`unveil` is called once per file or directory to whitelist, with a specified permission (read/write/execute/create), and a final call with null arguments to prevent further whitelisting. Attempts to access anything not unveiled result in `ENOENT` or `EACCES`. The kernel’s method of keeping track of these files is somewhat expensive, so typical practice is to call `unveil` as late as possible with as few files as possible. Here I only pass it the savestates.

Libepoxy is not compatible with these pledges; its deferred `dlopen()` calls cause the the program to be killed on `glDeleteTextures()`. I disabled it conditionally in CMake.

I’ve tested normal gameplay, USB gamepad, networked [sndio](http://www.sndio.org/) audio, and basic `-d` usage (read: started program and hit “c”).

Questions:

* Right now only the base names of the savestates are unveiled, so they only work if the ROM is in `$PWD`. How do I pass the directory through to the SDL frontend?
* Any obvious file accesses or syscalls that I missed? Particularly in the debuggers?
* Any style issues?


bentley:
New revision, pulling in the savestate path the same way savestate writes do. (Kosher?) `unveil` works now for a ROM in any location.

bentley:
That’s what I thought, but it actually worked.

endrift:
That's not reliable though, it just happens to be that the struct elements were in the right order for that to work out.

bentley:
Bummer… reverted. So hooking the VDir creation routines will be the next step.

bentley:
Clearly I’m not going to get around to finishing unveil support soon… however, the pledge support works well, and has been part of the OpenBSD package for over a year. So let’s reduce the scope of this pull request to pledge’s syscall filtering, which is ready to go in. What are your thoughts?

endrift:
I'll take a more in-depth look tomorrow (it's 4 AM here, I shouldn't still be awake) but you should probably start by restricting the status output and checks in CMake to just being on OSes that have it (OpenBSD and...SerenityOS I think?)

bentley:
I think SerenityOS shows up as “Generic” in CMake and isn’t otherwise detectable, but I don’t know for sure. Because of that I only gated pledge() support behind a variable that defaults to off. I can change that to `if (CMAKE_SYSTEM_NAME STREQUAL OpenBSD)` if you want though.

CMakeLists.txt
```diff
@@ -473,6 +476,10 @@ find_feature(USE_SQLITE3 "sqlite3")
 find_feature(USE_ELF "libelf")
 find_feature(ENABLE_PYTHON "PythonLibs")
 
+if(USE_PLEDGE)
```
endrift, line 479: This should go above the `find_feature` calls. Probably above the block of `set(WANT_`, but not necessarily.

CMakeLists.txt
```diff
@@ -1221,6 +1232,9 @@ if(NOT QUIET AND NOT LIBMGBA_ONLY)
 		message(STATUS "	CLI debugger: ${USE_EDITLINE}")
 	endif()
 	message(STATUS "	GDB stub: ${USE_GDB_STUB}")
+	if(CMAKE_SYSTEM_NAME STREQUAL OpenBSD)
```
endrift, line 1235: There probably isn't a strong reason to print this in the status output.

src/platform/sdl/main.c
```diff
@@ -149,6 +154,15 @@ int main(int argc, char** argv) {
 	renderer.player.bindings = &renderer.core->inputMap;
 	mSDLInitBindingsGBA(&renderer.core->inputMap);
 	mSDLInitEvents(&renderer.events);
+
+#ifdef USE_PLEDGE
+	if (!mPledgeBroad(&args)) {
```
endrift, line 159: Is there any reason this isn't done immediately after generating the `args` struct? There's a lot of stuff you're not freeing or deiniting here.

src/platform/sdl/main.c
```diff
@@ -149,6 +154,15 @@ int main(int argc, char** argv) {
...
+		freeArguments(&args);
+		fprintf(stderr, "pledge\n");
```
endrift, line 161: You should use a more descriptive error message. Also probably fputs, but that's more minr.

src/platform/sdl/main.c
```diff
@@ -264,6 +278,12 @@ int mSDLRun(struct mSDLRenderer* renderer, struct mArguments* args) {
 					state->close(state);
 				}
 			}
+#ifdef USE_PLEDGE
+		if (!mPledgeNarrow(args)) {
```
endrift, line 282: What's the purpose of this second, narrower call? To drop further privs after everything is initialized? It looks like all you drop is `dns` and `unix`, and I'm not sure what those are used for in the first place. SDL I guess. Note that the runloop will still get called if this fails.

Also the indentation doesn't match.

src/platform/sdl/main.c
```diff
@@ -264,6 +278,12 @@ int mSDLRun(struct mSDLRenderer* renderer, struct mArguments* args) {
...
+			didFail = true;
+			fprintf(stderr, "pledge\n");
```
endrift, line 284: Ditto about the error message.

src/platform/sdl/main.c
bentley: Joystick initialization calls some USB ioctls that are not allowed by any pledge, so this is as early as the pledge call can go.

src/platform/sdl/main.c
bentley: Yes, that’s the purpose—pledge is intended to ratchet down permissions before entering the main loop.

`dns` and `unix` are needed for [audio](https://man.openbsd.org/sio_open.3#Use_with_pledge_2):

> * `unix` is needed when the device is a local sndiod(8) sub-device.
> * `inet` and `dns` are needed when the device is a remote sndiod(8) sub-device.
>
> Once no further calls to sio_open() will be made, all these pledge(2) promises may be dropped, except for the `audio` promise.

bentley: Is the right thing here to return didFail right away and let the caller deinit everything?

endrift: Everything after the runloop except for `mCoreThreadHasCrashed` still needs to be called.

src/platform/sdl/main.c
bentley: I gave this another try.

CMakeLists.txt
```diff
@@ -489,6 +496,10 @@ find_feature(USE_SQLITE3 "sqlite3")
 find_feature(USE_ELF "libelf")
 find_feature(ENABLE_PYTHON "PythonLibs")
 
+if(USE_PLEDGE)
```
endrift, line 499: You forgot to delete it here.

src/platform/sdl/main.c
```diff
@@ -149,6 +154,20 @@ int main(int argc, char** argv) {
...
+
+#ifdef USE_PLEDGE
+	if (!mPledgeBroad(&args)) {
+		freeArguments(&args);
+		mCoreConfigDeinit(&renderer.core->config);
+		mInputMapDeinit(&renderer.core->inputMap);
+		renderer.core->deinit(renderer.core);
+		mSDLDeinitEvents(&renderer.events);
+		mSDLDeinit(&renderer);
+		fputs("Broad pledge() failed\n", stderr);
```
endrift, line 166: `fputs` already appends the `\n`.

src/platform/sdl/main.c
```diff
@@ -149,6 +154,20 @@ int main(int argc, char** argv) {
...
 	mSDLInitBindingsGBA(&renderer.core->inputMap);
 	mSDLInitEvents(&renderer.events);
+
+#ifdef USE_PLEDGE
+	if (!mPledgeBroad(&args)) {
+		freeArguments(&args);
+		mCoreConfigDeinit(&renderer.core->config);
+		mInputMapDeinit(&renderer.core->inputMap);
+		renderer.core->deinit(renderer.core);
+		mSDLDeinitEvents(&renderer.events);
```
endrift, line 164: `mSDLDeinitEvents` is called by `mSDLDeinit`, you don't need to call it yourself. `mCoreConfigFreeOpts` should be called though.

src/platform/sdl/main.c
```diff
@@ -264,6 +283,12 @@ int mSDLRun(struct mSDLRenderer* renderer, struct mArguments* args) {
 					state->close(state);
 				}
 			}
+#ifdef USE_PLEDGE
+			if (!mPledgeNarrow(args)) {
+				didFail = true;
+				fputs("Narrow pledge() failed\n", stderr);
```
endrift, line 289: Ditto about the `\n`.

PR #1275, opened 2019-01-22
Tools: Allow for separate fast forward ratios for held/toggled speedup

briaguya:
The current implementation allows for a toggle or hold of fast forward. This, however, means that there is no way to do something similar to snes9x in which there is a increase speed/decrease speed hotkey as well as a "turbo" hotkey (which just uncaps the fps).

I have implemented a separate setting to control the speedup ratio when the fast forward button is held, and the logic to allow the user to (for example):
1. Set the toggled fast forward speed to 200%
2. Set the "held" fast forward speed to uncapped
3. Toggle fast forward on (press shift tab) and play at 2x speed
4. Hold fast forward (tab) to uncap
5. Let go of fast forward (tab) to go back to 2x speed
6. Toggle fast forward off (press shift tab) and go back to 1x speed


endrift:
Please rebase, I'd like to merge this.

briaguya:
@endrift I've pulled the latest master into my branch and verified the functionality on my local machine. I updated the PR, merge conflicts resolved.

PR #1288, opened 2019-01-31
Changes needed for MINIMAL_CORE=3 ...

kode54:
... to remove even more unneeded extra fluff. Useful
if compiling the library only for GSF playback, for
instance.

### DIFF
--- a/src/gba/core.c
+++ b/src/gba/core.c
@@ -164,8 +164,10 @@ static bool _GBACoreInit(struct mCore* core) {
 	mRTCGenericSourceInit(&core->rtc, core);
 	gba->rtcSource = &core->rtc.d;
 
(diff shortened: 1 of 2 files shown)


endrift:
I think having a flag DISABLE_VIDEO would be better than overloading MINIMAL_CORE. In fact, I wanted to branch out with feature flags to replace MINIMAL_CORE entirely.

kode54:
Alright, just let me know what you do settle on, since I don’t need video rendering for GSF playback, but it doesn’t get in the way, either, it just means more needs to be compiled into libmgba.

endrift:
This is pretty stale and I've already done a bunch of work on ENABLE flags instead, so I'm gonna close this.

PR #1289, opened 2019-01-31
Attempt simple fix for channel 4

kode54:
This fix attempts to boost the volume of the divider averaged samples to be comparable to the original zero-order-hold or nearest-neighbor approach. It sounds almost as loud, and produces the intended averaged samples.

### DIFF
--- a/src/gb/audio.c
+++ b/src/gb/audio.c
@@ -37,7 +37,7 @@ static bool _updateSweep(struct GBAudioSquareChannel* sweep, bool initial);
 static void _updateSquareSample(struct GBAudioSquareChannel* ch);
 static int32_t _updateSquareChannel(struct GBAudioSquareChannel* ch);
 
-static int8_t _coalesceNoiseChannel(struct GBAudioNoiseChannel* ch);
+static int16_t _coalesceNoiseChannel(struct GBAudioNoiseChannel* ch);
 
 static void _updateFrame(struct mTiming* timing, void* user, uint32_t cyclesLate);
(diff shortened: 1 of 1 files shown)


src/gb/audio.c
```diff
@@ -762,12 +767,12 @@ static int32_t _updateSquareChannel(struct GBAudioSquareChannel* ch) {
...
 
-static int8_t _coalesceNoiseChannel(struct GBAudioNoiseChannel* ch) {
+static int16_t _coalesceNoiseChannel(struct GBAudioNoiseChannel* ch) {
 	if (!ch->nSamples) {
-		return ch->sample;
+		return ch->sample * 8;
 	}
 	// TODO keep track of timing
-	int8_t sample = ch->samples / ch->nSamples;
+	int16_t sample = ch->samples * 16 / ch->nSamples;
```
endrift: Why 16?

src/gb/audio.c
```diff
@@ -627,19 +629,22 @@ void GBAudioSamplePSG(struct GBAudio* audio, int16_t* left, int16_t* right) {
...
 		if (audio->ch4Right) {
-			sampleRight += sample;
+			sampleRightUnscaled += sample;
 		}
 	}
 
 	sampleLeft <<= 3;
 	sampleRight <<= 3;
 
+	sampleLeft += sampleLeftUnscaled;
```
endrift: It may make sense to just move the upshift to before adding in channel 4 instead of managing two sets of variables.

PR #1341, opened 2019-03-04
GBA: Add override for Touhoumon MMW and WL

dengler9:
Hacks from Ruby and Emerald respectively, they use a custom Game ID (AICE and AIME).
Activating rtc in game overrides works, but you have to activate it every time you exit and reopen the emulator.
Not sure if this change is acceptable because these are hacks and here are only licensed games.
If this change is unacceptable a possible workaround would be to change in a hex editor the lines TOUHOUMONMMWAIME with POKEMON RUBYAXVE for "Touhoumon Marisa's Magic World" and TOUHOUMON_WLAICE with POKEMON EMERBPEE for "Touhoumon World Link" but I have no knowledge if this can have a side effect (i don't think so).

### DIFF
--- a/src/gba/overrides.c
+++ b/src/gba/overrides.c
@@ -182,6 +182,12 @@ static const struct GBACartridgeOverride _overrides[] = {
 
 	// Aging cartridge
 	{ "TCHK", SAVEDATA_EEPROM, HW_NONE, IDLE_LOOP_NONE, false },
+	
+	// Touhoumon Marisa's Magic World
+	{ "AIME", SAVEDATA_FLASH1M, HW_RTC, IDLE_LOOP_NONE, false },
+
+	// Touhoumon World Link
+	{ "AICE", SAVEDATA_FLASH1M, HW_RTC, 0x80008C6, false },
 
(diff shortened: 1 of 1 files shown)


endrift:
Sorry, ROM hacks aren't allowed in that section. You can save the overrides, however.

dengler9:
> Sorry, ROM hacks aren't allowed in that section. You can save the overrides, however.

Well, logic told me that but i still had to try, sorry.
Please tell me how to save the overrides then because i can't.
1. open mGBA
2. Tools -> Game overrides option
3. Uncheck autodetect -> check realtime clock -> Save?
4. Open Touohoumon WL or MMW -> The internal battery...... message doesn't appear
6. Tools -> Game overrides option -> Save again
5. Exit emulator -> Open it again
6. Open Touohoumon WL or MMW -> The internal battery...... message now appears
Maybe i am doing something wrong, and again sorry for bothering you.

endrift:
That sounds like a bug. I'll try and fix that, thanks.

dengler9:
Yes, I think it's a regression because it works as intended in 0.6.3 but not in 0.7 and 0.7.1.

PR #1404, opened 2019-05-05
Update mgba-zh_CN.ts

EmpyreusX:
Fully retranslated and re-aligned zh-CN lang file (with all strings fixed).


src/platform/qt/ts/mgba-zh_CN.ts
```diff
@@ -85,7 +85,7 @@ Game Boy Advance 是 Nintendo Co., Ltd. 的注册商标。</translation>
         <location filename="../AssetTile.ui" line="48"/>
         <location filename="../AssetTile.ui" line="72"/>
         <source>0</source>
-        <translation>0</translation>
```
endrift: You're still removing these for no apparent reason. Don't do that.

PR #1418, opened 2019-05-23

kddlb:


src/platform/qt/ts/mgba-es.ts
```diff
@@ -417,7 +501,7 @@ Game Boy Advance es una marca registrada de Nintendo Co., Ltd.</translation>
     <message>
         <location filename="../LogView.ui" line="28"/>
         <source>Debug</source>
-        <translation>Depuración (Debug)</translation>
+        <translation>Debug</translation>
```
fitojb: Why did’t you leave the Spanish translation of the term? _Debug_ literally means _depuración_.

Issue #54, opened 2015-07-01
Get building with Visual Studio 2015


endrift:
Left to do:
- [x] Get replacement for `getopt` (possibly [from Napoleon-OS/libc](https://github.com/Napoleon-OS/libc/blob/master/src/misc/getopt_long.c)).
- [x] Port `vfs-dirent.c` to MSVC.
- [x] Get extern symbols working.
- [x] Figure out how to handle dependencies.
- [x] Add AppVeyor integration.

karasuhebi:
Yay, I should be able to build mGBA soon :D Too bad I have 2013 so I'll probably need to upgrade XD

endrift:
I did post instructions for setting up a self-contained build system using MSYS2 to the readme last night (11098b3). You don't get VS' debugger, but you get a functional Windows build still!

endrift:
A bit late in the game for this. I'll try to do it early in 0.4.

endrift:
VDir support added in 3a51c59768a0e4bbba9dc516b0554eb520474c05.

endrift:
getopt and general build fixes merged in #257!

endrift:
VideoView.cpp fails to build in MSVC due to the use of C99isms. I'll fix that in the next version. Beyond that, managing dependencies is still an issue.

endrift:
cc @waddlesplash, how do you think I should handle dependencies for MSVC, specifically ImageMagick and ffmpeg being the biggest obstacles?

rootfather:
I would include the necessary header files for all dependencies in the repository itself. Would be the best option in my opinion. But let's what what @waddlesplash says.

waddlesplash:
I guess my recommendation would be to add some sort of script that can download binaries of all the dependencies.

For FFmpeg: FFmpeg itself doesn't distribute binaries, but their official website sends you to to the [Zeranoe builds](https://ffmpeg.zeranoe.com/builds/) (and everybody under the sun who uses FFmpeg on Windows does use those). There are static+devel versions available, so that should work.

ImageMagick: If you have a custom ImageMagick build, maybe throw it up on the ImageMagick fork on @mgba-emu as a release, and download that as well? I don't know, I don't have very much experience with ImageMagick.

waddlesplash:
Er, scratch the FFmpeg comment; it looks like Zeranoe doesn't provide MSVC builds anymore :/ (or if they do I can't find them.) Drat.

endrift:
[...]

PR #1452, opened 2019-06-18
All: Microsoft Visual C++ compatibility code

arves100:
This commit adds Microsoft Visual C++ compatibility. (Should fix " Get extern symbols working." and some dependency handle (SDL2+epoxy) #54 )

Changes:
 - Native cmake doesn't correctly find deps with pkg_search_module, I added a new folder called cmake containing the cmake directives to find the required libraries. (Right now I have added SDL2 and epoxy, while forcing epoxy for Windows builds under there)
- Added a new header: mgba/exports.h generated by cmake, by adding this header and some decorations in the missing functions, I was able to export the remaning variables.
- Modified find package of SDL2 because it doesn't look for the correct sdl2.
- Forced the console to be spawned in mgba-sdl
- Added UNICODE definition and a check to ignore wmain, since it's already defined by SDL_main

Notes:
- I have done just a minimum build with almost no 3rd party module support
- I haven't tested QT build yet.
- Although it should not be an issue, static libraries would need to be tested and check if they don't use __declspec directives. (BUILD_STATIC should be defined as a CXX flag, but perhaps it isn't required if BUILD_STATIC is ON)


cmake/FindSDL2.cmake
```diff
@@ -0,0 +1,162 @@
...
+# Copyright 2003-2009 Kitware, Inc.
+#
+# Distributed under the OSI-approved BSD License (the "License");
+# see accompanying file Copyright.txt for details.
+#
+# This software is distributed WITHOUT ANY WARRANTY; without even the
+# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the License for more information.
+#=============================================================================
+# (To distribute this file outside of CMake, substitute the full
```
endrift: Please follow this instruction

cmake/Findepoxy.cmake
```diff
@@ -0,0 +1,10 @@
+find_path(EPOXY_INCLUDE_DIRS NAMES epoxy/gl.h PATH_SUFFIX include)
```
endrift: cmake files go in `src/platform/cmake`

src/platform/sdl/CMakeLists.txt
```diff
@@ -3,12 +3,15 @@ set(SDL_VERSION "2" CACHE STRING "Version of SDL to use (1.2 or 2)")
...
 	pkg_search_module(SDL2 sdl2)
-	if (SDL2_FOUND)
-		set(SDL_INCLUDE_DIR ${SDL2_INCLUDE_DIRS})
-		set(SDL_LIBRARY ${SDL2_LIBRARIES})
-		link_directories(${SDL2_LIBDIR})
-		set(SDL_VERSION_DEBIAN "2-2.0-0")
-		set(SDL_FOUND ON PARENT_SCOPE)
+	if (NOT SDL2_FOUND)
+		find_package(SDL2)
+		if (SDL2_FOUND)
```
endrift: Unnest this if, it'll fail if pkg_search_module succeeds right now.

cmake/FindSDL2.cmake
arves100: I should have followed the istruction in the last commit. (I am not sure to be honest)

arves100:
The last commit should be ok.
Just a clarification, exports.h (now called dllexports.h because a file called export.h already exists in mgba-util and I didn't want to create a confusion with the names), is automaticly generated by Cmake, so it should be shipped by default.
The line that does this job is the following:
generate_export_header(${BINARY_NAME} BASE_NAME mgba STATIC_DEFINE BUILD_STATIC EXPORT_FILE_NAME include/mgba-util/dllexports.h)

CMakeLists.txt
```diff
@@ -944,6 +948,9 @@ if(NOT SKIP_LIBRARY)
 		add_library(${BINARY_NAME} STATIC ${SRC})
 	endif()
 
+	include(GenerateExportHeader)
+	generate_export_header(${BINARY_NAME} BASE_NAME mgba STATIC_DEFINE BUILD_STATIC EXPORT_FILE_NAME include/mgba-util/dllexports.h)
```
endrift: Apologies for missing this the first time around and thus giving you the wrong advice, but this should be exported to `${CMAKE_CURRENT_BINARY_DIR}/include/mgba-util`, which may need to be created manually.

src/platform/cmake/Findepoxy.cmake
```diff
@@ -0,0 +1,10 @@
+find_path(EPOXY_INCLUDE_DIRS NAMES epoxy/gl.h PATH_SUFFIX include)
+find_library(EPOXY_LIBRARIES NAMES epoxy_0 PATH_SUFFIX)
+
+if(EPOXY_LIBRARIES AND EPOXY_INCLUDE_DIRS)
+    set(epoxy_FOUND TRUE)
+endif()
+
+if (WIN32 AND NOT epoxy_FOUND)
+    message(FATAL_ERROR "Windows requires epoxy module!")
```
endrift: This fails in mingw where the library is `epoxy` and the path suffix is `lib` (but it should just fall back to pkg-config if it can't find them manually)

arves100:
Ok it should be everything in order. I don't have Mingw so I can't report if it does work fine or not. Under MSVC the detection also fails (due to Windows not having a package manager and for some reason epoxy_0 is not found), I manually set EPOXY_LIBRARIES to make sure it does link the correct one.

PR #1460, opened 2019-06-23
CMake: fix libretro building

funbars:


endrift:
This change isn't correct if SKIP_LIBRARY isn't set.

funbars:
I successful build using this with PR patch:
```
mkdir build
cd build
cmake --version
cmake .. -G "Visual Studio 15 2017" -DLIBMGBA_ONLY=ON -DBUILD_LIBRETRO=ON
msbuild mgba_libretro.vcxproj /t:build /p:Configuration="%config%";Platform="%platform%" /m
```

but don't know CMake well. Original error:
```message(FATAL_ERROR "Windows requires epoxy module!")```

which libretro win32 doesn't need.

PR #1542, opened 2019-10-06
Fix bug in deleting breakpoints

Cleverking2003:
Fix a very silly bug which prevented breakpoints from being deleted. I'm surprised that it went unnoticed.


endrift:
This is a functional change, changing from deleting by ID to deleting by address, which would revert an intentional change I made in the past. You can list breakpoint IDs already so there's no reason you can't delete breakpoints as is.

Cleverking2003:
Lol, my bad. Maybe you could add command arguments in help command, so there'd be less confusion?

PR #1543, opened 2019-10-07
Feature: GBA Link connection emulation

shim80:
Hi, 
Is there a way to emulate the GBA Link connection to play multiplayer games using WiFi or Bluetooth?
As far I know, it's possible in some emulators on PC and Android...


endrift:
Not currently

PR #1662, opened 2020-02-08
Always use HLE-BIOS when loading multiboot games.

niemand-deu:
mGBA can't run multiboot games when skipping the BIOS (somehow muddles up the working HLE code path) or while using an accurately cloned / official BIOS (broken, game doesn't get loaded into EWRAM or ROM; even then would still need to be properly set up).


endrift:
I think forcing skipping BIOS is a better idea than forcing no BIOS for multiboot games. Can you verify if that works?

endrift:
Obviated by fb12372ae3e5b71b9c2fc07657bd2601f842709e

PR #1682, opened 2020-03-01
Fix Haiku config directory and SDL 1.2 build

Crestwave:


src/core/config.c
```diff
@@ -227,6 +231,12 @@ void mCoreConfigDirectory(char* out, size_t outLength) {
 	UNUSED(portable);
 	snprintf(out, outLength, "/%s", projectName);
 	FSUSER_CreateDirectory(sdmcArchive, fsMakePath(PATH_ASCII, out), 0);
+#elif defined(__HAIKU__)
+	UNUSED(portable);
```
endrift: You should probably have a portable.ini check like the other desktop OSes do.

Crestwave: Oh, I thought that it was for a portable build distributed for specific OSes or something. I haven't tested it yet but I'm guessing a copy of the other check will do

Crestwave: Okay, so I built it and launched it in a directory with a file named portable.ini and the config files were placed in that directory. That's the expected behavior, yes?

endrift: Yes, that's expected behavior.

PR #1710, opened 2020-03-29
Qt: Update Italian GUI translation

Mte90:
My way to contribute to this tool :-)


lotharsm:
Thanks for your contribution :)

Could you please modify the commit message so it reads something along the lines of

 - Qt: Update Italian GUI translation ?

Mte90:
You should be able to do it with squash merge instead of classic pull request merge.

lotharsm:
Ah, right! Merging then :)

PR #1719, opened 2020-04-04
QT: Better text colors in the debugger

Mte90:
immagine
immagine

### DIFF
--- a/include/mgba/debugger/debugger.h
+++ b/include/mgba/debugger/debugger.h
@@ -99,6 +99,7 @@ struct mDebuggerPlatform {
 	void (*deinit)(struct mDebuggerPlatform*);
 	void (*entered)(struct mDebuggerPlatform*, enum mDebuggerEntryReason, struct mDebuggerEntryInfo*);
 
+	bool (*supportHTML)(struct mDebuggerPlatform*);
 	bool (*hasBreakpoints)(struct mDebuggerPlatform*);
 	void (*checkBreakpoints)(struct mDebuggerPlatform*);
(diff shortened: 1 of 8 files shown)


Mte90:
Before that I continue to work on that is helpful? There is anything that I need to know to avoid code duplication?

I cannot find the definition of this method `debugger->backend->printf` to replicate the behaviour.

include/mgba/debugger/debugger.h
```diff
@@ -99,6 +99,7 @@ struct mDebuggerPlatform {
 	void (*deinit)(struct mDebuggerPlatform*);
 	void (*entered)(struct mDebuggerPlatform*, enum mDebuggerEntryReason, struct mDebuggerEntryInfo*);
 
+	bool (*supportHTML)(struct mDebuggerPlatform*);
```
Mte90, line 102: Probably I can remove this line

src/platform/qt/DebuggerConsole.ui
```diff
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>480</width>
+    <width>508</width>
```
Mte90, line 9: better round to 510

endrift:
I'm not really interested making the backend support HTML just because the frontend can render it. HTML is bad here in a number of ways, and the debugger console is kind of temporary until the real debugger UI is done.

Mte90:
I know that HTML is not very good but it was an easy way to get a better output and maybe also to open new integration to mgba.

PR #1736, opened 2020-04-22
Use consistent conditionals for big endian detection in PSR union

lifning:
This was breaking mGBA rather badly on my PPC little-endian host, starting from bf8c1d1b4b52d45cd84590dc7bd7d40b0dffc7bd with the introduction of the `#if defined(__BIG_ENDIAN__)` below the one I changed here.

Optimistically, this fix might also help more than just folks with modern POWER systems -- after all, not all PowerPC hosts are big-endian, and PowerPC isn't the only big-endian architecture.

Cheers
-lif

### DIFF
--- a/include/mgba/internal/arm/arm.h
+++ b/include/mgba/internal/arm/arm.h
@@ -70,7 +70,7 @@ struct ARMCore;
 
 union PSR {
 	struct {
-#if defined(__POWERPC__) || defined(__PPC__)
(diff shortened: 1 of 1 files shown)


endrift:
Thanks, I wasn't sure if I'd fixed all of these places, and I didn't have a ppcle machine to test on. Clearly I'd missed one!

PR #1742, opened 2020-04-27
Update menu.c

issame:
Fix Vita Menu
Menu on PS Vita looks cropped on the right. This fixes it


issame:
Did some changes, so the changes are Vita only

PR #1760, opened 2020-05-23
Dutch translation

Maxane:


Maxane:
My apoligies, I thought I could merge this into the master of the work, of which I could work from there

Issue #1784, opened 2020-06-10
[feature req.] Rom file name instead of coded name in the titlebar


Liger0:
It would be cool to have the option to decide to show the official rom name or the specified filename in the program titlebar.
It would be useful to instantly notice, especially when in link connection, which instance is using which rom, to avoid confusion.

Testsr:
Should have been closed with #1807.

PR #1807, opened 2020-06-24
Setting to show filename in title bar instead of ROM name.

mdx97:
This implements the feature requested in #1784 

One caveat: the best source of data I could find for the currently loaded ROM's file name was the `baseName` field of the `mDirectorySet` struct (the instance held by `mCore`). However, this doesn't seem to carry the file extension.

So, if you loaded the file `pokemon-fire-red.gba`, you would only see `pokemon-fire-red`.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,7 +2,6 @@
 Features:
  - e-Reader card scanning
  - Add APNG recording
- - Setting to display the file name of the currently loaded ROM instead of the game name in the title bar (closes mgba.io/i/1784)
 Emulation fixes:
  - ARM: Fix ALU reading PC after shifting
  - ARM: Fix STR storing PC after address calculation
@@ -37,6 +36,7 @@ Misc:
  - Qt: Add hex index to palette view
  - Qt: Add transformation matrix info to sprite view
  - Qt: Add per-page scrolling to memory view (fixes mgba.io/i/1795)
+ - Qt: Setting to display the file name of the currently loaded ROM instead of the game name in the title bar (closes mgba.io/i/1784)
 
 0.8.2: (2020-06-14)
 Emulation fixes:
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,7 @@
 Features:
  - e-Reader card scanning
  - Add APNG recording
+ - Setting to display the file name of the currently loaded ROM instead of the game name in the title bar (closes mgba.io/i/1784)
 Emulation fixes:
  - ARM: Fix ALU reading PC after shifting
  - ARM: Fix STR storing PC after address calculation
--- a/CHANGES
+++ b/CHANGES
@@ -36,7 +36,7 @@ Misc:
  - Qt: Add hex index to palette view
  - Qt: Add transformation matrix info to sprite view
  - Qt: Add per-page scrolling to memory view (fixes mgba.io/i/1795)
- - Qt: Setting to display the file name of the currently loaded ROM instead of the game name in the title bar (closes mgba.io/i/1784)
+ - Qt: Add setting to display ROM filename in title (closes mgba.io/i/1784)
 
 0.8.2: (2020-06-14)
 Emulation fixes:
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -328,7 +328,8 @@ void mCoreLoadForeignConfig(struct mCore* core, const struct mCoreConfig* config
 
 	mCoreConfigCopyValue(&core->config, config, "cheatAutosave");
 	mCoreConfigCopyValue(&core->config, config, "cheatAutoload");
-
+	mCoreConfigCopyValue(&core->config, config, "showFilename");
+	
 	core->loadConfig(core, config);
 }
 
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -328,8 +328,7 @@ void mCoreLoadForeignConfig(struct mCore* core, const struct mCoreConfig* config
 
 	mCoreConfigCopyValue(&core->config, config, "cheatAutosave");
 	mCoreConfigCopyValue(&core->config, config, "cheatAutoload");
-	mCoreConfigCopyValue(&core->config, config, "showFilename");
-	
+
 	core->loadConfig(core, config);
 }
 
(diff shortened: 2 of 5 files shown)


src/platform/qt/Window.cpp
```diff
@@ -1032,13 +1032,23 @@ void Window::updateTitle(float fps) {
...
+
 		title = gameTitle;
+		delete[] gameTitle;
 
 #ifdef USE_SQLITE3
-		if (db && crc32 && NoIntroDBLookupGameByCRC(db, crc32, &game)) {
+		if (db && crc32 && NoIntroDBLookupGameByCRC(db, crc32, &game) && !showFilename) {
 			title = QLatin1String(game.name);
 		}
 #endif
```
endrift, line 1054: This entire block should be gated on the config, not oddly beaten into submission. The other branch doesn't need to deal with char*s directly, and can just convert things directly to QString (the value of `title`).

src/platform/qt/Window.cpp
```diff
@@ -1032,13 +1032,23 @@ void Window::updateTitle(float fps) {
 		uint32_t crc32 = 0;
 		m_controller->thread()->core->checksum(m_controller->thread()->core, &crc32, CHECKSUM_CRC32);
 
-		char gameTitle[17] = { '\0' };
+		char* gameTitle;
 		mCore* core = m_controller->thread()->core;
-		core->getGameTitle(core, gameTitle);
+		int showFilename = 0;
+
+		if (mCoreConfigGetIntValue(&core->config, "showFilename", &showFilename) && showFilename) {
```
endrift, line 1039: Use `m_config->getOption` instead.

src/platform/qt/Window.cpp
```diff
@@ -1031,27 +1031,23 @@ void Window::updateTitle(float fps) {
...
 
 #ifdef USE_SQLITE3
-		if (db && crc32 && NoIntroDBLookupGameByCRC(db, crc32, &game) && !showFilename) {
-			title = QLatin1String(game.name);
-		}
+			if (db && crc32 && NoIntroDBLookupGameByCRC(db, crc32, &game)) {
+				title = QLatin1String(game.name);
+			}
 #endif
+		}
```
mdx97: This what you meant when you said gate by the config? @endrift

endrift: Yup, that looks a lot better. Though you should probably be using the file path associated with the window (`windowFilePath()`) to get the full path and then use `QFileInfo::fileName()` to get the filename. There are times where that isn't set (like when only the BIOS is loaded), so you should make sure those work too.

CHANGES
```diff
@@ -2,6 +2,7 @@
 Features:
  - e-Reader card scanning
  - Add APNG recording
+ - Setting to display the file name of the currently loaded ROM instead of the game name in the title bar (closes mgba.io/i/1784)
```
endrift, line 5: This isn't major enough to be marked as a feature (it's a visual tweak), so put this in Misc instead.

src/platform/qt/Window.cpp
```diff
@@ -1031,17 +1031,23 @@ void Window::updateTitle(float fps) {
 		NoIntroGame game{};
 		uint32_t crc32 = 0;
 		m_controller->thread()->core->checksum(m_controller->thread()->core, &crc32, CHECKSUM_CRC32);
-
-		char gameTitle[17] = { '\0' };
+		
 		mCore* core = m_controller->thread()->core;
```
endrift, line 1035: Move this above the checksum and then change it to use `core`. (Yes, it was like this before. However, this is a fix I'd like and since the code around it is being modified now is a good time to fix it.)

src/core/core.c
```diff
@@ -328,7 +328,8 @@ void mCoreLoadForeignConfig(struct mCore* core, const struct mCoreConfig* config
 
 	mCoreConfigCopyValue(&core->config, config, "cheatAutosave");
 	mCoreConfigCopyValue(&core->config, config, "cheatAutoload");
-
+	mCoreConfigCopyValue(&core->config, config, "showFilename");
```
endrift, line 331: This is no longer needed and this file should be reverted.

mdx97:
I've incorporated the changes you've requested. But I'm testing now with the BIOS and we definitely need to do a check for that (it does not grab the file name, we should want it to display (BIOS) in the title regardless of the setting I'd imagine here).

I'm looking at how it is determined when to display (BIOS) on the screen. It looks like that is determined down at the `<GameboyType>GetGameTitle` function. It has a couple guard clauses and if those fail it defaults back to using (BIOS) as the game title. Is there an easier, existing way for us to check this from `Window::updateTitle`?

endrift:
You could fall back to the getGameTitle path if `windowFilePath().isNull()`.

mdx97:
Right, of course 😃 That works!

Also, do you know why the build keeps failing on osx? I don't really see where my code changes would be breaking it...

mdx97:
Also: since this last commit reverts the changes to `core.c`, the tag for the original commit should be changed to Qt instead of All if this gets merged, I think.

CHANGES
```diff
@@ -36,6 +36,7 @@ Misc:
  - Qt: Add hex index to palette view
  - Qt: Add transformation matrix info to sprite view
  - Qt: Add per-page scrolling to memory view (fixes mgba.io/i/1795)
+ - Qt: Setting to display the file name of the currently loaded ROM instead of the game name in the title bar (closes mgba.io/i/1784)
```
endrift, line 39: Please try to state this a bit more succinctly.

src/platform/qt/SettingsView.ui
```diff
@@ -657,6 +657,23 @@
...
+       <item row="21" column="1">
+        <widget class="QCheckBox" name="showFilename">
+         <property name="text">
+          <string>Show filename instead of ROM name in title bar</string>
+         </property>
+         <property name="checked">
+          <bool>true</bool>
+         </property>
+        </widget>
+       </item>
```
endrift, line 676: I want to see what this looks like before I approve the PR. However, I'm not sure further changes will be needed yet.

endrift:
I'm going to be squashing the commits anyway.

I think the Mac build in Travis is just hosed. I'm probably gonna ignore it.

endrift:
I'll merge after I test myself and make sure the layout of the settings is fine.

PR #1809, opened 2020-06-24
Fix italian real time clock translation

### LINKED ISSUE #1798: Real time clock setting on italian localization is broken
Annotazione 2020-06-19 133630

Hey there's this problem with real time clock, only in the italian localization (I checked english, german and french and they are fine).
Also in game it doesn't work, even if I use one of the working languages above, it's always night.
It's strange because on VBA-M it actually works, so I think it's not the rom's fault.

EddyHg80:
Title.

Looking at the other languages this should fix it.

Closes #1798 

In the spanish localization the hours were HH and not hh, is there a particular reason?


endrift:
HH is 24 hour, hh is 12 hour. Given that you removed the AM/PM you probably want to change it to HH too.

EddyHg80:
Alright, this should be it :)

PR #1819, opened 2020-07-15
Qt: Update Simplified Chinese GUI translation

LinRs:
- remove unnecessary whitespace
- fix non-quote punctuation

### DIFF
--- a/src/platform/qt/ts/mgba-zh_CN.ts
+++ b/src/platform/qt/ts/mgba-zh_CN.ts
@@ -126,7 +126,7 @@ Game Boy Advance 是任天堂有限公司（Nintendo Co., Ltd.）的注册商标
     </message>
 </context>
 <context>
-       <name>BattleChipView</name>
+    <name>BattleChipView</name>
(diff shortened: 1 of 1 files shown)


src/platform/qt/ts/mgba-zh_CN.ts
```diff
@@ -3250,7 +3245,7 @@ Game Boy Advance 是任天堂有限公司（Nintendo Co., Ltd.）的注册商标
     <message>
         <location filename="../MapView.cpp" line="107"/>
         <source>Map Addr.</source>
-        <translation>映射地址</translation>
+        <translation>映射地址.</translation>
```
endrift, line 3248: Any particular reason for this punctuation? In the English it's to mark that it's an abbreviation, which I assume isn't the case in the translation.

LinRs: There is no need to add this punctuation, I just want to make sure that the translation ends with the same punctuation as the source text.

Issue #1689, opened 2020-03-13
 Problem blinking white on PSVita


pinpon-1592:
When you try to load and play a game, you experience the following issues:

![vita-min](https://user-images.githubusercontent.com/62130172/76598233-ca55f080-6545-11ea-9d21-56ec87eeb4a5.png)

Video link: https://youtu.be/h-tQyTjHsyI

1. The screen flashes white like a video.
2.If you reset it as shown in the picture, the game screen before reset will remain.

The game I play with videos and photos is Wario Ware Twisted, but this happens with other games as well.

The version is 0.81.

What should I do?

(Google Translate use)

371784711:
me too

gotoAndDie:
I can confirm that this occurs on releases 0.8.1 and 0.8.2. However, version 0.8 beta 1 is fine, and the issue does not occur.

gotoAndDie:
The issue appears to be caused by commit 97e2cf0 (Vita: Avoid uncached memcpy).

endrift:
I was afraid that might be the case. It made it a lot faster, but...well, this happened.

gotoAndDie:
It looks like the bug is related to frameskip; when frameskip is 0, the screen doesn't flash white anymore. When I turn frameskip up to 1 again, the screen seems to be displaying the last frame it was displaying before frameskip every other frame.

endrift:
That would explain why I wasn't able to reproduce it earlier.

endrift:
I plan to either merge #1822 or come up with an alternate solution tomorrow.

PR #1822, opened 2020-07-20
Vita: Fix flickering when using frameskip

gotoAndDie:
This pull request should fix issue #1689. I've also reduced the size of vita2d_texture from 4 to 2; I haven't found anything that needs 4 buffers rather than 2, and interframe blending works with 2 buffers.

### DIFF
--- a/src/platform/psp2/psp2-context.c
+++ b/src/platform/psp2/psp2-context.c
@@ -52,7 +52,7 @@ static enum ScreenMode {
 
 static void* outputBuffer;
 static int currentTex;
-static vita2d_texture* tex[4];
+static vita2d_texture* tex[2];
 static vita2d_texture* screenshot;
 static Thread audioThread;
 static bool interframeBlending = false;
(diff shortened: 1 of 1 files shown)


src/platform/psp2/psp2-context.c
```diff
@@ -588,15 +584,27 @@ void _drawTex(vita2d_texture* t, unsigned width, unsigned height, bool faded, bo
 }
 
 void mPSP2Swap(struct mGUIRunner* runner) {
-	currentTex = (currentTex + 1) & 3;
-	runner->core->setVideoBuffer(runner->core, vita2d_texture_get_datap(tex[currentTex]), 256);
+	bool frameAvailable;
+	if (runner->core->platform(runner->core) == PLATFORM_GBA) {
+		struct GBA* gba = runner->core->board;
+		frameAvailable = gba->video.frameskipCounter <= 0;
```
endrift, line 590: Interesting approach. However, I don't like the idea of having to cast to a specific core to grab the frameskip counter. It's not generic, and I want to *reduce* the number of things peering into the cores, not increase. Not sure if there's a better approach however.

endrift:
If I can't come up with a better solution I'll merge this PR (and into 0.8.3) and add an API for figuring out if there's been a new frame posted in 0.9 so this can be generalized.

PR #1825, opened 2020-07-23
Qt: Add Japanese GUI translation

JaKid20:
Japanese translation by JaKid20


endrift:
The seems like a work in progress; a sizable number of strings are untranslated.

JaKid20:
We have not translated anything that changes meaning.
In addition, the significance of this pull request is that the translation can be basically used by Japanese people.
If you can't accept it, drop it.

PR #1832, opened 2020-07-29
Stack traces: add detection for other calling conventions and stack manipulation

ahigerd:
The stack trace implementation from the last PR didn't account for calling conventions common on the GBA and had some noteworthy oversights. This PR adds support for:
* calling a function by writing to `lr` followed by `b` or `bx` (since ARMv4 doesn't have `blx`)
* returning from a function using `mov pc, lr` (oops!)
* heuristically determining if `mov pc, rN` is meant to be a function return based on `sp` and `pc`
* skipped/unwound stack frames
* branches not taken (oops!)
* two-instruction `bl` in Thumb mode

It also cleans up backtrace rendering and some cosmetic miscalculations.

### DIFF
--- a/include/mgba/internal/arm/isa-inlines.h
+++ b/include/mgba/internal/arm/isa-inlines.h
@@ -107,4 +107,39 @@ static inline uint32_t _ARMPCAddress(struct ARMCore* cpu) {
 	return cpu->gprs[ARM_PC] - _ARMInstructionLength(cpu) * 2;
 }
 
+static inline bool ARMTestCondition(struct ARMCore* cpu, unsigned condition) {
+	switch (condition) {
+		case 0x0:
+			return ARM_COND_EQ;
+		case 0x1:
+			return ARM_COND_NE;
+		case 0x2:
+			return ARM_COND_CS;
+		case 0x3:
+			return ARM_COND_CC;
+		case 0x4:
+			return ARM_COND_MI;
+		case 0x5:
+			return ARM_COND_PL;
+		case 0x6:
+			return ARM_COND_VS;
+		case 0x7:
+			return ARM_COND_VC;
+		case 0x8:
+			return ARM_COND_HI;
+		case 0x9:
+			return ARM_COND_LS;
+		case 0xA:
+			return ARM_COND_GE;
+		case 0xB:
+			return ARM_COND_LT;
+		case 0xC:
+			return ARM_COND_GT;
+		case 0xD:
+			return ARM_COND_LE;
+		default:
+			return true;
+	}
+}
+
 #endif
--- a/src/arm/debugger/debugger.c
+++ b/src/arm/debugger/debugger.c
@@ -8,6 +8,7 @@
 #include <mgba/core/core.h>
 #include <mgba/internal/arm/arm.h>
 #include <mgba/internal/arm/decoder.h>
+#include <mgba/internal/arm/decoder-inlines.h>
 #include <mgba/internal/arm/isa-inlines.h>
 #include <mgba/internal/arm/debugger/memory-debugger.h>
 #include <mgba/internal/debugger/parser.h>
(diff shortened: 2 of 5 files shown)


include/mgba/internal/arm/isa-inlines.h
```diff
@@ -107,4 +107,41 @@ static inline uint32_t _ARMPCAddress(struct ARMCore* cpu) {
...
+			return ARM_COND_LS;
+		case 0xA:
+			return ARM_COND_GE;
+		case 0xB:
+			return ARM_COND_LT;
+		case 0xC:
+			return ARM_COND_GT;
+		case 0xD:
+			return ARM_COND_LE;
+		case 0x0E:
```
endrift, line 140: Nit: Extra 0

src/arm/arm.c
```diff
@@ -209,58 +209,9 @@ static inline void ARMStep(struct ARMCore* cpu) {
 	LOAD_32(cpu->prefetch[1], cpu->gprs[ARM_PC] & cpu->memory.activeMask, cpu->memory.activeRegion);
 
 	unsigned condition = opcode >> 28;
-	if (condition != 0xE) {
```
endrift, line 212: This will almost certainly cause a performance regression.

src/arm/arm.c
```diff
@@ -209,58 +209,9 @@ static inline void ARMStep(struct ARMCore* cpu) {
...
-		case 0xB:
-			conditionMet = ARM_COND_LT;
-			break;
-		case 0xC:
-			conditionMet = ARM_COND_GT;
-			break;
-		case 0xD:
-			conditionMet = ARM_COND_LE;
-			break;
-		default:
```
endrift, line 257: The 0xF case that you're implicitly removing here is needed on the medusa branch.

src/arm/debugger/debugger.c
```diff
@@ -16,14 +17,56 @@
...
+
 static bool ARMDebuggerUpdateStackTraceInternal(struct mDebuggerPlatform* d, uint32_t pc) {
 	struct ARMDebugger* debugger = (struct ARMDebugger*) d;
 	struct ARMCore* cpu = debugger->cpu;
 	struct ARMInstructionInfo info;
-	uint32_t instruction = cpu->prefetch[0];
 	struct mStackTrace* stack = &d->p->stackTrace;
+
+	struct mStackFrame* frame = mStackTraceGetFrame(stack, 0);
+	if (frame && frame->frameBaseAddress < (uint32_t)cpu->gprs[ARM_SP]) {
```
endrift, line 39: Nit: space between cast and castee.

src/arm/debugger/debugger.c
```diff
@@ -69,10 +107,42 @@ static bool ARMDebuggerUpdateStackTraceInternal(struct mDebuggerPlatform* d, uin
 		}
 		destAddress = info.op1.immediate + cpu->gprs[ARM_PC];
 	} else if (info.operandFormat & ARM_OPERAND_REGISTER_1) {
-		if (!isCall && info.op1.reg != ARM_LR && !(_ARMModeHasSPSR(cpu->cpsr.priv) && info.op1.reg == ARM_PC)) {
-			return false;
+		if (isCall) {
+			destAddress = cpu->gprs[info.op1.reg];
+		} else {
+			bool isExceptionReturn = _ARMModeHasSPSR(cpu->cpsr.priv) && info.affectsCPSR && info.op1.reg == ARM_PC;
+			bool isMovPcLr = (info.operandFormat & ARM_OPERAND_REGISTER_2) && info.op1.reg == ARM_PC && info.op2.reg == ARM_LR;
```
endrift, line 114: Though I doubt anything does `cmp pc, lr` (or `cmn`, `tst`, or `teq`), you should probably check `ARM_OPERAND_AFFECTED_1` as well.

src/arm/debugger/debugger.c
```diff
@@ -69,10 +107,42 @@ static bool ARMDebuggerUpdateStackTraceInternal(struct mDebuggerPlatform* d, uin
...
+			destAddress = cpu->gprs[reg];
+			if (isBranch || (info.op1.reg == ARM_PC && !isMovPcLr)) {
+				// ARMv4 doesn't have the BLX opcode, so it uses an assignment to LR before a BX for that purpose.
+				struct ARMInstructionInfo prevInfo;
+				if (cpu->executionMode == MODE_ARM) {
+					ARMDecodeARM(cpu->memory.load32(cpu, pc - 4, NULL), &prevInfo);
+				} else {
+					ARMDecodeThumb(cpu->memory.load16(cpu, pc - 2, NULL), &prevInfo);
+				}
+				if ((prevInfo.operandFormat & ARM_OPERAND_REGISTER_1) && (prevInfo.operandFormat & ARM_OPERAND_AFFECTED_1) && prevInfo.op1.reg == ARM_LR) {
```
endrift, line 126: `(prevInfo.operandFormat & (ARM_OPERAND_REGISTER_1 | ARM_OPERAND_AFFECTED_1))  == (ARM_OPERAND_REGISTER_1 | ARM_OPERAND_AFFECTED_1)` is fewer checks, and since this is a short circuit I'm not sure the compiler will optimize them together. (Though it may, since `prevInfo` isn't volatile.)

include/mgba/internal/arm/isa-inlines.h
```diff
@@ -107,4 +107,41 @@ static inline uint32_t _ARMPCAddress(struct ARMCore* cpu) {
...
+			return true;
+		default:
```
endrift, line 142: Unfortunately `0xF`. Varies between ARMv4 (where it's `NV`, thus always skipped) and ARMv5 (where it's a secondary set of instructions that are always executed). Some extra plumbing will be needed here on medusa.

src/arm/debugger/debugger.c
ahigerd: That's handled farther up the function by checking `info.branchType`.

endrift: In that case you should probably fix the bugs that makes that get set when `ARM_OPERAND_AFFECTED_1` isn't: https://github.com/mgba-emu/mgba/blob/master/src/arm/decoder-arm.c#L112 and https://github.com/mgba-emu/mgba/blob/master/src/arm/decoder-thumb.c#L125

(Also it looks like decoding `ldr pc, [...]` doesn't set that value: https://github.com/mgba-emu/mgba/blob/master/src/arm/decoder-arm.c#L177)

src/arm/debugger/debugger.c
ahigerd: Yeah, I couldn't really decide which form of repeating myself was less offensive. I'll switch it.

src/arm/arm.c
ahigerd: Hm. I'm not sure I agree. Lemme godbolt it.

endrift: This pre-condition was added because most ARM opcodes use that, and I actually benchmarked it (several years ago though).

ahigerd: Oh, that's what you meant. I thought you meant extracting the switch to a function would. Yeah, I really should have known better, and looking at the godbolt I realize where my mistake was in thinking it was equivalent.

endrift: Ah, yeah, and combining that with the fact that 0xF needs to be handled specially it probably just makes sense to leave `ARMStep` alone.

ahigerd: Rolled back.

endrift:
Will merge this in a bit, after the buildbot cools down some.

PR #1833, opened 2020-07-30
medusa: stack trace fixes

ahigerd:
This is the same as #1832 but for the medusa branch. The primary difference is in how the condition abstraction is handled in order to support ARMv5.


endrift:
I'm likely going to want to merge master into medusa and then have you rebase this instead of merging a slightly different version and then having merge conflicts later.

ahigerd:
That's fine. Just let me know when I should do it.

endrift:
I did the merging, you can rebase now.

ahigerd:
Done. You were right, this was a better idea.

PR #1835, opened 2020-07-31
Debugger: define symbols in CLI, show symbols in stack traces

ahigerd:
mGBA can already load symbols out of ELF files (and I think there's work being done on loading them from separate sym files). This PR adds the ability to define new symbols in the CLI debugger and to resolve addresses to their matching symbol table entries in stack traces.

As a bonus feature, it also adds a new scripting engine to allow running a list of debugger commands from a file. It's pretty simplistic, but it does the job.

### DIFF
--- a/src/debugger/cli-debugger-scripting.c
+++ b/src/debugger/cli-debugger-scripting.c
@@ -1,0 +1,137 @@
+/* Copyright (c) 2013-2020 Jeffrey Pfau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include <mgba/core/scripting.h>
+#include <mgba-util/string.h>
+#include <mgba-util/vfs.h>
+
+#include <mgba/debugger/debugger.h>
+#include <mgba/internal/debugger/cli-debugger.h>
+
+static const char* CLIScriptEngineName(struct mScriptEngine*);
+static bool CLIScriptEngineInit(struct mScriptEngine*, struct mScriptBridge*);
+static void CLIScriptEngineDeinit(struct mScriptEngine*);
+static bool CLIScriptEngineIsScript(struct mScriptEngine*, const char* name, struct VFile* vf);
+static bool CLIScriptEngineLoadScript(struct mScriptEngine*, const char* name, struct VFile* vf);
+static void CLIScriptEngineRun(struct mScriptEngine*);
+static bool CLIScriptEngineLookupSymbol(struct mScriptEngine*, const char* name, int32_t* out);
+static void CLIScriptDebuggerEntered(struct mScriptEngine*, enum mDebuggerEntryReason, struct mDebuggerEntryInfo*);
+
+struct CLIScriptStatement {
+	char* command;
+	size_t commandLen;
+};
+
+DECLARE_VECTOR(CLIScript, struct CLIScriptStatement);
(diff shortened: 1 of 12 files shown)


src/debugger/cli-debugger-scripting.c
```diff
@@ -0,0 +1,134 @@
+/* Copyright (c) 2013-2016 Jeffrey Pfau
```
endrift: It's still not 2016.

src/debugger/cli-debugger-scripting.c
```diff
@@ -0,0 +1,134 @@
...
+void CLIScriptEngineDeinit(struct mScriptEngine* se) {
+	struct CLIScriptEngine* engine = (struct CLIScriptEngine*) se;
+	CLIScriptClear(engine);
+	free(se);
+}
+
+bool CLIScriptEngineIsScript(struct mScriptEngine* se, const char* name, struct VFile* vf) {
+	UNUSED(se);
+	UNUSED(vf);
+	return endswith(name, ".mgba");
```
endrift, line 83: Not sure this extension makes sense, given that...medusa exists.

src/debugger/cli-debugger-scripting.c
```diff
@@ -0,0 +1,134 @@
...
+static bool CLIScriptEngineIsScript(struct mScriptEngine*, const char* name, struct VFile* vf);
+static bool CLIScriptEngineLoadScript(struct mScriptEngine*, const char* name, struct VFile* vf);
+static void CLIScriptEngineRun(struct mScriptEngine*);
+static bool CLIScriptEngineLookupSymbol(struct mScriptEngine*, const char* name, int32_t* out);
+static void CLIScriptDebuggerEntered(struct mScriptEngine*, enum mDebuggerEntryReason, struct mDebuggerEntryInfo*);
+
+struct CLIScriptStatement {
+	char* command;
+	size_t commandLen;
+	struct CLIScriptStatement* next;
```
endrift, line 25: Should probably use a Vector instead of a linked list here.

src/debugger/symbols.c
```diff
@@ -16,11 +17,13 @@ struct mDebuggerSymbol {
 
 struct mDebuggerSymbols {
 	struct Table names;
+	struct Table reverse;
```
endrift, line 20: This PR is missing the cleanup from this reverse table (including removing symbols).

src/util/table.c
```diff
@@ -203,10 +213,32 @@ void HashTableInsert(struct Table* table, const char* key, void* value) {
 	++table->size;
 }
 
+void HashTableInsertBinary(struct Table* table, const void* key, size_t keylen, void* value) {
```
endrift, line 216: This PR is missing `HashTableRemoveBinary`.

src/debugger/cli-debugger-scripting.c
endrift: whoops you fixed this before I finished the review

src/debugger/cli-debugger-scripting.c
ahigerd: Oh right. That was a placeholder until I could talk to you about that. Suggestions?

ahigerd: `.mds` for "m(GBA|edusa) debugger script"?

endrift: Maybe `.mrc` for "m(GBA|edusa) run commands"? Not entirely sure.

ahigerd: ... so _that's_ what rc stands for. I always thought it meant resource!

Sure. Sounds good to me.

PR #1850, opened 2020-08-10
Stack traces: fix tracing of indirect jumps and interrupt handlers

ahigerd:
I noticed totally bogus pointers in stack traces and it turns out I had misunderstood what the decoder was referring to as "indirect" branches. Also having interrupt handlers switch around the stack pointer was causing frames to be spuriously popped.

### DIFF
--- a/src/arm/debugger/debugger.c
+++ b/src/arm/debugger/debugger.c
@@ -29,6 +29,16 @@ static bool ARMDecodeCombined(struct ARMCore* cpu, struct ARMInstructionInfo* in
 	}
 }
 
+static inline int ARMGetStack(struct ARMRegisterFile* regs) {
+	if (!regs) {
+		return MODE_USER;
+	}
+	if (regs->cpsr.priv == MODE_SYSTEM) {
+		return MODE_USER;
+	}
+	return regs->cpsr.priv;
+}
+
 static bool ARMDebuggerUpdateStackTraceInternal(struct mDebuggerPlatform* d, uint32_t pc) {
 	struct ARMDebugger* debugger = (struct ARMDebugger*) d;
 	struct ARMCore* cpu = debugger->cpu;
@@ -36,7 +46,8 @@ static bool ARMDebuggerUpdateStackTraceInternal(struct mDebuggerPlatform* d, uin
 	struct mStackTrace* stack = &d->p->stackTrace;
 
 	struct mStackFrame* frame = mStackTraceGetFrame(stack, 0);
-	if (frame && frame->frameBaseAddress < (uint32_t) cpu->gprs[ARM_SP]) {
+	int currentStack = ARMGetStack(&cpu->regs);
+	if (frame && frame->frameBaseAddress < (uint32_t) cpu->gprs[ARM_SP] && currentStack == ARMGetStack(frame->regs)) {
 		// The stack frame has been popped off the stack. This means the function
 		// has been returned from, or that the stack pointer has been otherwise
 		// manipulated. Either way, the function is done executing.
@@ -45,7 +56,7 @@ static bool ARMDebuggerUpdateStackTraceInternal(struct mDebuggerPlatform* d, uin
 			shouldBreak = shouldBreak || frame->breakWhenFinished;
 			mStackTracePop(stack);
 			frame = mStackTraceGetFrame(stack, 0);
-		} while (frame && frame->frameBaseAddress < (uint32_t) cpu->gprs[ARM_SP]);
+		} while (frame && frame->frameBaseAddress < (uint32_t) cpu->gprs[ARM_SP] && currentStack == ARMGetStack(frame->regs));
 		if (shouldBreak) {
 			struct mDebuggerEntryInfo debuggerInfo = {
 				.address = pc,
@@ -84,10 +95,10 @@ static bool ARMDebuggerUpdateStackTraceInternal(struct mDebuggerPlatform* d, uin
 		return false;
 	}
 
-	bool isCall = interrupt || (info.branchType & ARM_BRANCH_LINKED);
+	bool isCall = (info.branchType & ARM_BRANCH_LINKED);
 	uint32_t destAddress;
 
-	if (interrupt && info.branchType == ARM_BRANCH_NONE) {
+	if (interrupt && !isCall) {
 		// The stack frame was already pushed up above, so there's no
 		// action necessary here, but we still want to check for a
 		// breakpoint down below.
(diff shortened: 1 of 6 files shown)


src/arm/debugger/debugger.c
```diff
@@ -29,14 +29,25 @@ static bool ARMDecodeCombined(struct ARMCore* cpu, struct ARMInstructionInfo* in
 	}
 }
 
+static inline int ARMGetStack(struct ARMRegisterFile* regs) {
```
endrift, line 32: arm.c has `_ARMSelectBank` for this, which could be moved to isa-inlines.h.

src/arm/debugger/debugger.c
```diff
@@ -84,10 +95,10 @@ static bool ARMDebuggerUpdateStackTraceInternal(struct mDebuggerPlatform* d, uin
 		return false;
 	}
 
-	bool isCall = interrupt || (info.branchType & ARM_BRANCH_LINKED);
+	bool isCall = (info.branchType & ARM_BRANCH_LINKED);
```
endrift, line 98: Nit: Remove the parens.

src/arm/debugger/debugger.c
```diff
@@ -115,6 +126,9 @@ static bool ARMDebuggerUpdateStackTraceInternal(struct mDebuggerPlatform* d, uin
 			bool isBranch = ARMInstructionIsBranch(info.mnemonic);
 			int reg = (isBranch ? info.op1.reg : info.op2.reg);
 			destAddress = cpu->gprs[reg];
+			if ((info.operandFormat & ARM_OPERAND_MEMORY_2) && (info.branchType & ARM_BRANCH_INDIRECT)) {
+				destAddress = cpu->memory.load32(cpu, destAddress, NULL);
```
endrift, line 130: This won't work; you'll need to parse `info.memory`, which I don't think there's a helper for. However, I should add one at some point. Though, I'm not entirely clear on if this is running before or after the instruction is executed, so it might already be in `pc` (if `info.op1.reg` is 15 that is) by the time this is called.

ahigerd: With this last push, I'm about 98% sure I've done this right but in particular I can't quite tell how preincrement is supposed to work.

src/arm/decoder.c
```diff
@@ -488,3 +488,43 @@ int ARMDisassemble(struct ARMInstructionInfo* info, uint32_t pc, char* buffer, i
 	buffer[blen - 1] = '\0';
 	return total;
 }
+
+uint32_t ARMResolveMemoryAccess(struct ARMInstructionInfo* info, struct ARMRegisterFile* regs, uint32_t pc) {
+	uint32_t address = 0;
+	int32_t offset = 0;
+	if (info->memory.format & ARM_MEMORY_REGISTER_BASE) {
+		if (info->memory.baseReg == ARM_PC && info->memory.format & ARM_MEMORY_IMMEDIATE_OFFSET) {
+			address = pc & 0xFFFFFFFC;
```
endrift, line 497: This is incorrect for Thumb. Also its force-aligned already so you don't need to mask it.

src/arm/decoder.c
```diff
@@ -488,3 +488,43 @@ int ARMDisassemble(struct ARMInstructionInfo* info, uint32_t pc, char* buffer, i
...
+		} else {
+			address = regs->gprs[info->memory.baseReg];
+		}
+	}
+	if (info->memory.format & ARM_MEMORY_IMMEDIATE_OFFSET) {
+		offset = info->memory.offset.immediate;
+	} else if (info->memory.format & ARM_MEMORY_REGISTER_OFFSET) {
+		offset = regs->gprs[info->memory.offset.reg];
```
endrift, line 505: This can be `pc`.

src/arm/decoder.c
```diff
@@ -488,3 +488,43 @@ int ARMDisassemble(struct ARMInstructionInfo* info, uint32_t pc, char* buffer, i
...
+				offset = ROR(offset, shiftSize);
+				break;
+			case ARM_SHIFT_RRX:
+				offset = (regs->cpsr.c << 31) | ((uint32_t) offset >> 1);
+				break;
+			default:
+				break;
+		};
+	}
+	return address + (info->memory.format & ARM_MEMORY_OFFSET_SUBTRACT ? -offset : offset);
```
endrift, line 529: Offset should be ignored for post-increment.

src/arm/debugger/debugger.c
```diff
@@ -126,8 +118,9 @@ static bool ARMDebuggerUpdateStackTraceInternal(struct mDebuggerPlatform* d, uin
 			bool isBranch = ARMInstructionIsBranch(info.mnemonic);
 			int reg = (isBranch ? info.op1.reg : info.op2.reg);
 			destAddress = cpu->gprs[reg];
-			if ((info.operandFormat & ARM_OPERAND_MEMORY_2) && (info.branchType & ARM_BRANCH_INDIRECT)) {
-				destAddress = cpu->memory.load32(cpu, destAddress, NULL);
+			if (!isBranch && info.branchType & ARM_BRANCH_INDIRECT && info.op1.reg == ARM_PC && info.operandFormat & ARM_OPERAND_MEMORY_2) {
```
endrift, line 121: Wouldn't the mixing of `&` and `&&` generate a warning?

src/arm/decoder.c
```diff
@@ -488,3 +488,43 @@ int ARMDisassemble(struct ARMInstructionInfo* info, uint32_t pc, char* buffer, i
...
+			address = regs->gprs[info->memory.baseReg];
+		}
+	}
+	if (info->memory.format & ARM_MEMORY_IMMEDIATE_OFFSET) {
+		offset = info->memory.offset.immediate;
+	} else if (info->memory.format & ARM_MEMORY_REGISTER_OFFSET) {
+		offset = regs->gprs[info->memory.offset.reg];
+	}
+	if (info->memory.format & ARM_MEMORY_SHIFTED_OFFSET) {
+		uint8_t shiftSize = (info->operandFormat & ARM_OPERAND_SHIFT_REGISTER_1) ? regs->gprs[info->memory.offset.shifterReg] : info->memory.offset.shifterImm;
```
endrift, line 508: This check is unneeded; the memory operand can never be shifted by a register, only by an immediate.

src/arm/debugger/debugger.c
endrift: Pre-increment means that the increment that's done is applied to the address before the actual access (and then written back), post-increment means the increment is done after the access.

src/arm/debugger/debugger.c
ahigerd: You'd think, but apparently not. I'll put parens on it anyway.

src/arm/decoder.c
ahigerd: Oh right, I don't need to account for the switch-to-Thumb low bit on the _current_ pc value.

src/arm/decoder.c
ahigerd: Oh, I suppose that does make sense, thinking about it long enough. Took several "but what about..." moments before I realized that there really isn't a good use case for generically allowing variable pointer shifting, and in the odd chance you might it you can just do it in separate instruction.

ahigerd:
Thanks for the feedback. I kinda guessed there would be some rough edges, but I didn't know enough about the architecture to know where they'd be. Glad to see I was mostly right.

Issue #1260, opened 2019-01-08
Automatically load matching No$GBA-style .sym files when loading a ROM


Zeturic:
No$GBA uses `.sym` files to load information about the symbols in a ROM. For example, if you have the following:

```00000000 0
80000000 .dbl:0010
80000010 main
8000002C subroutine
80240000 newblock
```

In a file named `rom.sym` and then open `rom.gba` in No$GBA, it will load those symbols and then you can use them for setting breakpoints and jump to them in the disassembly view (etc).

Would it be possible to include this in mGBA?

endrift:
Symbol files should already be loaded, but the symbols don't show up in disassembly yet.

Sagiri:
I swear I tried and they weren't being recognized for breakpoints in the debug menu. I'm probably just on an outdated build, though.

Zeturic:
PPSSPP and PCSX2 have a slight variation on the format which can include the lengths of functions. For example:

```
00000000 0
80000000 .dbl:0010
80000010 Main
8000002C Subroutine,0000001C
80240000 NewBlock,00000014
```

Would it be possible for mGBA to support this format as well? I can't think of anything mGBA could do with the function lengths (especially since they're not guaranteed for every label), so it'd basically just be throwing away everything after an including a comma.

endrift:
Both parts of this issue should now be fixed.

PR #1878, opened 2020-09-08
Support variant symfile format with function sizes

Zeturic:
This basically implements what I asked for a while back in comment on #1260, adding support for a slight variation on the symfile format, which includes sizes. ARMIPS outputs this format when supplied with `-sym2` (as opposed to `-sym`); the ARMIPS readme says that it was originally from PCSX2 and PPSSPP, and gives the following example:

```
00000000 0
80000000 .dbl:0010
80000010 Main
8000002C Subroutine,0000001C
80240000 NewBlock,00000014
```

As is, mGBA will parse that and create symbols called `Subroutine,0000001C` and `NewBlock,00000014`. This PR basically just checks for a comma in the name of a symbol in a symfile and throws away it and everything after it (so they'd be correctly identified as `Subroutine` and `NewBlock`).

It's a fairly trivial change, but I'm not overly confidant in C, so hopefully I didn't make any obvious errors. Let me know if there's anything that needs changing.

### DIFF
--- a/src/debugger/symbols.c
+++ b/src/debugger/symbols.c
@@ -97,6 +97,13 @@ void mDebuggerLoadARMIPSSymbols(struct mDebuggerSymbols* st, struct VFile* vf) {
 			continue;
 		}
 
+		char* buf2 = strchr(buf, ',');
+
+		if (buf2 != NULL) {
+			// Commas separate names from function sizes
+			*buf2 = '\0';
+		}
+
 		mDebuggerSymbolAdd(st, buf, address, -1);
 	}
 }


src/debugger/symbols.c
```diff
@@ -76,8 +76,8 @@ void mDebuggerLoadARMIPSSymbols(struct mDebuggerSymbols* st, struct VFile* vf) {
 			line[bytesRead - 1] = '\0';
 		}
 		uint32_t address = 0;
-		const char* buf = line;
-		buf = hex32(buf, &address);
+		char* buf = line;
+		buf = (char*)hex32(buf, &address);
```
endrift, line 80: `strchr` returns `char*` in C even with a `const char*` argument, so this isn't needed.

src/debugger/symbols.c
```diff
@@ -97,6 +97,13 @@ void mDebuggerLoadARMIPSSymbols(struct mDebuggerSymbols* st, struct VFile* vf) {
 			continue;
 		}
 
+        char* buf2;
+
+        if ((buf2 = strchr(buf, ',')) != NULL) {
+            // Commas separate names from function sizes
+            *buf2 = '\0';
+        }
```
endrift, line 105: Tabs, not spaces. Also, move the assignment to the variable declaration. There's no reason it needs to be inside the `if`, so it shouldn't be there.

endrift (changes requested):
Some nitpicks, but the general implementation seems fine.

Zeturic:
Really sorry about all of those force pushes. I confused myself by switching between two PCs, but it's correct now.

Also, thank you for such a quick reply.

PR #1886, opened 2020-09-12
Qt: Always apply SmoothPixMapTransform to the background logo

lotharsm:
Currently, the background logo of the Qt application looks not very nice if the bilinear filter is disabled in the mGBA settings and tends to be pretty blurry.
mgba-no-antialias

This simple patch always enables the `SmoothPixmapTransform` rendering hint for drawing the background logo, so the hinting filter will always be applied, resulting in a much cleaner image of the background logo regardless of the bilinear filter for the emulation being enabled or disabled.
mgba-with-antialias

Any chance to get this merged or am I breaking something here? :-)

### DIFF
--- a/src/platform/qt/Window.cpp
+++ b/src/platform/qt/Window.cpp
@@ -1978,7 +1978,7 @@ void WindowBackground::paintEvent(QPaintEvent* event) {
 	QWidget::paintEvent(event);
 	const QPixmap& logo = pixmap();
 	QPainter painter(this);
-	painter.setRenderHint(QPainter::SmoothPixmapTransform, m_filter);
+	painter.setRenderHint(QPainter::SmoothPixmapTransform, true);
 	painter.fillRect(QRect(QPoint(), size()), Qt::black);
 	QSize s = size();
 	QSize ds = s;


endrift:
This breaks the still image used when the game is paused or behind the savestate select, see `Window::updateFrame`. The correct thing to do would be to call `m_screenWidget->filter(false)` in the constructor and `Window::gameStopped`. That said, that code should also be refactored out because it's common between the two functions.

lotharsm:
Ah, thanks! Would have been too easy :D

endrift:
Anything that calls `Window::updateFrame` expects that code you changed to have the same filtering as the game. It might only be noticeable if you use the Software display driver. Not sure.

endrift:
Are you still looking into this? I can probably fix it myself if you're having trouble reproducing the case I described.

lotharsm:
I was pretty busy in the past days, so this issue was a bit 'on hold' - however, I'd like to look at it in the next couple of days, so I can start gaining some experience :-)

endrift:
Any update?

lotharsm:
Sorry, I'm afraid I won't be able to take a look at this issue soon :/

endrift:
Do you mind if I just fix it on my end then? I'd like to include a fix for this in 0.8.4, since it should be a safe fix, and I'd like to release that soon.

lotharsm:
It wouldn't make any sense to hold back a release just waiting for me trying to fix stuff while it seems that you already know the solution to the problem, so please, go ahead 😃

endrift:
It wouldn't be holding back the release, it would be skipping fixing it for the release, but yeah I'll go ahead and do that soon then. Thanks.

PR #1910, opened 2020-10-15
OpenEmu: Update implementation

clobber:
* Fix cheat handling.
* Rely on cheat autodetection.
* Use direct-rendering API.
* Use `-fileSystemRepresentation` instead of `-UTF8String` for file names.
* Limit rewind buffer to 60 seconds to reduce memory use.
* Remove calls to `release`/`dealloc`; OpenEmu uses ARC.


endrift (commented):
The OpenEmu core can still be built with the CMake flow upstream. Since this moves to ARC, I'd recommend adding `-fobjc-arc` to `CMAKE_OBJC_FLAGS` or similar in the CMakeLists.

endrift:
Adding a line `COMPILE_FLAGS "-fobjc-arc"` to the `set_target_properties(${BINARY_NAME}-openemu PROPERTIES` statement around line 921 should do the job.

endrift:
Thanks!

PR #1931, opened 2020-11-07
Libretro: Only set camera peripheral when it is available

michaelforney:
Otherwise, if the frontend does not support GET_CAMERA_INTERFACE,
the start and stop functions are NULL, causing a crash when a gameboy
ROM is unloaded.


endrift:
I could have sworn that the downstream libretro branch had a fix for this, but I can't find it. I might have argued that the frontend should handle this. Regardless, this seems like a safe bet.

PR #1946, opened 2020-11-21
Return correct byte in out of bounds ROM load

jsmolka:
According to GBATEK unused memory is filled the following way:

> Because Gamepak uses the same signal-lines for both 16bit data and for lower 16bit halfword address, the entire gamepak ROM area is effectively filled by incrementing 16bit values (Address/2 AND FFFFh).

mGBA already adheres to that in word / halfword loads but the byte variant seems to be wrong. Currently it returns the same value for both bytes of the actual 16bit value. Please correct me if I am wrong here.

### DIFF
--- a/src/gba/memory.c
+++ b/src/gba/memory.c
@@ -679,7 +679,7 @@ uint32_t GBALoad8(struct ARMCore* cpu, uint32_t address, int* cycleCounter) {
 			value = GBAVFameGetPatternValue(address, 8);
 		} else {
 			mLOG(GBA_MEM, GAME_ERROR, "Out of bounds ROM Load8: 0x%08X", address);
-			value = (address >> 1) & 0xFF;
+			value = ((address >> 1) >> ((address & 1) * 8)) & 0xFF;
 		}
 		break;
 	case REGION_CART_SRAM:


endrift:
I'd like to verify that this is in fact a bug first, but you're probably correct.

endrift:
I wrote a quick test and verified that this does in fact fix a bug. Thanks for the fix! I forgot to ask you to update the changelog though, so I'll do that now. How should I credit you?

profi200:
Doesn't your test suite already do the ROM open bus test with bytes? If not it would make sense to add it.

endrift:
It does, but I guess it only tests one half of the 16-bit words for OOB ROMs. I should add it though, you're correct.

jsmolka:
@endrift There is no need to credit me for such a small contribution.

I do have another question though. Which setup are you using for hardware tests? I'm asking because I use an EZ-FLash Omega and some mGBA suite tests fail on it, especially the open bus / out of bounds ROM reads.

endrift:
The open bus tests in the suite don't really work on flashcarts. They need to be migrated into a separate multiboot ROM. In this case I just threw together a tiny multiboot ROM and loaded it onto my GBP using GBI.

profi200:
For non-gamecart hardware tests you can also use open_agb_firm on a 3DS (possibly the cheaper solution depending on how expensive a GBP + GameCube is these days). I say non-gamecart because the hardware can't simulate all cart sizes accurately. Only 16/32 MiB depending on the save type. As a workaround i do fake open bus padding (will break on sequential accesses).

PR #1953, opened 2020-11-28
Fix LIBMGBA_ONLY build

pdagobert:
Building with LIBMGBA_ONLY fails because USE_EPOXY is required for Windows build. This warning should probably be ignored when building with LIBMGBA_ONLY.


CMakeLists.txt
```diff
@@ -732,7 +732,7 @@ elseif(BUILD_GLES2)
 	set(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS},libgles2")
 endif()
 
-if(WIN32 AND NOT SKIP_LIBRARY AND NOT USE_EPOXY)
+if(WIN32 AND NOT LIBMGBA_ONLY AND NOT SKIP_LIBRARY AND NOT USE_EPOXY)
```
endrift, line 735: Might be more readable if it's turned into this:

```suggestion
if(WIN32 AND NOT (LIBMGBA_ONLY OR SKIP_LIBRARY OR USE_EPOXY))
```

But otherwise seems like a good idea.

endrift (approved):

PR #1961, opened 2020-12-02
Fix typo in SettingsView.ui

Margen67:


endrift:
You can't change the name of the field without changing where it's used. This is failing to compile as a result.

PR #1966, opened 2020-12-04
Qt: Fix overriding buttons setting when option is null.

### LINKED ISSUE #1965: Game name and framerate aren't shown when the display driver is set to software QT
Capture

Capture2

Testsr:
fixes #1965. Please edit if needed.

### DIFF
--- a/src/platform/qt/SettingsView.cpp
+++ b/src/platform/qt/SettingsView.cpp
@@ -563,8 +563,8 @@ void SettingsView::reloadConfig() {
 	loadSetting("gb.bios", m_ui.gbBios);
 	loadSetting("gbc.bios", m_ui.gbcBios);
 	loadSetting("sgb.bios", m_ui.sgbBios);
-	loadSetting("sgb.borders", m_ui.sgbBorders);
-	loadSetting("useCgbColors", m_ui.useCgbColors);
+	loadSetting("sgb.borders", m_ui.sgbBorders, true);
+	loadSetting("useCgbColors", m_ui.useCgbColors, true);
 	loadSetting("useBios", m_ui.useBios);
 	loadSetting("skipBios", m_ui.skipBios);
 	loadSetting("audioBuffers", m_ui.audioBufferSize);
@@ -577,9 +577,9 @@ void SettingsView::reloadConfig() {
 	loadSetting("lockAspectRatio", m_ui.lockAspectRatio);
 	loadSetting("lockIntegerScaling", m_ui.lockIntegerScaling);
 	loadSetting("interframeBlending", m_ui.interframeBlending);
-	loadSetting("showOSD", m_ui.showOSD);
+	loadSetting("showOSD", m_ui.showOSD, true);
 	loadSetting("volume", m_ui.volume, 0x100);
-	loadSetting("mute", m_ui.mute);
+	loadSetting("mute", m_ui.mute, false);
 	loadSetting("fastForwardVolume", m_ui.volumeFf, m_ui.volume->value());
 	loadSetting("fastForwardMute", m_ui.muteFf, m_ui.mute->isChecked());
 	loadSetting("rewindEnable", m_ui.rewind);
@@ -596,12 +596,12 @@ void SettingsView::reloadConfig() {
 	loadSetting("cheatsPath", m_ui.cheatsPath);
 	loadSetting("showLibrary", m_ui.showLibrary);
 	loadSetting("preload", m_ui.preload);
-	loadSetting("showFps", m_ui.showFps);
-	loadSetting("cheatAutoload", m_ui.cheatAutoload);
-	loadSetting("cheatAutosave", m_ui.cheatAutosave);
-	loadSetting("showFilename", m_ui.showFilename);
-	loadSetting("autoload", m_ui.autoload);
-	loadSetting("autosave", m_ui.autosave);
+	loadSetting("showFps", m_ui.showFps, true);
+	loadSetting("cheatAutoload", m_ui.cheatAutoload, true);
+	loadSetting("cheatAutosave", m_ui.cheatAutosave, true);
+	loadSetting("showFilename", m_ui.showFilename, false);
+	loadSetting("autoload", m_ui.autoload, true);
+	loadSetting("autosave", m_ui.autosave, false);
 	loadSetting("logToFile", m_ui.logToFile);
 	loadSetting("logToStdout", m_ui.logToStdout);
 	loadSetting("logFile", m_ui.logFile);
@@ -735,9 +735,10 @@ void SettingsView::saveSetting(const char* key, const QVariant& field) {
 	m_controller->setOption(key, field);
 	m_controller->updateOption(key);
 }
+
 void SettingsView::loadSetting(const char* key, QAbstractButton* field) {
 	QString option = loadSetting(key);
-	if(option.isNull()){
+	if (option.isNull()) {
 		return;
 	}
 	field->setChecked(option != "0");
--- a/src/platform/qt/SettingsView.cpp
+++ b/src/platform/qt/SettingsView.cpp
@@ -563,8 +563,8 @@ void SettingsView::reloadConfig() {
 	loadSetting("gb.bios", m_ui.gbBios);
 	loadSetting("gbc.bios", m_ui.gbcBios);
 	loadSetting("sgb.bios", m_ui.sgbBios);
-	loadSetting("sgb.borders", m_ui.sgbBorders, true);
-	loadSetting("useCgbColors", m_ui.useCgbColors, true);
+	loadSetting("sgb.borders", m_ui.sgbBorders);
+	loadSetting("useCgbColors", m_ui.useCgbColors);
 	loadSetting("useBios", m_ui.useBios);
 	loadSetting("skipBios", m_ui.skipBios);
 	loadSetting("audioBuffers", m_ui.audioBufferSize);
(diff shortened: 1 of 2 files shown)


endrift:
I don't get it. This removes functionality that was added intentionally, and is also not the correct fix.

Testsr:
What functionality, exactly?

endrift:
The one where you can actually set the default value? As opposed to just baking it into the UI, which is only there as a guideline, not as the ground truth.

Not to mention the fact that it has not one, but three style guide violations.

Testsr:
The one where you can actually set the default value?
 Yes, you can still do that.   
`loadSetting("fastForwardMute", m_ui.muteFf, m_ui.mute->isChecked());`

Testsr:
I should maybe name the commit better.

Testsr:
Not to mention that the guidelines

Testsr:
@endrift, What style guides violation, are there?

src/platform/qt/SettingsView.cpp
```diff
@@ -563,8 +563,8 @@ void SettingsView::reloadConfig() {
 	loadSetting("gb.bios", m_ui.gbBios);
 	loadSetting("gbc.bios", m_ui.gbcBios);
 	loadSetting("sgb.bios", m_ui.sgbBios);
-	loadSetting("sgb.borders", m_ui.sgbBorders, true);
```
endrift, line 566: This should not be removed; it's for setting the default when no option is present and entirely unrelated to the bug in question. Why did you remove it?

src/platform/qt/SettingsView.cpp
```diff
@@ -563,8 +563,8 @@ void SettingsView::reloadConfig() {
...
-	loadSetting("useCgbColors", m_ui.useCgbColors, true);
```
endrift, line 567: This should not be removed; it's for setting the default when no option is present and entirely unrelated to the bug in question. Why did you remove it?

src/platform/qt/SettingsView.cpp
```diff
@@ -577,9 +577,9 @@ void SettingsView::reloadConfig() {
 	loadSetting("lockAspectRatio", m_ui.lockAspectRatio);
 	loadSetting("lockIntegerScaling", m_ui.lockIntegerScaling);
 	loadSetting("interframeBlending", m_ui.interframeBlending);
-	loadSetting("showOSD", m_ui.showOSD, true);
```
endrift, line 580: This should not be removed; it's for setting the default when no option is present and entirely unrelated to the bug in question. Why did you remove it?

src/platform/qt/SettingsView.cpp
```diff
@@ -577,9 +577,9 @@ void SettingsView::reloadConfig() {
...
+	loadSetting("showOSD", m_ui.showOSD);
 	loadSetting("volume", m_ui.volume, 0x100);
-	loadSetting("mute", m_ui.mute, false);
```
endrift, line 582: This should not be removed; it's for setting the default when no option is present and entirely unrelated to the bug in question. Why did you remove it?

src/platform/qt/SettingsView.cpp
```diff
@@ -596,12 +596,12 @@ void SettingsView::reloadConfig() {
 	loadSetting("cheatsPath", m_ui.cheatsPath);
 	loadSetting("showLibrary", m_ui.showLibrary);
 	loadSetting("preload", m_ui.preload);
-	loadSetting("showFps", m_ui.showFps, true);
```
endrift, line 599: This should not be removed; it's for setting the default when no option is present and entirely unrelated to the bug in question. Why did you remove it?

src/platform/qt/SettingsView.cpp
```diff
@@ -596,12 +596,12 @@ void SettingsView::reloadConfig() {
...
-	loadSetting("cheatAutoload", m_ui.cheatAutoload, true);
```
endrift, line 600: This should not be removed; it's for setting the default when no option is present and entirely unrelated to the bug in question. Why did you remove it?

src/platform/qt/SettingsView.cpp
```diff
@@ -596,12 +596,12 @@ void SettingsView::reloadConfig() {
...
-	loadSetting("cheatAutosave", m_ui.cheatAutosave, true);
```
endrift, line 601: This should not be removed; it's for setting the default when no option is present and entirely unrelated to the bug in question. Why did you remove it?

src/platform/qt/SettingsView.cpp
```diff
@@ -596,12 +596,12 @@ void SettingsView::reloadConfig() {
...
-	loadSetting("showFilename", m_ui.showFilename, false);
```
endrift, line 602: This should not be removed; it's for setting the default when no option is present and entirely unrelated to the bug in question. Why did you remove it?

src/platform/qt/SettingsView.cpp
```diff
@@ -596,12 +596,12 @@ void SettingsView::reloadConfig() {
...
-	loadSetting("autoload", m_ui.autoload, true);
```
endrift, line 603: This should not be removed; it's for setting the default when no option is present and entirely unrelated to the bug in question. Why did you remove it?

src/platform/qt/SettingsView.cpp
```diff
@@ -596,12 +596,12 @@ void SettingsView::reloadConfig() {
...
-	loadSetting("autosave", m_ui.autosave, false);
```
endrift, line 604: This should not be removed; it's for setting the default when no option is present and entirely unrelated to the bug in question. Why did you remove it?

src/platform/qt/SettingsView.cpp
```diff
@@ -735,7 +735,12 @@ void SettingsView::saveSetting(const char* key, const QVariant& field) {
 	m_controller->setOption(key, field);
 	m_controller->updateOption(key);
 }
-
```
endrift, line 738: Line break needed between functions.

Why did you add a version of this function that doesn't have a default setting? The UI is just a guideline and does not hold what the ground truth for these default should be, but with this change that's where the defaults are coming from.

src/platform/qt/SettingsView.cpp
```diff
@@ -735,7 +735,12 @@ void SettingsView::saveSetting(const char* key, const QVariant& field) {
...
+void SettingsView::loadSetting(const char* key, QAbstractButton* field) {
+	QString option = loadSetting(key);
+	if(option.isNull())
```
endrift, line 740: Space between `if` and the `(`

src/platform/qt/SettingsView.cpp
```diff
@@ -735,7 +735,12 @@ void SettingsView::saveSetting(const char* key, const QVariant& field) {
...
+		return;
```
endrift, line 741: No bare statements. Needs braces.

src/platform/qt/SettingsView.cpp
```diff
@@ -735,7 +735,12 @@ void SettingsView::saveSetting(const char* key, const QVariant& field) {
...
+	field->setChecked(option != "0");
+}
```
endrift, line 743: Line break needed between functions.

endrift (changes requested):
Ok, I pointed out the style nits, as well as every unrelated and seemingly pointless (as well as breaking) change you made in this PR.

It's every line.

src/platform/qt/SettingsView.cpp
```diff
@@ -735,6 +735,13 @@ void SettingsView::saveSetting(const char* key, const QVariant& field) {
 	m_controller->setOption(key, field);
 	m_controller->updateOption(key);
 }
+void SettingsView::loadSetting(const char* key, QAbstractButton* field) {
```
endrift, line 738: Still need a line break here.

src/platform/qt/SettingsView.cpp
```diff
@@ -735,6 +735,13 @@ void SettingsView::saveSetting(const char* key, const QVariant& field) {
...
+	QString option = loadSetting(key);
+	if(option.isNull()){
```
endrift, line 740: Still need that space between `if` and `(`, and now it also needs the space between `)` and `{`.

If you want to contribute to this project, or really any other project, you should take a look at surrounding code and match its style. There's a contributing.md that contains some very basic style rules that were not followed here.

endrift (changes requested):
You didn't resolve all of the nits I pointed out last time and you added a new issue...

src/platform/qt/SettingsView.cpp
Testsr: Why should the UI be just a guideline? The only example that would warrant overriding it, IMO would be fastFowardMute.

endrift: Why would the UI not be just a guideline? It's a terrible way to view defaults since you need a visual editor or skimming through lots and lots of XML, and is several levels detached from the actual names of the settings. The actual defaults are dictated by code, and this SettingsView class doesn't even dictate what the defaults are where they're actually used: it already needs to be kept in sync. Making the UI need to be kept in sync too is just a headache and hard to debug.

src/platform/qt/SettingsView.cpp
endrift, line 567: You left out the comma when you "reverted" this.

endrift (changes requested):
You still haven't told me what motivates this change; it still migrates defaults to the UI XML from this file on the offhand chance that they differ, which now means another place to look when something goes wrong, increasing maintenance burden. It doesn't actually fix any bugs in the process, either, despite your claim: it just, as I said, migrated the default, which coincidentally fixed it, instead of setting the default the way that's done within the framework you're changing.

Testsr:
> You still haven't told me what motivates this change; it still migrates defaults to the UI XML from this file on the offhand chance that they differ, which now means another place to look when something goes wrong, increasing maintenance burden. It doesn't actually fix any bugs in the process, either, despite your claim: it just, as I said, migrated the default, which coincidentally fixed it, instead of setting the default the way that's done within the framework you're changing.

It does not "migrates defaults to the UI XML from this file on the offhand chance that they differ", instead it migrates defaults if this file does not provide a default.

endrift:
That's not a meaningful distinction. You've yet to actually defend this design change, and at this point I don't think you have a reason for it other than you changed a thing to see if it worked and it did without realizing what the change did or why it worked.

Also I fixed the bug already without changing the design.

Testsr:
The change are just for the cases where you forget to set a item to true.

Testsr:
I can however leave this failed pull request behind.

Testsr:
> That's not a meaningful distinction. You've yet to actually defend this design change, and at this point I don't think you have a reason for it other than you changed a thing to see if it worked and it did without realizing what the change did or why it worked.
> 
> Also I fixed the bug already without changing the design.
This change helps prevent the cases where you forget to set a new setting's default to true and therefore the function sets it to false by default. Having two places to look at can be a problem, however, so this is fine closed.

Issue #1973, opened 2020-12-08
The win64 installer exe still says 0.8.3


TheGinGear:
![image](https://user-images.githubusercontent.com/39550332/101538024-1eb9bf80-3962-11eb-9a5b-73ec9bc8ca77.png)
It still installs 0.8.4 normally, It's literally just a typo

endrift:
It's not a typo (technically speaking), and it can't be fixed without rolling a new release. See discussion in #1923.

PR #1976, opened 2020-12-10
Windows installer updates

endrift:
Not actually familiar with Inno Setup that much so I made some changes that I'd like @lotharsm to look over before they get merged.

- The easy one is just detecting if the iss script has changed. Nothing much to see here, it's just on this branch because it's related
- The version string can be out of sync with major/minor/patch sometimes (either by mistake, see #1973, or in the case of a tag, e.g. 0.9-b1, intentionally different). I don't know if what I did is _allowable_ for where it can be text instead of a 2/3/4 byte version tuple.
- The image was bothering me. Also it wasn't high DPI. This one is better, I think, but I'm still not thrilled with it.


lotharsm (approved):
Those changes look good to me, I don't see any potential issues here.

endrift:
I'm unhappy about having an uncompressed bitmap in the source. Does inno setup really only support bitmaps?

endrift:
There's also a "modern" style apparently, but when I tried to use it the installer generated didn't work. I might need additional files?

lotharsm:
Unfortunately, the website of the Inno setup creators is currently down for me, so I can't send you any "proof" - but I know that their documentation states that the only supported file type is .bmp for now and IIRC this is unlikely to change in the future.

Usually, the current version of Inno setup should compile the "Modern style" just fine, I'm using it for packaging ScummVM too - I just checked my installation and I can't remember I had to add any additional files besides the installer package itself.

Could you give me a few more details?

endrift:
the installer it built said something about "missing files" when I ran it. I might have used some reduced set of installer files in the docker image, I guess. I can get the exact error tomorrow.

PR #1977, opened 2020-12-10
Qt: General fixes to Simp. Chinese translation

EmpyreusX:
- New translations
- Delete unused entries
- Fix typos and invaild lines

### DIFF
--- a/src/platform/qt/ts/mgba-zh_CN.ts
+++ b/src/platform/qt/ts/mgba-zh_CN.ts
@@ -1176,7 +1176,7 @@ Game Boy Advance 是任天堂有限公司（Nintendo Co., Ltd.）的注册商标
         <source>Memory bank controller</source>
         <translation>内存 bank 控制器</translation>
     </message>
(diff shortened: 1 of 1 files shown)


src/platform/qt/ts/mgba-zh_CN.ts
```diff
@@ -832,72 +827,77 @@ Game Boy Advance 是任天堂有限公司（Nintendo Co., Ltd.）的注册商标
...
     <message>
-        <location filename="../MemoryView.ui" line="61"/>
+        <location filename="../MemoryView.ui" line="58"/>
+        <source>:</source>
+        <translation>:</translation>
+    </message>
+    <message>
+        <location filename="../MemoryView.ui" line="68"/>
         <source>0x</source>
         <translation>0</translation>
```
endrift, line 837: I should probably remove this string, but I'm pretty sure this should stay as 0x--it's a hexadecimal prefix marker.

PR #1978, opened 2020-12-11
Simplified MSYS2 intructions

hgdagon:
Not a lot changed. Most importantly, it's `pkgconf` now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,13 +163,7 @@ Note that you should not do a `make install` on macOS, as it will not work prope
 
 To build on Windows for development, using MSYS2 is recommended. Follow the installation steps found on their [website](https://msys2.github.io). Make sure you're running the 32-bit version ("MSYS2 MinGW 32-bit") (or the 64-bit version "MSYS2 MinGW 64-bit" if you want to build for x86_64) and run this additional command (including the braces) to install the needed dependencies (please note that this involves downloading over 1100MiB of packages, so it will take a long time):
 
-For x86 (32 bit) builds:
-
-	pacman -Sy --needed base-devel git mingw-w64-i686-{cmake,ffmpeg,gcc,gdb,libelf,libepoxy,libzip,pkg-config,qt5,SDL2,ntldd-git}
-
-For x86_64 (64 bit) builds:
-
-	pacman -Sy --needed base-devel git mingw-w64-x86_64-{cmake,ffmpeg,gcc,gdb,libelf,libepoxy,libzip,pkg-config,qt5,SDL2,ntldd-git}
+	pacman -Sy --needed base-devel git ${MINGW_PACKAGE_PREFIX}-{cmake,ffmpeg,gcc,gdb,libelf,libepoxy,libzip,pkgconf,qt5,SDL2,ntldd-git}
 
 Check out the source code by running this command:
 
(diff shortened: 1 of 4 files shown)


endrift:
Unfortunately `arch` always seems to report i686 for me, even in the MINGW64 build environment, so that won't actually work to replace both here.

Further I'm not fond of merging the multiple commands into one blob that may be impenetrable to new users. Turning it into this:

```sh
mkdir -p mgba/build
cd mgba/build
```

from the three commands that replaces is fine, but the remainder should stay separated. Further, the `-j$(nprocs)` flag always uses all cores, which may be undesirable given that it can cause the computer to become laggy if they weren't expecting it. `-j$(nprocs --ignore=1)` is probably okay though, and having it use more than one core by default is a good idea.

hgdagon:
@endrift On a 64-bit machine, `arch` should return `i686` only in the MinGW32 shell... Not entirely sure why you're facing that issue.  It works fine on my side. Will change CMake line in a bit.

endrift:
I'm not sure either, but it's definitely a 64-bit installation of Windows 10 with the MINGW64 shell. It's also a fairly old msys2 installation so it might have some weird bit rot.

hgdagon:
@endrift That might be it, since I always keep mine up-to-date.

Pushed the changes. I'd prefer to keep `arch`, and revert it, only if others face the issue as well.

endrift:
I'm going to look into what's going on my end before I merge this.

Testsr:
Quick note. Its $(nproc --ignore=1), not $(nprocs --ignore=1).

hgdagon:
@Testsr That's what I get for copying. Fixed.

endrift:
Sorry about that. Have been kind of frazzled all day.

endrift:
So I checked and it turns out I had a 32-bit installation of MSYS2, but MinGW worked as either 32-bit or 64-bit. Yet, installing a 64-bit installation of MSYS2 had the opposite problem--it always said x86_64 regardless of the MinGW environment in use. That said, you _can_ get that value from env variable `$MSYSTEM_CARCH`, which might work better here.

hgdagon:
That's interesting, I now have the same problem, `arch` returns `x86_64` in the MinGW32 shell. This wasn't the case 2 days ago. I'm 100% sure, since I had to check when you mentioned you had that issue. I *did* update my msys2 yesterday, so maybe something keeps breaking it?

Anyway, if we're sure `$MSYSTEM_CARCH`'s output is consistent, I'll go ahead and change to that.

I offered that solution to so many projects, now I have to go back and fix those... 🤦‍♂️

endrift:
I believe `MSYSTEM_CARCH` is explicitly set by `/etc/msystem` when opening the shell...so unless they change the mechanism for switching between MSYS2, MinGW32, and MinGW64, I doubt it'll change. That said, it's not impossible. Moving targets are fun!

hgdagon:
Hmm... well, then maybe we should straight up use `$MINGW_PACKAGE_PREFIX`?

endrift:
I didn't know that existed. That sounds like the right approach, yeah.

hgdagon:
That one's unlikely to ever change (at least without major bells and whistles), since that's what they use in the `PKGBUILD` files.

endrift:
Looks good now. Can you rebase/squash some commits and make the changes to the Chinese README that landed while this was in review?

hgdagon:
I'll just revert and make 1 commit with all 4 readme's (and hope that the Chinese one does mention using the appropriate shell).

endrift:
I don't think this got squashed properly. GitHub is a bit confused, I think.

hgdagon:
Sorry. I messed it up. Github Desktop is a little counter-intuitive.

PR #1981, opened 2020-12-11
New README_ZH_CN.md and updates to Simp. Chinese and Japanese translation

EmpyreusX:
- New ReadMe file (`README_ZH_CN.md`) in Simp. Chinese.
- Updated Simp. Chinese translation.
- Revamped (but partially) Japanese translation based on `mgba-template.ts`.

### DIFF
--- a/README_ZH_CN.md
+++ b/README_ZH_CN.md
@@ -1,0 +1,261 @@
+mGBA
+====
+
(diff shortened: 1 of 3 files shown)


endrift:
Wow, looking over your changes to the Japanese translation, it looks like there were some issues that either got introduced at some point, or the translation was just never really finished...your fixes look correct as far as I can tell.

PR #1983, opened 2020-12-12
Changes to README and review the Japanese translation.

EmpyreusX:
- Changes to README:
  - New `README_JA.md` (unfinished Japanese translation for `README.md`).
  - Updated sentences about Qt installation.
  - Standardized URLs.
  - New navigation links for README in different langs.
- Review the Japanese translation.


endrift:
I don't think I can merge this as-is. I understand why you're doing most of these changes, but I can't take them all at once:

- The Japanese README is unfinished. I'd rather not have that merged in as such.
- The changes to the README that aren't phrasing related (mostly just link updates) are fine and I can merge.
- Navigation links are a good idea but won't work when the READMEs are built into HTML files and distributed with the Windows installer, which only installs the one for your native language (if applicable)
- The phrasing changes to the README I don't necessarily like (could use some improvement) and I'd rather get the German and Spanish translators to redo those parts based on those changes than have someone who may not know it as well (I know my German is trash) attempt it.

EmpyreusX:
Well, it's up to you whatever to merge it or not, since any changes to `README.md` is indeed breaking "the status quo" to the main repo.
But changes to `mega-ja.ts` is legal and is fine to merge.

PR #1986, opened 2020-12-14
Qt: Review JPN translation

EmpyreusX:


endrift (commented):
Looks mostly good. I noticed you changed the word セーブ to 保存 in several places, which from context I believe correlates with where it's using "save" to refer to exporting some data, instead of a save in the game itself, but I'm not sure. Is this consistent with how other software translates the term?

EmpyreusX:
> Looks mostly good. I noticed you changed the word セーブ to 保存 in several places, which from context I believe correlates with where it's using "save" to refer to exporting some data, instead of a save in the game itself, but I'm not sure. Is this consistent with how other software translates the term?

セーブ is widely used in software dev scene in Japan, including but not limited to emulators (such as Famtasia and NesterJ). It is customary for them to make the distinction between these two terms.

> Why did you change load state to end with the verb but not save state? I'd imagine they should both end with the verb since it's a verb here, not a noun.

That's a slip-up in my side, and yes, both should be verb here.

PR #2060, opened 2021-03-05
CMake: Avoid caching variables and invoke git in the current directory when building libmgba only

Bonta0:
This enables including libmgba as a third-party library without polluting the main project's cache and prevents getting incorrect versioning data from git


CMakeLists.txt
```diff
@@ -8,8 +8,19 @@ if(POLICY CMP0072)
...
+
+if(NOT DEFINED LIBMGBA_ONLY)
+	get_directory_property(LIBMGBA_ONLY EXCLUDE_FROM_ALL)
+endif()
+
+if(NOT LIBMGBA_ONLY)
+	set(BINARY_NAME mgba CACHE INTERNAL "Name of output binaries")
+elseif(NOT DEFINED BINARY_NAME)
+	set(BINARY_NAME mgba)
+endif()
```
endrift, line 22: I don't know if this would work but it seems like a better solution if it does, to avoid repetition.

Why the defined check, by the way?

```suggestion
if(NOT DEFINED BINARY_NAME)
	set(BINARY_NAME mgba)
endif()
if(NOT LIBMGBA_ONLY)
	set(BINARY_NAME ${BINARY_NAME} CACHE INTERNAL "Name of output binaries")
endif()
```

CMakeLists.txt
```diff
@@ -111,8 +118,12 @@ else()
 	set(CMAKE_INSTALL_INCLUDEDIR "include")
 endif()
 
-set(LIBDIR "${CMAKE_INSTALL_LIBDIR}" CACHE PATH "Installed library directory")
-mark_as_advanced(LIBDIR)
+if(NOT LIBMGBA_ONLY)
+	set(LIBDIR "${CMAKE_INSTALL_LIBDIR}" CACHE PATH "Installed library directory")
+	mark_as_advanced(LIBDIR)
+elseif(NOT DEFINED LIBDIR)
+	set(LIBDIR "${CMAKE_INSTALL_LIBDIR}")
```
endrift, line 125: It might be better to just remove the use of `LIBDIR` entirely. I'm pretty sure it's legacy.

CMakeLists.txt
```diff
@@ -145,7 +156,7 @@ add_custom_target(version-info ALL
 	-DCONFIG_FILE=${CMAKE_CURRENT_SOURCE_DIR}/src/core/version.c.in
 	-DOUT_FILE=${CMAKE_CURRENT_BINARY_DIR}/version.c
 	-P ${CMAKE_CURRENT_SOURCE_DIR}/version.cmake
-	WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+	WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
```
endrift, line 159: This change is definitely wrong. What was the motivation here?

endrift (changes requested):
One thing is definitely wrong, but a few questionable things in here too.

CMakeLists.txt
Bonta0: Avoiding repetition is indeed better, in that case the defined check is definitely necessary or it would overwrite the cache otherwise

CMakeLists.txt
Bonta0: The goal of this PR is to remove side effects of libmgba without altering its behavior when building alone, I think it's fair to consider that to be out of scope
I could remove it of that's prefered, should that also include LIBRETRO_DIR and OE_DIR ?

CMakeLists.txt
Bonta0: That target will call git from outside of mgba's tree without this change. This is the only thing affected as the other files are referenced by their absolute paths. I have confirmed that this works as intended, unless I am missing something

CMakeLists.txt
endrift: Oh, yes, you're right.

CMakeLists.txt
endrift: You're right. I was thinking this affected the output directory.

endrift (approved):
So long as this works with a clean tree it should be fine.

PR #2064, opened 2021-03-15
add pkg-config file

jprjr:
Hi there - PR builds and installs pkg-config file for libmgba


CMakeLists.txt
```diff
@@ -981,11 +1002,23 @@ if(BUILD_EXAMPLE)
 	endif()
 endif()
 
+include(MgbaPkgConfig)
+
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include/mgba)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/core/flags.h.in ${CMAKE_CURRENT_BINARY_DIR}/include/mgba/flags.h)
+pkgcfg_configure("${CMAKE_CURRENT_SOURCE_DIR}/res/mgba.pc.in" "${CMAKE_CURRENT_BINARY_DIR}/mgba.pc"
+	NAME "MGBA"
+	DESC "MGBA as a library"
```
endrift, line 1011:
```suggestion
	NAME "${PROJECT_NAME}"
	DESC "${PROJECT_NAME} library"
```

src/platform/cmake/MgbaPkgConfig.cmake
```diff
@@ -0,0 +1,40 @@
...
+	
+	# CMake uses ; to separate values, in pkg-config we use spaces.
+	string(REPLACE ";" " " PKGCFG_CFLAGS "${PKGCFG_CFLAGS}")
+	string(REPLACE ";" " " PKGCFG_LDFLAGS_PRIV "${PKGCFG_LDFLAGS_PRIV}")
+	string(REPLACE ";" " " PKGCFG_LDFLAGS_PUB "${PKGCFG_LDFLAGS_PUB}")
+	string(REPLACE ";" " " PKGCFG_PKGS_PRIV "${PKGCFG_PKGS_PRIV}")
+	string(REPLACE ";" " " PKGCFG_PKGS_PUB "${PKGCFG_PKGS_PUB}")
+
+	set(libdir_for_pc_file "\${exec_prefix}/lib")
+	set(includedir_for_pc_file "\${prefix}/include")
```
endrift, line 36: If these are hardcoded, you should just stick it directly in the .pc.in instead.

endrift (changes requested):
The project is called mGBA, not MGBA or Mgba, and this is liable to change when medusa gets merged in. It looks like you just took a stock pkg-config cmake generator and willy-nilly stuffed it in, while you could really just use `configure_file` directly if you took a little bit of extra care with what you were doing. Most of the lines in that CMake file are comments or useless (the string replace bits can be completely removed if you set up PKGCFG_DEFINES properly when passing it in).

Adding a pkgconf config is a good idea, but this PR has a bit to go before it can be merged.

src/platform/cmake/MgbaPkgConfig.cmake
jprjr: Corrected in 8f9617d6948132830ee0dd5efc98b76ee7d7077e

endrift:
This looks fine now, though I'll need to review it a bit more closely later.

jprjr:
One thing I'm unsure about is what cflags/defines should be added to the pkg-config file (for a downstream user), vs which ones will strictly be internal flags. Right now it's not adding any to the file.

endrift:
Most of those should already be in the configured flags.h file.

CMakeLists.txt
```diff
@@ -220,6 +223,7 @@ endif()
 
 if(NOT HAIKU AND NOT MSVC AND NOT PSP2)
 	set(M_LIBRARY m)
+	set(PKGCFG_LDFLAGS_PRIV "${PKGCFG_LDFLAGS_PRIV} -lm")
```
endrift, line 226: This should probably be done by looping over `OS_LIB` at the end of the platform support block (right before the feature dependencies block)

CMakeLists.txt
```diff
@@ -345,6 +349,7 @@ if(USE_PTHREADS)
 		if(HAVE_PTHREAD AND NOT APPLE AND NOT HAIKU)
 			set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread")
 			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
+			set(PKGCFG_LDFLAGS_PRIV "${PKGCFG_LDFLAGS_PRIV} -lpthread")
```
endrift, line 352: Not 100% sure, but it's probably supposed to be this:
```suggestion
			set(PKGCFG_LDFLAGS_PRIV "${PKGCFG_LDFLAGS_PRIV} -pthread")
```

endrift (commented):
This is _probably_ fine, but two tweak suggestions. The former is most important on Windows (where very few people would care about pkg-config anyway), and the latter may or may not be necessary (I'm not sure).

CMakeLists.txt
jprjr: pkg-config `Requires.private` field is only used when linking against a static library (or more accurate, when you use the `--static` flag in your call to `pkg-config`). If pthreads isn't being used directly by the consuming app/library, then `-lpthread` should be fine to resolve missing symbols.

With dynamic libraries it's a non-issue, since the dynamic lib will already be linked against `libpthread`

endrift: I'm aware of the role of `Libs.private`, I was just worried that `-lpthread` wouldn't set all of the correct linker flags, since the objects being pulled in from the static lib haven't been linked prior.

CMakeLists.txt
```diff
@@ -996,11 +1021,19 @@ if(BUILD_EXAMPLE)
...
 endif()
 
+set(PKGCFG_NAME "${PROJECT_NAME}")
+set(PKGCFG_DESC "${PROJECT_NAME} library")
+set(PKGCFG_VERSION "${LIB_VERSION_STRING}")
+set(PKGCFG_LDFLAGS_PUB "-l${BINARY_NAME}")
+
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include/mgba)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/core/flags.h.in ${CMAKE_CURRENT_BINARY_DIR}/include/mgba/flags.h)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/res/mgba.pc.in ${CMAKE_CURRENT_BINARY_DIR}/mgba.pc.in @ONLY)
```
endrift, line 1031: Typo I didn't notice earlier:
```suggestion
configure_file(${CMAKE_CURRENT_SOURCE_DIR}/res/mgba.pc.in ${CMAKE_CURRENT_BINARY_DIR}/mgba.pc @ONLY)
```

endrift (changes requested):

CMakeLists.txt
endrift: Digging through my pkgconfig directory as-is, it seems that the libraries installed are pretty evenly divided on this one, so I'm not gonna press it.

CMakeLists.txt
endrift: Are you planning to do anything about this comment?

jprjr:
Yes, though I'm not sure when I'll be able to due to my schedule (hopefully this weekend I'll have some time). I plan to loop over the OS_LIB variable, like suggested.

On March 19, 2021 2:20:10 PM EDT, Vicki Pfau ***@***.***> wrote:
***@***.*** commented on this pull request.
>
>
>
>> @@ -220,6 +223,7 @@ endif()
> 
> if(NOT HAIKU AND NOT MSVC AND NOT PSP2)
> 	set(M_LIBRARY m)
>+	set(PKGCFG_LDFLAGS_PRIV "${PKGCFG_LDFLAGS_PRIV} -lm")
>
>Are you planning to do anything about this comment?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/mgba-emu/mgba/pull/2064#discussion_r597889096

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

jprjr:
Hey I'll just be really up-front. I'm not enjoying this. You're free to use what I've written, but I won't be contributing/interacting again. I'm sorry for wasting anybody's time.

PR #2065, opened 2021-03-16
Allow building libmgba with threads disabled

jprjr:
Unsure if this should be the default for the LIBMGBA_ONLY option, so it still defaults to wanting threads.

Also allows building a shared library with the LIBMGBA_ONLY target, by only setting it if not already defined.

The use-case I have in mind for this - I'm working on a minigsf decoder that uses mgba, and just don't need threads.


endrift:
I don't know how I feel about this. Being able to build libmgba without threads sounds okay, but it would immediately break all frontends if you tried it, so having it be an "advanced" cache option makes me a bit itchy.

That said, you should check out @kode54's downstream since I think he already has something for that.

kode54:
Yeah, my downstream features a slight modification to include a minimal build level of 3, which completely disables all rendering features, since a GSF player doesn't need those. It still has the register interact portions of the PPU, but no rendering support.

jprjr:
Would you be open to a minimal "core" option, similar to what @kode54 has? Similar to a `BUILD_LIBRETRO` or `BUILD_OPENEMU` option? Something like `BUILD_GSF` or `BUILD_MINIGSF`? It could output a separate library file as well.

endrift:
The reason I didn't want to add `MINIMAL_CORE=3` is already detailed in #1288--basically, instead of saying "turn off a bunch of components" I want it to enumerate the specific components to turn off. That said, installing a core that has no threads and no video isn't exactly appropriate for something packaged so a one-off core for e.g. `BUILD_GSF` may be a better solution.

jprjr:
Oh, I didn't see #1288 - I'll take a look through that and see if I can come up with something, thank you!

PR #2070, opened 2021-03-21
GBA: Fix wave RAM write/read behaviour

AntonioND:
Check the individual commit messages for more details.

### DIFF
--- a/src/gba/audio.c
+++ b/src/gba/audio.c
@@ -221,7 +221,27 @@ void GBAAudioWriteSOUNDBIAS(struct GBAAudio* audio, uint16_t value) {
 }
 
 void GBAAudioWriteWaveRAM(struct GBAAudio* audio, int address, uint32_t value) {
-	audio->psg.ch3.wavedata32[address | (!audio->psg.ch3.bank * 4)] = value;
+	int bank = !audio->psg.ch3.bank;
+
+	// When the audio hardware is turned off, it acts like bank 0 has been
+	// selected in SOUND3CNT_L, so any read comes from bank 1.
+	if (!audio->enable) {
+		bank = 1;
+	}
+
+	audio->psg.ch3.wavedata32[address | (bank * 4)] = value;
+}
+
+uint32_t GBAAudioReadWaveRAM(struct GBAAudio* audio, int address) {
+	int bank = !audio->psg.ch3.bank;
+
+	// When the audio hardware is turned off, it acts like bank 0 has been
+	// selected in SOUND3CNT_L, so any read comes from bank 1.
+	if (!audio->enable) {
+		bank = 1;
+	}
+
+	return audio->psg.ch3.wavedata32[address | (bank * 4)];
 }
 
 uint32_t GBAAudioWriteFIFO(struct GBAAudio* audio, int address, uint32_t value) {
--- a/src/gba/io.c
+++ b/src/gba/io.c
@@ -342,248 +342,254 @@ void GBAIOInit(struct GBA* gba) {
 
 void GBAIOWrite(struct GBA* gba, uint32_t address, uint16_t value) {
 	if (address < REG_SOUND1CNT_LO && (address > REG_VCOUNT || address < REG_DISPSTAT)) {
-		value = gba->video.renderer->writeVideoRegister(gba->video.renderer, address, value);
-	} else {
-		switch (address) {
-		// Video
-		case REG_DISPSTAT:
-			value &= 0xFFF8;
-			GBAVideoWriteDISPSTAT(&gba->video, value);
-			return;
+		gba->memory.io[address >> 1] = gba->video.renderer->writeVideoRegister(gba->video.renderer, address, value);
+		return;
+	}
 
-		case REG_VCOUNT:
-			mLOG(GBA_IO, GAME_ERROR, "Write to read-only I/O register: %03X", address);
-			return;
+	if (address >= REG_SOUND1CNT_LO && address <= REG_SOUNDCNT_LO && !gba->audio.enable) {
+		// Ignore writes to most audio registers if the hardware is off.
+		return;
+	}
 
-		// Audio
-		case REG_SOUND1CNT_LO:
-			GBAAudioWriteSOUND1CNT_LO(&gba->audio, value);
-			value &= 0x007F;
-			break;
-		case REG_SOUND1CNT_HI:
-			GBAAudioWriteSOUND1CNT_HI(&gba->audio, value);
-			break;
-		case REG_SOUND1CNT_X:
-			GBAAudioWriteSOUND1CNT_X(&gba->audio, value);
-			value &= 0x47FF;
-			break;
-		case REG_SOUND2CNT_LO:
-			GBAAudioWriteSOUND2CNT_LO(&gba->audio, value);
-			break;
-		case REG_SOUND2CNT_HI:
-			GBAAudioWriteSOUND2CNT_HI(&gba->audio, value);
-			value &= 0x47FF;
-			break;
-		case REG_SOUND3CNT_LO:
-			GBAAudioWriteSOUND3CNT_LO(&gba->audio, value);
-			value &= 0x00E0;
-			break;
-		case REG_SOUND3CNT_HI:
-			GBAAudioWriteSOUND3CNT_HI(&gba->audio, value);
-			value &= 0xE03F;
-			break;
-		case REG_SOUND3CNT_X:
-			GBAAudioWriteSOUND3CNT_X(&gba->audio, value);
-			// TODO: The low bits need to not be readable, but still 8-bit writable
-			value &= 0x47FF;
-			break;
-		case REG_SOUND4CNT_LO:
-			GBAAudioWriteSOUND4CNT_LO(&gba->audio, value);
-			value &= 0xFF3F;
-			break;
-		case REG_SOUND4CNT_HI:
-			GBAAudioWriteSOUND4CNT_HI(&gba->audio, value);
-			value &= 0x40FF;
-			break;
-		case REG_SOUNDCNT_LO:
-			GBAAudioWriteSOUNDCNT_LO(&gba->audio, value);
-			value &= 0xFF77;
-			break;
-		case REG_SOUNDCNT_HI:
-			GBAAudioWriteSOUNDCNT_HI(&gba->audio, value);
-			value &= 0x770F;
-			break;
-		case REG_SOUNDCNT_X:
-			GBAAudioWriteSOUNDCNT_X(&gba->audio, value);
-			value &= 0x0080;
-			value |= gba->memory.io[REG_SOUNDCNT_X >> 1] & 0xF;
-			break;
-		case REG_SOUNDBIAS:
-			GBAAudioWriteSOUNDBIAS(&gba->audio, value);
-			break;
+	switch (address) {
+	// Video
+	case REG_DISPSTAT:
+		value &= 0xFFF8;
+		GBAVideoWriteDISPSTAT(&gba->video, value);
+		return;
 
-		case REG_WAVE_RAM0_LO:
-		case REG_WAVE_RAM1_LO:
-		case REG_WAVE_RAM2_LO:
-		case REG_WAVE_RAM3_LO:
-			GBAIOWrite32(gba, address, (gba->memory.io[(address >> 1) + 1] << 16) | value);
-			break;
+	case REG_VCOUNT:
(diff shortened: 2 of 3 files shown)


AntonioND:
I have created this test ROM to make sure that my changes are correct. It passes in no$gba, but fails in mgba. It almost passes on mgba after the changes in this PR, but it looks like the default waveform is still incorrect in one case in mgba.

[wave_ram_test.zip](https://github.com/mgba-emu/mgba/files/6178087/wave_ram_test.zip)

A successful run of this test should just print the tests as they run (no "Failed" anywhere).

src/gba/audio.c
```diff
@@ -161,6 +161,9 @@ void GBAAudioWriteSOUND2CNT_HI(struct GBAAudio* audio, uint16_t value) {
 }
 
 void GBAAudioWriteSOUND3CNT_LO(struct GBAAudio* audio, uint16_t value) {
+	if (!audio->enable)
+		return;
```
endrift, line 165:
```suggestion
	if (!audio->enable) {
		return;
	}
```

src/gba/audio.c
```diff
@@ -224,6 +227,16 @@ void GBAAudioWriteWaveRAM(struct GBAAudio* audio, int address, uint32_t value) {
 	audio->psg.ch3.wavedata32[address | (!audio->psg.ch3.bank * 4)] = value;
 }
 
+uint32_t GBAAudioReadWaveRAM(struct GBAAudio* audio, int address) {
+	int bank = !audio->psg.ch3.bank;
+
+	// When the audio hardware is turned off, bank 0 is selected.
+	if (!audio->enable)
+		bank = 1;
```
endrift, line 235:
```suggestion
	if (!audio->enable) {
		bank = 1;
	}
```

endrift: Further, it the comment can be a bit confusing given the fact that it says "0" but assigns 1. If you put the negation after the block it'll probably be clearer.

endrift (changes requested):
Some style nits mostly

src/gba/audio.c
AntonioND: Done.

src/gba/audio.c
AntonioND: I see what you mean. I've reworded the comment a bit, what do you think now?

Also, I forgot to commit the same check in the function to write to wave RAM, I've added the change to the commit now.

endrift (approved):

endrift:
I want to test this on some games that make heavy use of channel 3 before I merge it but otherwise it should be fine.

src/gba/audio.c
```diff
@@ -161,6 +161,10 @@ void GBAAudioWriteSOUND2CNT_HI(struct GBAAudio* audio, uint16_t value) {
 }
 
 void GBAAudioWriteSOUND3CNT_LO(struct GBAAudio* audio, uint16_t value) {
+	if (!audio->enable) {
+		return;
+	}
```
endrift, line 166: Actually, I realize that this check needs to be done _outside_ of the function since stuff gets written into the I/O memory buffer outside of this function. That probably is similar for all of the other audio writes...there's actually a TODO about this in GBIORead.

AntonioND: Well, according to GBATEK:

> While Bit 7 is cleared, both PSG and FIFO sounds are disabled, and all PSG registers at 4000060h..4000081h are reset to zero (and must be re-initialized after re-enabling sound). However, registers 4000082h and 4000088h are kept read/write-able (of which, 4000082h has no function when sound is off, whilst 4000088h does work even when sound is off).

endrift: It should be done as a separate commit at least, but same PR might be fine.

src/gba/io.c
```diff
@@ -343,247 +343,256 @@ void GBAIOInit(struct GBA* gba) {
...
-		case REG_VCOUNT:
-			mLOG(GBA_IO, GAME_ERROR, "Write to read-only I/O register: %03X", address);
+	if (address >= REG_SOUND1CNT_LO && address <= REG_SOUNDCNT_LO) {
+		if (!gba->audio.enable) {
+			// Ignore writes to most audio registers if the hardware is off.
 			return;
+		}
+	}
 
-		// Audio
```
AntonioND, line 358: I've had to modify the flow of this function to avoid having checks in each case of each sound register. If you check this branch locally with gitk and tick the "ignore whitespace changes" you can see that the actual changes are minimal.

AntonioND:
This is another test ROM that also checks the initial state of the wave RAM:

[wave_ram.zip](https://github.com/mgba-emu/mgba/files/6186335/wave_ram.zip)

Unfortunately, it doesn't use libgba, it uses my library (https://github.com/AntonioND/ugba), so I've included source, ROM, and reference image. After all, I've detected this problem in mGBA because I was developing tests for my library and using mGBA for quick testing.

src/gba/io.c
```diff
@@ -343,247 +343,256 @@ void GBAIOInit(struct GBA* gba) {
...
-			GBAVideoWriteDISPSTAT(&gba->video, value);
-			return;
+		gba->memory.io[address >> 1] = value;
+		return;
+	}
 
-		case REG_VCOUNT:
-			mLOG(GBA_IO, GAME_ERROR, "Write to read-only I/O register: %03X", address);
+	if (address >= REG_SOUND1CNT_LO && address <= REG_SOUNDCNT_LO) {
+		if (!gba->audio.enable) {
```
endrift, line 351: Merge these two conditionals.

src/gba/io.c
```diff
@@ -343,247 +343,256 @@ void GBAIOInit(struct GBA* gba) {
...
 	if (address < REG_SOUND1CNT_LO && (address > REG_VCOUNT || address < REG_DISPSTAT)) {
 		value = gba->video.renderer->writeVideoRegister(gba->video.renderer, address, value);
-	} else {
-		switch (address) {
-		// Video
-		case REG_DISPSTAT:
-			value &= 0xFFF8;
-			GBAVideoWriteDISPSTAT(&gba->video, value);
-			return;
+		gba->memory.io[address >> 1] = value;
```
endrift, line 346: Merge these two lines.

endrift (changes requested):
I probably wouldn't have changed the control flow that way, but it's not actually bad to do it that way. That said, there are some extra bits that should be merged in the process.

src/gba/audio.c
```diff
@@ -57,6 +57,7 @@ void GBAAudioInit(struct GBAAudio* audio, size_t samples) {
 
 void GBAAudioReset(struct GBAAudio* audio) {
 	GBAudioReset(&audio->psg);
+	memset(audio->psg.ch3.wavedata8, 0, sizeof(audio->psg.ch3.wavedata8));
```
endrift, line 60: This should probably be moved to `GBAudioReset` and the pattern initialization be moved to `GBSkipBIOS` then.

endrift (changes requested):

src/gb/gb.c
```diff
@@ -565,6 +566,11 @@ void GBSkipBIOS(struct GB* gb) {
 			GBVideoDisableCGB(&gb->video);
 		}
 		nextDiv = 0xC;
+		size_t i;
```
endrift, line 569: I'm not sure this will compile everywhere since you're putting a variable declaration in the middle of a `switch` statement. That makes some compilers very unhappy.

endrift (approved):

src/gb/gb.c
AntonioND: I can put it inside the for, but I saw that this was the style of this file.

AntonioND: Also, I'm not sure about whether to initialize the wave RAM to the pattern always, or just in GCB and AGB.

endrift: No, that's worse. You might need to declare it _outside_ of the switch, or put a block inside the case.

As for CGB and AGB vs always, I'd err towards always since that's the old behavior.

AntonioND: In that case, I can just leave the loop outside of the switch for now.

src/gb/gb.c
```diff
@@ -568,6 +568,13 @@ void GBSkipBIOS(struct GB* gb) {
 		break;
 	}
 
+	// TODO: Emulate randomness of wave RAM in DMG. This is only valid in CGB and AGB.
```
AntonioND, line 571: Actually, this is wrong, this means that when the emulator starts in DMG mode with the boot ROM enabled the wave RAM will be 0 instead of having the pattern.

endrift: What is the default pattern on the DMG?

AntonioND: https://github.com/LIJI32/SameBoy/blob/72cb391612863a43cb3c876b1d210534a54a4020/Core/gb.c#L1286-L1315

Random, as the TODO comment said.

endrift: Is that a problem? You're right that it's different from the old behavior though.

src/gba/io.c
```diff
@@ -342,248 +342,254 @@ void GBAIOInit(struct GBA* gba) {
...
-			GBAVideoWriteDISPSTAT(&gba->video, value);
-			return;
+		gba->memory.io[address >> 1] = gba->video.renderer->writeVideoRegister(gba->video.renderer, address, value);
+		return;
+	}
 
-		case REG_VCOUNT:
-			mLOG(GBA_IO, GAME_ERROR, "Write to read-only I/O register: %03X", address);
-			return;
+	if ((address >= REG_SOUND1CNT_LO && address <= REG_SOUNDCNT_LO) && (!gba->audio.enable)) {
```
endrift, line 349: These extra parens aren't needed.

src/gb/gb.c
AntonioND: So the right behaviour would be that ``GBReset()`` always initializes the wave RAM to random values, and ``GBSkipBIOS()`` initializes the GBC/GBA values. However, that changes the behaviour, so I think that in this patch I should just add this loop to ``GBReset()`` unconditionally.

endrift: Doesn't that just revert what you'd done prior?

AntonioND: Well, it's a bit of a revert, but the idea is that I don't want to deal with the DMG randomness of the wave RAM, that's a topic that I don't understand well enough. I guess that the result I'm looking for right now is that the wave RAM has the right values (zero) in the case of GBA in GBA mode, and it behaves the same way as it did before the PR for all other cases.

endrift: I suppose that's fair for right now. I'd like to merge this before the release since it makes subtle API changes, so you have like two days to finish the PR :P

AntonioND: 😅

AntonioND: Ok, it should be fine now.

endrift: Honestly at this point the only perceptible difference from the third commit is that...now the memory is always zeroed on AGB (excluding CGB-on-AGB), and there's a more comprehensive comment...but it's in the wrong file now. I'm not sure this helps at all heh.

AntonioND: Well, the initial state of the wave RAM has some effect in at least the game R-Type: https://gbdev.gg8.se/wiki/articles/Gameboy_sound_hardware#Power_Control I don't know if there is any GBA game affected by this, though. However, I was testing my code to use the PSG channel 3 and noticed it worked on mGBA but not on hardware, so it's not a stretch to assume that someone else may find the same problem. 

Considering that most people would just assume that it behaves the same way in GBC mode or GBA mode, initializing it to 0 in the GBA is something worth doing. It is a significant difference from the random or square waves of GB mode that, at least, would be heard.

Regarding having it in the wrong files, it depends on how you look at it. The way I see it, it is the responsibility of ``GBSkipBIOS()`` to initialize the pattern in GBC and GBA (GBC mode), not of ``GBAudioReset()`` because it's shared between all GB and GBA modes.

As I said in a previous comment, the right behaviour would be that ``GBReset()`` always initializes the wave RAM to random values (because that's probably what happens, at least on DMG, maybe it changed in GBC but it doesn't make any difference), and ``GBSkipBIOS()`` initializes the GBC/GBA values. However, that would require to randomize the wave RAM in a realistic way.

AntonioND: However, if this is going to be that controversial, I can get rid of the third patch of this PR and push it as a different PR.

endrift: I think that would be a good idea. Plus that one doesn't change API so it can be merged after release if needed.

AntonioND: Ok, removed.

PR #2139, opened 2021-04-22
GBA: Fix #2134 without reverting 5c58186f03fb053038ee804b44051f21623e…

### LINKED ISSUE #2134: Metroid: Zero Mission - Some GSF files play burst of noise at boot
From the GSF set, which produce a self-booting ROM that does nothing but play music or audio, some of them produce a burst of static noise on boot, just before the intended sound output.

From the following, 98 - Save Complete.gsf, as converted to a stripped .GBA file:

[metroid_zero_mission_gsf.zip](https://github.com/mgba-emu/mgba/files/6346465/metroid_zero_mission_gsf.zip)

If I already posted this, I have no idea how to find it, other than somehow searching for my own issues.

kode54:
Increasing DMA fetch count to run a DMA transfer even if FIFO is only half empty fixes Metroid Zero Mission GSFs, but an extra bit of caution was added, implementing a separate FIFO filled counter that never increments past the FIFO size, and changing all the fifoSize calculations to use this running counter.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,10 @@
 0.10.0: (Future)
 Features:
  - Tool for converting scanned pictures of e-Reader cards to raw dotcode data
+Misc:
+ - Qt: Rearrange menus some
+
+0.9.1: (2021-04-18)
 Emulation fixes:
  - ARM: Fix LDM^ with empty rlist (fixes mgba.io/i/2127)
  - Core: Fix first event scheduling after loading savestate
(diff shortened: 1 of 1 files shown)


kode54:
Okay, this doesn't break the ABI of the structures, but does alter the meaning of an existing field. Serialization or deserialization should remain the same.

endrift:
I'm going to need @gericom to review this, but this eliminates the boundary condition of the read and write pointers matching causing issues, which I seem to recall hearing was, in fact, an issue. Also has savestate issues due to the fact that those fields only store 0-7 internally and cannot store 8.

kode54:
Yeah, and it currently expands the write pointer to the same value as the filled count. Hmm, you're right, this will need review.

PR #2145, opened 2021-04-26
GBA: Add override for DigiCommunication Nyo

sdhizumi:
I got a physical copy of the game, dumped the save data, and tried to get it to work with mGBA, But this game was not detected correctly in autodetect.
It uses EEPROM, seems like the same PCB as the Dragon Ball Z - The Legacy of Goku.


endrift:
This game looks like it's supposed to work with SRAM or EEPROM, but the SRAM code is broken so it only works with EEPROM. Good catch.

Issue #2150, opened 2021-04-28
Feature request: Fast forward mute on other platforms


rik-smeets:
I use mGBA as an emulator on the Wii. The function to mute audio whilst fast forwarding like in the Windows desktop app isn't available there (unless I missed it somehow). So when activating fast forward, I manually mute and unmute the audio.

Even though I'm unexperienced with coding in C, I did manage to build a working version which includes this feature. It's probably not optimal, but if you like I could open a pull request? It'd be open for review/changes of course.

RetroEdit:
endrift is happy to get pull requests, so feel free to send one. Worst that happens, it won't get accepted. Most likely, there will be a feedback and review process.

PR #2151, opened 2021-04-28
mGUI: Add fast forward mute

rik-smeets:
Adds the fast forward mute feature as requested in #2150. 

Please note I'm unexperienced coding in C; I did try to adhere to the coding style / standards.
I'd love to hear feedback on this. You can also make edits yourself if you prefer. Thanks!

### DIFF
--- a/src/feature/gui/gui-runner.c
+++ b/src/feature/gui/gui-runner.c
@@ -455,6 +455,12 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
 	int drawFps = false;
 	mCoreConfigGetIntValue(&runner->config, "fpsCounter", &drawFps);
 
+	int mute = false;
+	mCoreConfigGetIntValue(&runner->config, "mute", &mute);
+
+	int fastForwardMute = false;
+	mCoreConfigGetIntValue(&runner->config, "fastForwardMute", &fastForwardMute);
+
 	bool running = true;
 
 #ifndef DISABLE_THREADING
@@ -506,20 +512,32 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
 			if (guiKeys & (1 << mGUI_INPUT_SCREENSHOT)) {
 				mCoreTakeScreenshot(runner->core);
 			}
+			bool muteTogglePressed = guiKeys & (1 << mGUI_INPUT_MUTE_TOGGLE);
+			if (muteTogglePressed) {
+				mute = !mute;
+				mCoreConfigSetUIntValue(&runner->config, "mute", mute);
+				runner->core->reloadConfigOption(runner->core, "mute", &runner->config);
+			}
 			if (runner->setFrameLimiter) {
 				if (guiKeys & (1 << mGUI_INPUT_FAST_FORWARD_TOGGLE)) {
 					fastForward = !fastForward;
 				}
-				if (fastForward || (heldKeys & (1 << mGUI_INPUT_FAST_FORWARD_HELD))) {
+				bool fastForwarding = fastForward || (heldKeys & (1 << mGUI_INPUT_FAST_FORWARD_HELD));
+				if (fastForwarding) {
+					if (fastForwardMute && !mute && !muteTogglePressed) {
+						mCoreConfigSetUIntValue(&runner->core->config, "mute", fastForwardMute);
+						runner->core->reloadConfigOption(runner->core, "mute", NULL);
+					}
+
 					runner->setFrameLimiter(runner, false);
 				} else {
 					runner->setFrameLimiter(runner, true);
+
+					if (fastForwardMute && !mute && !muteTogglePressed) {
+						mCoreConfigSetUIntValue(&runner->core->config, "mute", !fastForwardMute);
+						runner->core->reloadConfigOption(runner->core, "mute", NULL);
+					}
 				}
-			}
-			if (guiKeys & (1 << mGUI_INPUT_MUTE_TOGGLE)) {
-				int mute = !runner->core->opts.mute;
-				mCoreConfigSetUIntValue(&runner->config, "mute", mute);
-				runner->core->reloadConfigOption(runner->core, "mute", &runner->config);
 			}
 			uint16_t keys = runner->pollGameInput(runner);
 			if (runner->prepareForFrame) {
@@ -651,6 +669,8 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
 		}
 		mCoreConfigGetIntValue(&runner->config, "fpsCounter", &drawFps);
 		mCoreConfigGetIntValue(&runner->config, "showOSD", &showOSD);
+		mCoreConfigGetIntValue(&runner->config, "mute", &mute);
+		mCoreConfigGetIntValue(&runner->config, "fastForwardMute", &fastForwardMute);
 #ifdef M_CORE_GB
 		if (runner->core->platform(runner->core) == mPLATFORM_GB) {
 			runner->core->reloadConfigOption(runner->core, "gb.pal", &runner->config);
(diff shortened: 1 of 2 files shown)


endrift:
The style is fine, unfortunately the implementation is not quite correct. The core is not aware of fast forward at all, as that's the job of the runner. Instead of modifying the core/ directory at all you should use the `fastForwardMute` setting to override the `mute` setting in the core's config, which is distinct from the runner's config. See here for how you can override it directly in the core's config:

https://github.com/mgba-emu/mgba/blob/9c7f15ad581e80a9649371f3aa6385cf1c634e3f/src/feature/gui/gui-runner.c#L519-L523

But since this would _not_ be saved in the runner's config you'd use `runner->core->config` instead of `runner->config`. You should also pass `NULL` as the last argument to `reloadConfigOption` as that informs the core to use its internal config as opposed to an externally provided one.

So it'd be similar to this:

```C
 if (fastForward && !mute) { 
 	mCoreConfigSetUIntValue(&runner->core->config, "mute", fastForwardMute); 
 	runner->core->reloadConfigOption(runner->core, "mute", NULL); 
 } 
```

But with the lookup in the runner->config for the value of fastForwardMute. Note the difference in the conditional: you probably don't want it _unmuting_ if you have it muted, and you probably want it to _unmute_ after you stop fast-forwarding (though this should only be executed when fast-forwarding is toggled).

rik-smeets:
Thanks for the feedback and explanation @endrift! My understanding of the code has certainly improved because of it. Thanks!

I amended my previous commit and I hope this is the implementation you are looking for. If not, please feel free to leave some more feedback.

include/mgba/core/config.h
```diff
@@ -55,6 +55,7 @@ struct mCoreOptions {
 
 	int volume;
 	bool mute;
+	bool fastForwardMute;
```
endrift, line 58: This still shouldn't be here.

src/core/config.c
```diff
@@ -408,6 +408,9 @@ void mCoreConfigMap(const struct mCoreConfig* config, struct mCoreOptions* opts)
 	if (_lookupIntValue(config, "mute", &fakeBool)) {
 		opts->mute = fakeBool;
 	}
+	if (_lookupIntValue(config, "fastForwardMute", &fakeBool)) {
+		opts->fastForwardMute = fakeBool;
+	}
```
endrift, line 413: Ditto

src/core/config.c
```diff
@@ -445,6 +448,7 @@ void mCoreConfigLoadDefaults(struct mCoreConfig* config, const struct mCoreOptio
 	ConfigurationSetIntValue(&config->defaultsTable, 0, "height", opts->height);
 	ConfigurationSetIntValue(&config->defaultsTable, 0, "volume", opts->volume);
 	ConfigurationSetIntValue(&config->defaultsTable, 0, "mute", opts->mute);
+	ConfigurationSetIntValue(&config->defaultsTable, 0, "fastForwardMute", opts->fastForwardMute);
```
endrift, line 451: Ditto.

endrift (changes requested):
Looking better, but there's still some stuff that needs to be removed from the previous version.

src/feature/gui/gui-runner.c
```diff
@@ -506,21 +512,33 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
 			if (guiKeys & (1 << mGUI_INPUT_SCREENSHOT)) {
 				mCoreTakeScreenshot(runner->core);
 			}
+			bool muteTogglePressed = guiKeys & (1 << mGUI_INPUT_MUTE_TOGGLE);
+			if (muteTogglePressed) {
+				mute = !runner->core->opts.mute;
```
endrift, line 517: You should probably be either storing the old value of mute outside of the core or be pulling it directly from the runner config since the core's concept can get desynchronized, e.g. if mute is off but fast-forward mute is on, then you fast-forward and toggle mute it'll get confused. Since you're already polling that whenever it could change you can probably just simplify this as `mute = !mute`, but I'm not 100% sure about other contexts. I would recommend testing toggling mute both while fast-forwarding and while not fast-forwarding with fastForwardMute enabled and not enabled to make sure the correct behavior is always observed.

Putting the entire block of logic for how to set mute currently in one place after the frame limiter stuff will probably simplify the code for that a bit too.

src/feature/gui/gui-runner.c
```diff
@@ -540,12 +558,13 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
...
 						unsigned w;
-						if (fastForward || (heldKeys & (1 << mGUI_INPUT_FAST_FORWARD_HELD))) {
+						bool fastForwarding = fastForward || (heldKeys & (1 << mGUI_INPUT_FAST_FORWARD_HELD));
+						if (fastForwarding) {
 							GUIFontDrawIcon(runner->params.font, origin, GUIFontHeight(runner->params.font) / 2, GUI_ALIGN_RIGHT, 0, 0x7FFFFFFF, GUI_ICON_STATUS_FAST_FORWARD);
 							GUIFontIconMetrics(runner->params.font, GUI_ICON_STATUS_FAST_FORWARD, &w, NULL);
 							origin -= w + GUIFontHeight(runner->params.font) / 2;
 						}
-						if (runner->core->opts.mute) {
+						if (runner->core->opts.mute || (fastForwarding && fastForwardMute)) {
```
endrift, line 567: `runner->core->opts.mute` is the core's conception of if it's muted (as mentioned above) so the added clause should be redundant here.

endrift (changes requested):
Took a closer look and it's _almost_ correct, but I'm a bit worried about the value of `mute` becoming wrong if you toggle it while fast forwarding.

src/feature/gui/gui-runner.c
rik-smeets: You're right, I removed the clause. Thanks!

src/feature/gui/gui-runner.c
rik-smeets: I already tested it before, but I did change this to `mute = !mute` just to be sure and tested it again. I tested all the scenarios you mentioned a couple of times and it worked as expected.

endrift:
I'll double check this tomorrow (by testing on hardware) to verify and then merge.

src/feature/gui/gui-runner.c
```diff
@@ -540,7 +558,8 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
 					if (showOSD) {
 						unsigned origin = runner->params.width - GUIFontHeight(runner->params.font) / 2;
 						unsigned w;
-						if (fastForward || (heldKeys & (1 << mGUI_INPUT_FAST_FORWARD_HELD))) {
+						bool fastForwarding = fastForward || (heldKeys & (1 << mGUI_INPUT_FAST_FORWARD_HELD));
```
endrift, line 561: This is fine to merge as-is, but I'm curious if there was any point to this change or if it just got confused as it got partially backed out?

endrift (approved):

src/feature/gui/gui-runner.c
rik-smeets: I believed this change was made because the variable was used in more places in earlier revisions of this pull request. As of now, it doesn't really add anything. Perhaps it's a bit easier to read like this, but that's a personal preference.
If you prefer this changed as well, let me know.

endrift: It is possibly easier to read but it's also changing code without a functional difference in a PR that has functional changes, which I prefer to avoid. It makes reading git blame more complicated, for example. So if you could revert this one specific part I'll go ahead and merge it.

rik-smeets: Alright, I reverted this specific part!

PR #2181, opened 2021-05-14
Core: make map cache aware of how many tiles to update in a write

ElementW:
Fixes a bug where ~~the memory editor didn't update renderer state when writing to VRAM, and another where~~ (chunked out to #2544) the map viewer only updates half the tiles when doing (16-bit) VRAM writes to an affine background (where tiles are 8 bit = 2 tiles updated at once).

### DIFF
--- a/include/mgba/core/map-cache.h
+++ b/include/mgba/core/map-cache.h
@@ -23,6 +23,7 @@ DECL_BITS(mMapCacheSystemInfo, TilesWide, 8, 4);
 DECL_BITS(mMapCacheSystemInfo, TilesHigh, 12, 4);
 DECL_BITS(mMapCacheSystemInfo, MacroTileSize, 16, 7);
 DECL_BITS(mMapCacheSystemInfo, MapAlign, 23, 2);
+DECL_BIT(mMapCacheSystemInfo, TilesPerWrite, 25);
 
 DECL_BITFIELD(mMapCacheEntryFlags, uint16_t);
 DECL_BITS(mMapCacheEntryFlags, PaletteId, 0, 4);
--- a/src/core/map-cache.c
+++ b/src/core/map-cache.c
@@ -75,6 +75,12 @@ void mMapCacheWriteVRAM(struct mMapCache* cache, uint32_t address) {
 		++status->vramVersion;
 		status->flags = mMapCacheEntryFlagsClearVramClean(status->flags);
 		status->tileStatus[mMapCacheEntryFlagsGetPaletteId(status->flags)].vramClean = 0;
+		if (mMapCacheSystemInfoGetTilesPerWrite(cache->sysConfig) == 1) {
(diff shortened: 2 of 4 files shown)


src/gba/extra/proxy.c
```diff
@@ -282,7 +282,8 @@ void GBAVideoProxyRendererWriteVRAM(struct GBAVideoRenderer* renderer, uint32_t
 		proxyRenderer->backend->writeVRAM(proxyRenderer->backend, address);
 	}
 	if (renderer->cache) {
-		mCacheSetWriteVRAM(renderer->cache, address);
+		mCacheSetWriteVRAM(renderer->cache, address & ~1);
+		mCacheSetWriteVRAM(renderer->cache, address | 1);
```
endrift, line 286: This is not the correct fix for this; for one, this function is called exactly the same way in multiple places, not just this function. For another, the API specifies what the width of some of these writes is supposed correlate with, but I'm not sure it's consistent so it might need to be reviewed in more depth.

src/gba/memory.c
```diff
@@ -1296,9 +1296,11 @@ void GBAPatch16(struct ARMCore* cpu, uint32_t address, int16_t value, int16_t* o
 		if ((address & 0x0001FFFF) < SIZE_VRAM) {
 			LOAD_16(oldValue, address & 0x0001FFFE, gba->video.vram);
 			STORE_16(value, address & 0x0001FFFE, gba->video.vram);
+			gba->video.renderer->writeVRAM(gba->video.renderer, address & 0x0001FFFE);
 		} else {
 			LOAD_16(oldValue, address & 0x00017FFE, gba->video.vram);
 			STORE_16(value, address & 0x00017FFE, gba->video.vram);
+			gba->video.renderer->writeVRAM(gba->video.renderer, address & 0x00017FFE);
```
endrift, line 1303: This fix is not complete; you need to apply this same logic to `GBAPatch32` too.

endrift (changes requested):
These two patches are a bit underbaked for both bugs in question.

src/gba/extra/proxy.c
ElementW: Solved by adding a map cache sysconfig bitfield member specifying whether a write updates only one, or two tiles. Please tell me if this approach is acceptable.

endrift: I'm still not sure how I feel about this approach. The other fix looks fine now so maybe split it into two PRs so I can merge one first and then bikeshed the other one later.

ElementW:
> The other fix looks fine now so maybe split it into two PRs so I can merge one first and then bikeshed the other one later.

@endrift I split out this PR into two. See #2544 for the renderer state VRAM update fix part.

endrift:
I fixed this in a slightly different way in 2969a8bf7a25e00616436b43c7a5e11255f20070

PR #2194, opened 2021-05-22
Updated translation

Felipefpl:
Updated brazilian portuguese translation.

### DIFF
--- a/src/platform/qt/ts/mgba-pt_BR.ts
+++ b/src/platform/qt/ts/mgba-pt_BR.ts
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.1" language="pt-BR" sourcelanguage="en">
+<TS version="2.1" language="pt_BR" sourcelanguage="en">
 <context>
     <name>AboutScreen</name>
     <message>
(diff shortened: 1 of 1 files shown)


src/platform/qt/ts/mgba-pt_BR.ts
```diff
@@ -117,27 +117,27 @@ Game Boy Advance é uma marca registrada da Nintendo Co., Ltd.</translation>
...
     <message>
         <location filename="../BattleChipView.ui" line="155"/>
         <source>Progress &amp;Gate</source>
-        <translation>Progress &amp;Gate</translation>
+        <translation type="unfinished">Portal do &amp;Progresso</translation>
     </message>
     <message>
         <location filename="../BattleChipView.ui" line="162"/>
         <source>Beast &amp;Link Gate</source>
-        <translation>Beast &amp;Link Gate</translation>
```
endrift, line 140: These are specific product names and may not make sense to translate.

endrift (commented):
Was it intentional to flip a bunch of these to unfinished status?

src/platform/qt/ts/mgba-pt_BR.ts
Felipefpl: yes, after all it's between >   and   <. ;)

endrift: I think you've misunderstood me, and you also didn't answer my other question.

Felipefpl: nope, i didnt change the status, i translated the file using qtlinguist, any changes about statuses made in the file was made by the program, not me.

endrift: Well if it wasn't intentional then you should tell Qt Linguist to change them back. The program doesn't have a mind of its own.

Felipefpl: the problem is: qtlinguist doesnt have a web site, at least there is no link in softpedia where i download it.

endrift: You moved several strings from finished to unfinished, and then told me it wasn't intentional, and when I asked you to flip them back, you gave not much of an explanation of why you couldn't change them back. I cannot accept it if you don't change it back.

Felipefpl: i'll change them but what damage this causes to mgba?

Felipefpl (commented):
Removed status unfinished.

src/platform/qt/ts/mgba-pt_BR.ts
endrift: The specific problem is that the new translation page (https://hosted.weblate.org/projects/mgba/mgba-qt/, you can use it in the future instead of a pull request) marks "unfinished" translations as "untranslated", which makes things very confusing.

Felipefpl: oh i see, thanks for the info, didnt know about the page.

PR #2258, opened 2021-07-26
Qt: Game Boy Printer output via external command

amber-sixel:
This gives the Qt interface an option to send Game Boy Printer image data to an external program. It saves the image in PNG format to a temp file. When the program exits, it signals to the Game Boy that printing has finished, and deletes the temp file (though I haven't actually verified that that last bit is happening, yet).

There's no UI yet. To use this, you have to open `config.ini` and add an `extPrintCmd` line to the `[ports.qt]` section. It takes a full command line, as parsed by `QProcess::splitCommand()`. Put a `%1` anywhere you want the temp filename to go.

I didn't go into this with any familiarity whatsoever with C++, Qt, or mGBA internals, so there may well be huge, glaring mistakes here.

### DIFF
--- a/include/mgba/internal/gb/sio/printer.h
+++ b/include/mgba/internal/gb/sio/printer.h
@@ -29,6 +29,13 @@ enum GBPrinterPacketByte {
 	GB_PRINTER_BYTE_UNCOMPRESSED_DATA
 };
 
+enum GBPrinterParameterByte {
+	GB_PRINTER_PARAMETER_BYTE_SHEETS,
+	GB_PRINTER_PARAMETER_BYTE_MARGINS,
+	GB_PRINTER_PARAMETER_BYTE_PALETTE,
+	GB_PRINTER_PARAMETER_BYTE_EXPOSURE
+};
+
 enum GBPrinterStatus {
 	GB_PRINTER_STATUS_CHECKSUM_ERROR = 0x01,
 	GB_PRINTER_STATUS_PRINTING = 0x02,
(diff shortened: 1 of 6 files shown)


amber-sixel:
Parameters to the print command sent to the emulated printer can now be used. `%2` is the exposure setting, as a number from 0 to 127. `%3` and `%4` are the number of blank pixel lines to print before and after the image, respectively.

I've also modified the GUI printer dialog to use margins, and lighten rendered images according to the exposure setting.

I'm not sure why the integration test failed. As far as I can tell, it doesn't look related to anything I did.

endrift:
In general this isn't the sort of thing I'd really want implemented in the frontend since it's such an edge use case of an edge use case. I'd prefer it to be done via the scripting interface, but that's not done yet.

That said, there are some glaring issues I can point to, probably mostly stemming due to unfamiliarity with Qt and C++. I can point them out but I'm unsure what I'd want this to look like to be mergeable.

amber-sixel:
Yeah, that makes sense. Thank you for the feedback. I'll go ahead and close this, then.

Some questions, if I may:

- Would it be worth resubmitting just [the GB SIO patch](https://github.com/mgba-emu/mgba/pull/2258/commits/ccf1275909ef78fb67b999662d371962ec784bbc)? (Possibly along with the changes to PrinterView to use it? Though that might be better left to someone who'd need less guidance than me.)
- I'm planning to maintain the ext-print patch for my own use, until the scripting interface is ready. I figure I don't need to worry as much about the dire state it's in, if it isn't going upstream. (And if I need the multiplayer I assume I broke, I can just switch EXEs.) That said, is there another part of mGBA that you think it might be easier for me to hack this sort of thing into, irrespective of whether the result would be appropriate for general use?

endrift:
The GB SIO patch might be, but I'd need to look over the printer spec again to double check. As for the hack, that's the sort of thing I'd usually put as a backend feature, but there's no backend way to spawn subprocesses at the moment. I may add one though, as I'm using it a bit for the updater, with ifdefs per platform.

PR #2284, opened 2021-08-20
CInema: Fix Windows/MSYS2 build

### LINKED ISSUE #1857: MSYS2/mingw-w64: Seemingly missing dependencies
Hi there,

I just tried to compile the latest mgba-master with my trusty MSYS2/mingw-w64 toolchain. However, I ran into the following issue at the cmake stage:
```
CMake Error at CMakeLists.txt:863 (add_library):
  Target "mgba" links to target "BZip2::BZip2" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?


CMake Error at CMakeLists.txt:863 (add_library):
  Target "mgba" links to target "LibLZMA::LibLZMA" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?


CMake Error at src/platform/sdl/CMakeLists.txt:112 (add_executable):
  Target "mgba-sdl" links to target "BZip2::BZip2" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?


CMake Error at src/platform/sdl/CMakeLists.txt:112 (add_executable):
  Target "mgba-sdl" links to target "LibLZMA::LibLZMA" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?


CMake Error at src/platform/qt/CMakeLists.txt:286 (add_executable):
  Target "mgba-qt" links to target "BZip2::BZip2" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?


CMake Error at src/platform/qt/CMakeLists.txt:286 (add_executable):
  Target "mgba-qt" links to target "LibLZMA::LibLZMA" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?


-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

I haven't done any bisecting this, but with the same toolchain (no updates inbetween), I was able to successfully build it about two weeks ago.

gifvex:
- Move municode flag from c364a65 into Qt. The previous placement causes linking errors with wmain/wWinMain in CMake's CheckFunctionExists (so everything fails) and CInema.
- ~~Prioritize FindPkgConfig if available to avoid incomplete package configs like libzip's. If this causes issues with other builds, find_feature could take a parameter for priority. Fixes #1857~~
- cinema-main.c: _GNU_SOURCE needs to be defined before stdio.h is included and MinGW's mkdir takes 1 argument


src/platform/cmake/FindFeature.cmake
```diff
@@ -12,9 +12,9 @@ function(find_feature FEATURE_NAME FEATURE_REQUIRES)
 		set(FOUND OFF)
 		foreach(REQUIRE ${NAMELIST})
 			if(NOT ${REQUIRE}_FOUND)
-				find_package(${REQUIRE} QUIET)
+				pkg_search_module(${REQUIRE} QUIET ${REQUIRE})
 				if(NOT ${REQUIRE}_FOUND)
-					pkg_search_module(${REQUIRE} ${REQUIRE})
+					find_package(${REQUIRE})
```
endrift, line 17: I cannot condone this change. The issue appears to be upstream (I filed a bug, but I'm unsure what to actually do about it...CMake is bad that way), and this is an actual behavior change from intended. Please build with libzip disabled in the meantime.

src/platform/test/cinema-main.c
```diff
@@ -3,6 +3,10 @@
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#ifdef HAVE_VASPRINTF
+#define _GNU_SOURCE
+#endif
```
endrift, line 8: This is not the correct place to define this; it should be done in CMake, and it should already be defined, though for some reason it might not be getting through to CInema.

src/platform/test/cinema-main.c
```diff
@@ -715,7 +719,11 @@ static struct VDir* _makeOutDir(const char* testName) {
 		strncpy(pathEnd, testName, len);
 		pathEnd += len;
 
+#ifndef __MINGW32__
```
endrift, line 722:
```suggestion
#ifndef _WIN32
```

endrift (changes requested):
Most of these changes are not quite right, though one I cannot really allow at all (as marked). The others are solvable however.

src/platform/test/cinema-main.c
gifvex: When building on MSYS2, the definition for vasprintf exists in libmingwex.a (automatically linked). That succeeds the linking-only verification from check_function_exists which both declares and references a function name. CInema references vasprintf but does not get a declaration because stdio.h has it behind _GNU_SOURCE.

So, I'd like to ask which impact level is preferred for other options:
1. Modify find_function to take a list of defines to keep grouped with the HAVE_ flags, for example `find_function(vasprintf _GNU_SOURCE)` appends `HAVE_VASPRINTF _GNU_SOURCE` to FUNCTION_DEFINES
2. Check for HAVE_VASPRINTF and MINGW somewhere to add the _GNU_SOURCE defintion for CInema

endrift: You're thinking about this wrong. The problem isn't "this function needs this define on mingw", the problem is "this define needs to be set on mingw regardless", and the question is why it's set for the main binary and not CInema.

gifvex: I don't think it is, cinema-main.c is the only file with `asprintf` in any form. If you mean why the Linux CInema has it set and this is a Windows issue, it's from
https://github.com/mgba-emu/mgba/blob/7c57a5cc13c3c009916a59189d7784474aee29c0/CMakeLists.txt#L221-L223

The other information was to explain why CInema is getting the HAVE_ flag without the declaration, in other words because CheckFunctionExists.c declares it iself and only verifies linking.

gifvex:
- Reverted find_feature priority
  - I used the USE_LIBZIP=OFF flag to work on another branch earlier this month and didn't notice any issues, so it's fine on the development build anyway
- Defined _GNU_SOURCE for everything instead of only cinema-main.c
- Use _WIN32 for other Windows builds' mkdir too

The scope of this is now only CInema without the libzip accomodation

endrift:
Sorry we had to reduce the scope of the PR, but it's a difficult issue to figure out properly since in part it seems like CMake is just full of bad ideas.

PR #2319, opened 2021-09-30
msvc : fix compilation for c11/c17 lang standards

shuffle2:
msvc currently supports up to c17 but does not implement some optional features. see https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160 and https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-160

### DIFF
--- a/include/mgba-util/threading.h
+++ b/include/mgba-util/threading.h
@@ -11,7 +11,7 @@
 CXX_GUARD_START
 
 #ifndef DISABLE_THREADING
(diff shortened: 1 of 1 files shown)


endrift:
> msvc currently supports up to c17

Funny joke. Wonder when they'll get support for the additions people actually use, like _Atomic.

PR #2330, opened 2021-10-15
[debugger] Silence a -Wswitch warning out of gdb-stub.c

fpdotmonkey:
No one likes unhandled branches, and the compiler was just kindly reminding us
to take care of them.  Well, remind us no more, compiler.


endrift:
I don't like this approach. This assumes that for any other watchpoint type we never want to do anything, even if we add more. We should instead have an empty case for the existing types (that can't be hit as-is anyway, so it's effectively dead code). That way if more are added later it will raise warnings again to let us know to update the code.

PR #2345, opened 2021-11-06
[SDL]: Add logFile support

Lorenzooone:
Since we were talking about it here: https://twitter.com/Lorenzooone/status/1456780759012872196 , I decided to implement logging to file for SDL.

Feel free to point out any and all issues you might have with the code!


Lorenzooone:
Btw, I copied the default behaviour which makes it so you don't get the popup for crashes when you get a FATAL unless the filter allows seeing FATALs. It can be observed here https://github.com/mgba-emu/mgba/blob/master/src/platform/qt/LogController.cpp at postLog and here https://github.com/mgba-emu/mgba/blob/master/src/core/thread.c at _mCoreLog.

Is it on purpose though? I get not logging them, but not getting the popup entirely seems weird to me.

src/platform/sdl/sdl-log.c
```diff
@@ -0,0 +1,87 @@
+/* Copyright (c) 2013-2015 Jeffrey Pfau
```
endrift, line 1: This whole file should probably be merged into `main.c`.

src/platform/sdl/sdl-log.c
```diff
@@ -0,0 +1,87 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include "sdl-log.h"
+#include <mgba/core/core.h>
+#include <mgba/core/log.h>
+#include <mgba/core/thread.h>
+#include <mgba/core/config.h>
+
+static bool _logToStdout;
+static FILE* _logFile;
```
endrift, line 13: Please use the `VFile` API.

src/platform/sdl/sdl-log.c
```diff
@@ -0,0 +1,87 @@
...
+{
+	int value = 0;
+	bool searched = mCoreConfigGetValue(&core->config, key);
+	if(searched)
+	{
+		mCoreConfigGetIntValue(&core->config, key, &value);
+		return value != 0 ? true: false;
+	}
+	return defaultValue;
+}
```
endrift, line 28: The existing convention is to use an int value called `fakeBool` and coerce to bool as needed. I may add an `mCoreConfigGetBoolValue` helper at some point, but not in this PR, and not a private one.

src/platform/sdl/sdl-log.c
```diff
@@ -0,0 +1,87 @@
...
+
+struct mLogger getLogger(struct mCore* core)
```
endrift, line 30: Do not return a struct larger than 8 bytes. This could all be handled in a void function after moving to `main.c` since the logger will probably wind up being a static variable anyway.

endrift (changes requested):
First pass review definitely has some issues that will need to be cleaned up. You also don't match existing brace style, but I didn't go and comment every single one of those.

endrift:
> Btw, I copied the default behaviour which makes it so you don't get the popup for crashes when you get a FATAL unless the filter allows logging FATALs. It can be observed here https://github.com/mgba-emu/mgba/blob/master/src/platform/qt/LogController.cpp at postLog.
> 
> Is it on purpose though? I get not logging them, but not getting the popup entirely seems weird to me.

That does not sound like it was on purpose, no.

src/platform/sdl/sdl-log.c
Lorenzooone: This commit should solve that (and the brace style): https://github.com/mgba-emu/mgba/pull/2345/commits/4137ad686a57123ce1d578ad01d468971ffdd1b4

src/platform/sdl/sdl-log.c
Lorenzooone: Should be fixed in: https://github.com/mgba-emu/mgba/pull/2345/commits/02b165005bcfac32caf4ed53419a46ade0a6086f

src/platform/sdl/sdl-log.c
Lorenzooone: Should be fixed in: https://github.com/mgba-emu/mgba/pull/2345/commits/02b165005bcfac32caf4ed53419a46ade0a6086f

src/platform/sdl/sdl-log.c
Lorenzooone: Should be fixed in: https://github.com/mgba-emu/mgba/pull/2345/commits/02b165005bcfac32caf4ed53419a46ade0a6086f

src/platform/sdl/main.c
```diff
@@ -321,3 +337,64 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+	// Create the filter
+	mLogFilterInit(&_filter);
+	mLogFilterSet(&_filter, "gba.bios", mLOG_STUB | mLOG_FATAL);
+	mLogFilterSet(&_filter, "core.status", mLOG_ALL & ~mLOG_DEBUG);
+	_filter.defaultLevels = filterLevels;
+
+	// Fill the logger
+	_logger.log = _mCoreLog;
+	_logger.filter = &_filter;
+}
```
endrift, line 373: Why are these separate functions if they're only ever called once?

src/platform/sdl/main.c
```diff
@@ -321,3 +337,64 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
 	SDL_Quit();
 }
+
+void setLogger(struct mCore* core) {
+	int fakeBool = 0;
+	bool logToStdout = true;
+	bool logToFile = false;
+
+	if (mCoreConfigGetIntValue(&core->config, "logToStdout", &fakeBool))
+		logToStdout = fakeBool;
```
endrift, line 347:
```suggestion
	if (mCoreConfigGetIntValue(&core->config, "logToStdout", &fakeBool)) {
		logToStdout = fakeBool;
	}
```
Style nit

src/platform/sdl/main.c
```diff
@@ -321,3 +337,64 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+	if (mCoreConfigGetIntValue(&core->config, "logToFile", &fakeBool))
+		logToFile = fakeBool;
```
endrift, line 349:
```suggestion
	if (mCoreConfigGetIntValue(&core->config, "logToFile", &fakeBool)) {
		logToFile = fakeBool;
	}
```
Style nit

src/platform/sdl/main.c
```diff
@@ -321,3 +337,64 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+	_setLogger(logToStdout, logToFile, logFile, filterLevels);
+}
+
+void _setLogger(bool logToStdout, bool logToFile, const char* logFile, int filterLevels) {
+	// Assign basic static variables
+	_logToStdout = logToStdout;
+	_logFile = NULL;
+	
+	if(logToFile && logFile)
+		_logFile = VFileOpen(logFile, O_WRONLY | O_CREAT | O_TRUNC);
```
endrift, line 362:
```suggestion
	if(logToFile && logFile) {
		_logFile = VFileOpen(logFile, O_WRONLY | O_CREAT | O_TRUNC);
	}
```
Style nit

endrift, line 362: I don't believe the Qt frontend truncates the log file; I believe it appends instead. It's best to keep this consistent.

src/platform/sdl/main.c
```diff
@@ -321,3 +337,64 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+
+	// Create the filter
+	mLogFilterInit(&_filter);
+	mLogFilterSet(&_filter, "gba.bios", mLOG_STUB | mLOG_FATAL);
```
endrift, line 366: This line is unneeded; the reason mLOG_STUB was put in there in the Qt frontend was to give users a dialog when an unimplemented BIOS function is hit, but the SDL frontend doesn't have dialogs.

src/platform/sdl/main.c
```diff
@@ -321,3 +337,64 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+	const int MAX_BUF = 1024;
+	char* buffer = malloc(MAX_BUF);
+
+	int length = 0;
+	length += snprintf(buffer+length, MAX_BUF-length, "%s: ", mLogCategoryName(category));
+	length += vsnprintf(buffer+length, MAX_BUF-length, format, args);
+	length += snprintf(buffer+length, MAX_BUF-length, "\n");
+
+	if(_logToStdout)
+		printf("%s", buffer);
```
endrift, line 394:
```suggestion
	if(_logToStdout) {
		printf("%s", buffer);
	}
```
Style nit

src/platform/sdl/main.c
```diff
@@ -321,3 +337,64 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+		return;
+	}
+
+	const int MAX_BUF = 1024;
+	char* buffer = malloc(MAX_BUF);
+
+	int length = 0;
+	length += snprintf(buffer+length, MAX_BUF-length, "%s: ", mLogCategoryName(category));
+	length += vsnprintf(buffer+length, MAX_BUF-length, format, args);
+	length += snprintf(buffer+length, MAX_BUF-length, "\n");
```
endrift, line 391:
```suggestion
	length += snprintf(buffer + length, MAX_BUF - length, "%s: ", mLogCategoryName(category));
	length += vsnprintf(buffer + length, MAX_BUF - length, format, args);
	length += snprintf(buffer + length, MAX_BUF - length, "\n");
```
Style nits

src/platform/sdl/main.c
```diff
@@ -321,3 +337,64 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+	if (thread && level == mLOG_FATAL) {
+		mCoreThreadMarkCrashed(thread);
+	}
+
+	if (!mLogFilterTest(logger->filter, category, level)) {
+		return;
+	}
+
+	const int MAX_BUF = 1024;
+	char* buffer = malloc(MAX_BUF);
```
endrift, line 386: Any reason this buffer isn't static on the stack? You could just use sizeof() in that case too, and not worry about freeing it.

endrift (changes requested):
Mostly style nits and some questionable implementation choices, but once that's flattened out I think it'll be pretty decent to merge.

src/platform/sdl/main.c
Lorenzooone: I wanted to keep them logically separated, but you're right, as just keeping them one after the other should suffice.

src/platform/sdl/main.c
Lorenzooone: True. Snprintf already handles the buffer too and makes sure it doesn't go out of bounds.

Lorenzooone: Fixed in https://github.com/mgba-emu/mgba/pull/2345/commits/9440ca24e9b66396445955640548bab981801c2d .

src/platform/sdl/main.c
Lorenzooone: Fixed in https://github.com/mgba-emu/mgba/pull/2345/commits/9440ca24e9b66396445955640548bab981801c2d .

(I just removed the STUB part though... Don't know if you wanted the FATAL removed as well)

src/platform/sdl/main.c
Lorenzooone: Fixed in https://github.com/mgba-emu/mgba/pull/2345/commits/9440ca24e9b66396445955640548bab981801c2d .

src/platform/sdl/main.c
Lorenzooone: Fixed in https://github.com/mgba-emu/mgba/pull/2345/commits/9440ca24e9b66396445955640548bab981801c2d .

src/platform/sdl/main.c
```diff
@@ -38,11 +38,20 @@
 #include <signal.h>
 
 #define PORT "sdl"
+#define MAX_LOG_BUF 1024
 
 static void mSDLDeinit(struct mSDLRenderer* renderer);
 
 static int mSDLRun(struct mSDLRenderer* renderer, struct mArguments* args);
 
+void setLogger(struct mCore* core);
```
endrift, line 47: This function should be static (and as such probably prefixed with an _).

src/platform/sdl/main.c
```diff
@@ -38,11 +38,20 @@
...
+static void _mCoreLog(struct mLogger* logger, int category, enum mLogLevel level, const char* format, va_list args);
+
+static bool _logToStdout;
+static struct VFile* _logFile;
```
endrift, line 51:
```suggestion
static bool _logToStdout = true;
static struct VFile* _logFile = NULL;
```

src/platform/sdl/main.c
```diff
@@ -321,3 +337,61 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
 
 	SDL_Quit();
 }
+
+void setLogger(struct mCore* core) {
+	int fakeBool = 0;
+	bool logToStdout = true;
```
endrift, line 343: This var is unnecessary if we just directly load into _logToStdout.

src/platform/sdl/main.c
```diff
@@ -321,3 +337,61 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+	bool logToFile = false;
+
+	if (mCoreConfigGetIntValue(&core->config, "logToStdout", &fakeBool)) {
+		logToStdout = fakeBool;
```
endrift, line 347:
```suggestion
		_logToStdout = fakeBool;
```

src/platform/sdl/main.c
```diff
@@ -321,3 +337,61 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+	}
+	if (mCoreConfigGetIntValue(&core->config, "logToFile", &fakeBool)) {
+		logToFile = fakeBool;
+	}
+	const char* logFile = mCoreConfigGetValue(&core->config, "logFile");
+
+	// Assign basic static variables
+	_logToStdout = logToStdout;
+	_logFile = NULL;
```
endrift, line 356: These lines are no longer necessary with the static initialization above.

src/platform/sdl/main.c
```diff
@@ -321,3 +337,61 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+	
+	if(logToFile && logFile) {
+		_logFile = VFileOpen(logFile, O_WRONLY | O_CREAT | O_APPEND);
+	}
+
+	// Create the filter
+	mLogFilterInit(&_filter);
+	mLogFilterSet(&_filter, "gba.bios", mLOG_FATAL);
```
endrift, line 364: mLOG_FATAL is in the default filter so this isn't needed anymore.

src/platform/sdl/main.c
```diff
@@ -321,3 +337,61 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+	mLogFilterSet(&_filter, "core.status", mLOG_ALL & ~mLOG_DEBUG);
```
endrift, line 365: The point of this was so that status messages bubble up into the OSD on the Qt version. However, there is no OSD in the SDL version, so this might create spam for users by default. I'm unsure if that's a _problem_ though.

src/platform/sdl/main.c
```diff
@@ -321,3 +337,61 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+	if (!mLogFilterTest(logger->filter, category, level)) {
+		return;
+	}
+
+	char buffer[MAX_LOG_BUF];
+
+	int length = 0;
+	length += snprintf(buffer + length, sizeof(buffer) - length, "%s: ", mLogCategoryName(category));
+	length += vsnprintf(buffer + length, sizeof(buffer) - length, format, args);
+	length += snprintf(buffer + length, sizeof(buffer) - length, "\n");
```
endrift, line 388: You should probably make sure length never exceeds `sizeof(buffer)`, as the argument to snprintf is unsigned, so this could lead to a buffer overflow.

endrift (changes requested):
_Almost_ there. Things were cleanly organized enough this time that I could go over it in detail. Just some minor nits about temporary variables and initialization and one security concern. Once those are addressed it'll be good to merge.

src/platform/sdl/main.c
Lorenzooone: Hmm... Are you 100% sure? Because I tried removing it before, and I'd get a list of values that were in the registers when it went into VBlank at the end of my test ROM's execution.

If I leave that there, then I don't get that list...

src/platform/sdl/main.c
Lorenzooone: Fixed in https://github.com/mgba-emu/mgba/pull/2345/commits/453ffc29a1418b6ae327fc528e8c1f39d2cbc81e (although it might not be pretty... Let me know whether this is fine or not).

src/platform/sdl/main.c
Lorenzooone: Fixed in https://github.com/mgba-emu/mgba/pull/2345/commits/fac41951f3984e0395765df6c02f0fb02c0ab800 .

src/platform/sdl/main.c
Lorenzooone: Fixed in https://github.com/mgba-emu/mgba/pull/2345/commits/fac41951f3984e0395765df6c02f0fb02c0ab800 .

src/platform/sdl/main.c
Lorenzooone: Fixed in https://github.com/mgba-emu/mgba/pull/2345/commits/fac41951f3984e0395765df6c02f0fb02c0ab800 .

src/platform/sdl/main.c
Lorenzooone: Just checked, and it makes it so you can't edit the logLevel for gba.bios by using the -l option...

Lorenzooone: Btw, the default, if no -l is specified, is 0. So no logging whatsoever...

endrift: That's odd, the default is supposed to be 3--fatal and error.

src/platform/sdl/main.c
endrift: Well, the idea was to just not call the further snprintfs if length exceeds the size of the buffer. This merely clamps the length variable, which seems like an odd approach.

Lorenzooone: _logFile->write uses length a bit below that piece of code. Doing it the way you described will work for sure. One just has to remember to also check there whether the length exceeds the buffer's size or not.

Let me know which one you prefer. It's an easy change, tbh.

src/platform/sdl/main.c
endrift: Let's remove both of the log level overrides then, and manage a default log level of 3 in a later commit. It looks like it's only 3 in the Qt frontend right now.

src/platform/sdl/main.c
endrift: I would prefer not to use a mess of ternary operators to be sure :stuck_out_tongue:. Mostly I just don't want to invoke the snprintfs at all if they're not going to do anything due to the buffer already being full.

src/platform/sdl/main.c
Lorenzooone: So, I added a default of WARN | FATAL | ERROR, which is the one QT uses for anything that isn't gba.bios and core.status, for which it uses this line and the one below this one.

The issue is that, by removing this line in particular, if you select mLOG_DEBUG (in order to see messages from the game, which is what I use the log function for), you get a log message for each single SWI call. Which can be VERY spammy.

You can see what I mean below, with before and after.

[test_prev.log](https://github.com/mgba-emu/mgba/files/7611047/test_prev.log)
[test.log](https://github.com/mgba-emu/mgba/files/7611048/test.log)

Maybe adding some sort of configuration parameter to this? Otherwise I can just keep a fork of mine and let people use that for my use case.

endrift: That's the already existing behavior so it's not a good idea to change it in a mostly unrelated PR. That said, you can change this with e.g. -C logLevel.gba.bios=0 to turn off all BIOS logging.

Lorenzooone: That does work in Qt, and it is... a good solution!
But not in SDL (simply because the code isn't there yet).

Will definetly look into putting it in though. Seems very useful.

endrift: That's strange? It worked for me on master. I set -l -1 to enable all logs and then that to disable BIOS logs and it worked. Maybe the issue is local to your branch?

Lorenzooone: Yeah, because SDL in main uses the thread.c implementation of the logger, which has this line:
mLogFilterLoad(threadContext->logger.d.filter, &core->config);

Looking at that file makes me wonder...

Should I add that line to this file, in order to keep logger and filter initialization together, or should I remove the filter initialization entirely from here, since thread.c will handle that...?

Lorenzooone: Fixed in https://github.com/mgba-emu/mgba/pull/2345/commits/e938be4ba83a65f1012f4943af60eb195f7af610 , let me know if you'd prefer setting the filter to NULL for the logger and letting thread.c initialize it instead.

src/platform/sdl/main.c
Lorenzooone: Fixed in https://github.com/mgba-emu/mgba/pull/2345/commits/e938be4ba83a65f1012f4943af60eb195f7af610 .

src/platform/sdl/main.c
```diff
@@ -321,3 +338,63 @@ static void mSDLDeinit(struct mSDLRenderer* renderer) {
...
+	char buffer[MAX_LOG_BUF];
+
+	// Prepare the string
+	size_t length = snprintf(buffer, sizeof(buffer), "%s: ", mLogCategoryName(category));
+	if (length < sizeof(buffer)) {
+		length += vsnprintf(buffer + length, sizeof(buffer) - length, format, args);
+		if (length < sizeof(buffer)) {
+			length += snprintf(buffer + length, sizeof(buffer) - length, "\n");
+		}
+	}
```
endrift, line 386:
```suggestion
	}
	if (length < sizeof(buffer)) {
		length += snprintf(buffer + length, sizeof(buffer) - length, "\n");
	}
```
The can actually be separated since the latter one will never be true if the former one isn't.

endrift (approved):
Looks good, just one nit that's not super important. If you don't integrate it by the time I get home I'll just merge it.

Though I'm beginning to wonder if this shouldn't just be put in the default thread logger at this point.

PR #2366, opened 2021-11-28
NetBSD build fixes

alarixnia:
We've had these fixes in pkgsrc for well over a year now. Please merge them!

- Use `newlocale` and `snprintf_l` without using `uselocale` - NetBSD does not have per-thread locales.
- `popcount32` is in libc on NetBSD, don't override its definition.
- `pthread_setname_np` takes a printf format string.


include/mgba-util/platform/posix/threading.h
```diff
@@ -94,6 +94,8 @@ static inline int ThreadSetName(const char* name) {
 #elif defined(__HAIKU__)
 	rename_thread(find_thread(NULL), name);
 	return 0;
+#elif defined(__NetBSD__)
+	return pthread_setname_np(pthread_self(), "%s", (void *)name);
```
endrift, line 98: Why does this need to be cast to void*? This is C code.

CMakeLists.txt
```diff
@@ -368,13 +370,17 @@ if(USE_PTHREADS)
...
 	list(APPEND FUNCTION_DEFINES HAVE_LOCALE)
 	if (HAVE_SNPRINTF_L)
 		list(APPEND FUNCTION_DEFINES HAVE_SNPRINTF_L)
 	endif()
 endif()
 
+if(HAVE_USELOCALE OR APPLE)
+	list(APPEND FUNCTION_DEFINES HAVE_USELOCALE)
+endif()
+
```
endrift, line 383: These lines are unnecessary--`HAVE_USELOCALE` will be defined within `find_function`.

CMakeLists.txt
```diff
@@ -320,6 +320,8 @@ find_function(setlocale)
 find_function(snprintf_l)
 find_function(uselocale)
 
+find_function(popcount32)
```
endrift, line 323: Make sure to add this to `flags.h.in` since it's used in a header file.

endrift (commented):
Most of these changes are easily mergeable, but there are some lingering questions first.

include/mgba-util/platform/posix/threading.h
alarixnia: Technically unnecessary, but it gets rid of a "discards const" warning.

endrift: Ohh fair enough. Please add a space between the cast and the variable then.

endrift:
Was there really no HAVE flags section yet?? I need to fix that.

PR #2395, opened 2021-12-27
Send an S05 for write only watchpoint instead of T05

aldelaro5:
This bypasses GDB's internal logic to check old vs new value which is not only something mgba already handles, but it's also wrong because GDB cannot understand the concept of savestates with mgba. Sending less information to GDB allows GDB to let the inferior stop instead of doing the watchpoint check. It also solves another issue: the PC GDB will see is one instruction after the write instead of 2 since we no longer let GDB single step to see the effects of the write.

~~I would also like to mention something that's not strictly related to this PR: this allows us in theory to bypass an imo severe limitation of GDB which is that it's impossible to request unchanged write only watchpoint. You can only request write with changes or read AND write, but as a reverse engineer, I car about more writes then the former and the later becomes an inconvenience since it happens often that something is read often, but written to SOMETIMES. My question is would it make sense that a type 2 breakpoint (`watch` command) would add a `WATCHPOINT_WRITE` type instead of `WATCHPOINT_WRITE_CHANGE`?~~

~~Regardless of what you decide on this matter, this current PR is required for either to work correctly anyway, but I am curious about your thoughts on this. My biggest fear is if you would use the gdb stub for game dev, you would expect GDB's behavior to be unchanged. The reality is this problem has to be properly addressed in GDB itself.~~ 
EDIT: disregard this, I opened a new draft pr for this #2400

### DIFF
--- a/src/debugger/gdb-stub.c
+++ b/src/debugger/gdb-stub.c
@@ -56,16 +56,16 @@ static void _gdbStubEntered(struct mDebugger* debugger, enum mDebuggerEntryReaso
 			const char* type = 0;
 			switch (info->type.wp.watchType) {
 			case WATCHPOINT_WRITE:
-				if (info->type.wp.newValue == info->type.wp.oldValue) {
-					if (stub->d.state == DEBUGGER_PAUSED) {
-						stub->d.state = DEBUGGER_RUNNING;
-					}
-					return;
-				}
 				// Fall through
 			case WATCHPOINT_WRITE_CHANGE:
 				type = "watch";
-				break;
+				// We send S05 instead of T05watch because it bypasses GDB's internal
+				// logic to check if the value changed. We need to bypass it because
+				// we knows more than GDB in regards to using savestates and mgba already
+				// has this logic anyway
+				snprintf(stub->outgoing, GDB_STUB_MAX_LINE - 4, "S%02x", SIGTRAP);
+				_sendMessage(stub);
+				return;
 			case WATCHPOINT_READ:
 				type = "rwatch";
 				break;


src/debugger/gdb-stub.c
```diff
@@ -65,7 +65,13 @@ static void _gdbStubEntered(struct mDebugger* debugger, enum mDebuggerEntryReaso
 				// Fall through
 			case WATCHPOINT_WRITE_CHANGE:
 				type = "watch";
-				break;
+				// We send S05 instead of T05watch because it bypasses GDB's internal
+				// logic to check if the value changed. We need to bypass it because
+				// mgba knows more than GDB in regards to using savestates and mgba already
```
endrift, line 70:
```suggestion
				// we knows more than GDB in regards to using savestates and we already
```

src/debugger/gdb-stub.c
```diff
@@ -65,7 +65,13 @@ static void _gdbStubEntered(struct mDebugger* debugger, enum mDebuggerEntryReaso
 				// Fall through
```
endrift, line 65: The logic inside of WATCHPOINT_WRITE should probably be removed then. However, gdb will get confused about why it's breaking.

endrift (changes requested):
This will confuse the hell out of GDB and mess up its accounting. I'm unsure if it should be merged. But even so, it needs fixing.

src/debugger/gdb-stub.c
```diff
@@ -55,17 +55,15 @@ static void _gdbStubEntered(struct mDebugger* debugger, enum mDebuggerEntryReaso
 		if (info) {
 			const char* type = 0;
 			switch (info->type.wp.watchType) {
-			case WATCHPOINT_WRITE:
```
endrift, line 58: I said the logic, not the whole case.

endrift (changes requested):

aldelaro5:
> This will confuse the hell out of GDB and mess up its accounting. I'm unsure if it should be merged. But even so, it needs fixing.

It turns out I did tested what would happen on both mgba and dolphin and you are correct that gdb can't know that you broke due to a watchpoint meaning it can't for example count the amount of time one was hit. What will happen is GDB will let the inferior stop and continue as normal. As a reverser, this is IDEAL because I don't have to worry about when I savestates or ctrl+c manually after I load one. However, the question is more if the watchpoint works in a way that GDB CANNOT possibly be correct while mgba can, wouldn't it still make sense to not tell GDB?

Honestly, I am open to discussions here. This is mostly an issue with GDB's limitation and not mgba, but it's an issue annoying enough that I would want it addressed, even if it takes more effort.

src/debugger/gdb-stub.c
aldelaro5: sorry, done.

aldelaro5:
I am closing this because as discussed on discord, we want to integrate this pr with #2400 via a radio button setting in the GDB window. Since this solution requires one pr instead of 2, one has to close and the other one is in draft so I might as well close this one and move the implementation of this to #2400

PR #2399, opened 2021-12-30
GDB Stub: Add target descriptior and memory map xml

aldelaro5:
This allows gdb to know the memory mapping as well as the architecture once connected.

I think I got the mappings correct, but lmk if it's wrong.


src/debugger/gdb-stub.c
```diff
@@ -29,6 +29,26 @@ enum {
 	MACH_O_ARM_V4T = 5
 };
 
+const char* TARGET_XML = "<target version=\"1.0\">"
```
endrift, line 32:
```suggestion
static const char* TARGET_XML = "<target version=\"1.0\">"
```

src/debugger/gdb-stub.c
```diff
@@ -29,6 +29,26 @@ enum {
...
+                             "<memory type=\"ram\" start=\"0x04000000\" length=\"0x00000400\"/>"
+                             "<memory type=\"ram\" start=\"0x05000000\" length=\"0x00000400\"/>"
+                             "<memory type=\"ram\" start=\"0x06000000\" length=\"0x00018000\"/>"
+                             "<memory type=\"ram\" start=\"0x07000000\" length=\"0x00000400\"/>"
+                             "<memory type=\"ram\" start=\"0x08000000\" length=\"0x02000000\"/>"
+                             "<memory type=\"ram\" start=\"0x0A000000\" length=\"0x02000000\"/>"
+                             "<memory type=\"ram\" start=\"0x0C000000\" length=\"0x02000000\"/>"
+                             "<memory type=\"ram\" start=\"0x0E000000\" length=\"0x00020000\"/>"
+                             "<memory type=\"ram\" start=\"0x0F000000\" length=\"0x00020000\"/>"
+                             "</memory-map>";
```
endrift, line 50: This needs to be dynamically generated, as the SRAM region can change. As I mentioned, you should be using core->listMemoryBlocks to do this. Further, you have ROM listed as ram, for...some reason.

src/debugger/gdb-stub.c
```diff
@@ -433,7 +453,35 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
 		}
 		message = end + 1;
 	}
-	strncpy(stub->outgoing, "swbreak+;hwbreak+", GDB_STUB_MAX_LINE - 4);
+	strncpy(stub->outgoing, "swbreak+;hwbreak+;qXfer:features:read+;qXfer:memory-map:read+", GDB_STUB_MAX_LINE - 4);
+}
+
+static void _processQXferCommand(struct GDBStub* stub, const char* params, const char* data) {
+	unsigned int offset = 0;
+	unsigned int length = 0;
```
endrift, line 461:
```suggestion
	unsigned offset = 0;
	unsigned length = 0;
```

src/debugger/gdb-stub.c
```diff
@@ -433,7 +453,35 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+
+	unsigned int index = 0;
```
endrift, line 463:
```suggestion
	unsigned index;
```

src/debugger/gdb-stub.c
```diff
@@ -433,7 +453,35 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	while (params[index] != ',') {
+		offset <<= 4;
+		offset |= _hex2int(&params[index], 1);
+		index++;
+	}
```
endrift, line 468:
```suggestion
	for (index = 0; params[index] != ','; ++index) {
		offset <<= 4;
		offset |= _hex2int(&params[index], 1);
	}
```

endrift: Also there may already be a function for constructing these hex numbers automatically. I would need to check.

src/debugger/gdb-stub.c
```diff
@@ -433,7 +453,35 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+
+	index++;
+	while (index < strlen(params) - 3) {
+		length <<= 4;
+		length |= _hex2int(&params[index], 1);
+		index++;
+	}
```
endrift, line 475:
```suggestion
	++index;
	for (; params[index] != ','; ++index) {
		offset <<= 4;
		offset |= _hex2int(&params[index], 1);
	}
```

src/debugger/gdb-stub.c
```diff
@@ -433,7 +453,35 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+
+	memset(stub->outgoing, 0, GDB_STUB_MAX_LINE);
+	if (strlen(data) - offset < length) {
```
endrift, line 478: These values are unsigned and thus vulnerable to underflow. Also what is this check for? Telling GDB it's reached the end of the value?

```suggestion
	if (strlen(data) < length + offset) {
```

endrift (changes requested):

src/debugger/gdb-stub.c
aldelaro5: Sending an l first means it's the last packet, an m means there is more data after, GDB handles the logic of what offset and length to send.

this check is to know if we are about to read until the end which would change the first character to send and the amount of bytes to copy.

src/debugger/gdb-stub.c
aldelaro5: I asked in the discord and someone told me that the first 0x4000 would be read only, but that the rest are writable.

rom just means it's read only and ram means read/write. If it's wrong, then which regions should be marked as rom? the 3 cart one + bios? are there others?

src/debugger/gdb-stub.c
aldelaro5: Just to confirm: you just want to convert the while to a for? because your suggestion changes the check in a way that might not work (there is only one comma, the length - 3 is to ignore the checksum so this just checks if it's the end of the packet minus the checksum part).

endrift: Sorry, that was just me being blind and copy/pasting. It should be fixed now.

src/debugger/gdb-stub.c
endrift: The API in question contains this information...

Also, the ROM is read-only except in very specific circumstances: there is a small region reserved for GPIO access, but that only exists in a handful of games, and there is another small (but not as small) region which is used for EEPROM in some games. Beyond that, it cannot be modified except by internal API (the rawWrite* calls, specifically).

aldelaro5:
I changed it so it generates dynamically, but only once then we cache it until we reconnect.

src/debugger/gdb-stub.c
```diff
@@ -433,7 +440,52 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	index += 26;
+	const struct mCoreMemoryBlock* blocks;
+	size_t nBlocks = stub->d.core->listMemoryBlocks(stub->d.core, &blocks);
+	size_t i;
+	for (i = 0; i < nBlocks; ++i) {
+		if (!(blocks[i].flags & mCORE_MEMORY_MAPPED))
+			continue;
+
+		char* type = blocks[i].flags & mCORE_MEMORY_WRITE ? "ram" : "rom";
+		snprintf(&memoryMap[index], 60, "<memory type=\"%s\" start=\"0x%08x\" length=\"0x%08x\"/>", type, blocks[i].start, blocks[i].size);
```
endrift, line 485: This is not at all safe and not how you should be doing this, The size  you should be passing is the size remaining in the buffer, not the size of the string you're copying...

src/debugger/gdb-stub.c
```diff
@@ -433,7 +440,52 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+		length = strlen(data) - offset;
+		stub->outgoing[0] = 'l';
+	} else {
+		stub->outgoing[0] = 'm';
+	}
+	strncpy(&stub->outgoing[1], &data[offset], length);
+}
+
+static void _generateMemoryMapXml(struct GDBStub* stub, char* memoryMap) {
+	int index = 0;
```
endrift, line 474:
```suggestion
	size_t index = 0;
```

src/debugger/gdb-stub.c
```diff
@@ -433,7 +440,52 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	strncpy(memoryMap, "<memory-map version=\"1.0\">", 27);
+	index += 26;
```
endrift, line 476:
```suggestion
	index += strlen(memoryMap);
```

src/debugger/gdb-stub.c
```diff
@@ -433,7 +440,52 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	const struct mCoreMemoryBlock* blocks;
+	size_t nBlocks = stub->d.core->listMemoryBlocks(stub->d.core, &blocks);
+	size_t i;
+	for (i = 0; i < nBlocks; ++i) {
+		if (!(blocks[i].flags & mCORE_MEMORY_MAPPED))
+			continue;
```
endrift, line 482:
```suggestion
		if (!(blocks[i].flags & mCORE_MEMORY_MAPPED)) {
			continue;
		}
```

src/debugger/gdb-stub.c
```diff
@@ -433,7 +440,52 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+
+		char* type = blocks[i].flags & mCORE_MEMORY_WRITE ? "ram" : "rom";
+		snprintf(&memoryMap[index], 60, "<memory type=\"%s\" start=\"0x%08x\" length=\"0x%08x\"/>", type, blocks[i].start, blocks[i].size);
+		index += 59;
```
endrift, line 486: The output from snprintf is what you should be using to advance the index, as the number of characters is dynamic.

src/debugger/gdb-stub.c
```diff
@@ -452,6 +504,12 @@ static void _processQReadCommand(struct GDBStub* stub, const char* message) {
 		strncpy(stub->outgoing, "m1", GDB_STUB_MAX_LINE - 4);
 	} else if (!strncmp("sThreadInfo#", message, 12)) {
 		strncpy(stub->outgoing, "l", GDB_STUB_MAX_LINE - 4);
+	} else if (!strncmp("Xfer:features:read:target.xml:", message, 30)) {
+		_processQXferCommand(stub, message + 30, TARGET_XML);
+	} else if (!strncmp("Xfer:memory-map:read::", message, 22)) {
+		if (strlen(MEMORY_MAP_XML) == 0)
+			_generateMemoryMapXml(stub, MEMORY_MAP_XML);
```
endrift, line 511: This should be attached to the GDBStub struct, not global.

endrift (changes requested):
Lots of memory errors in this one...you need to be more careful about bounds.

aldelaro5:
> Lots of memory errors in this one...you need to be more careful about bounds.

yeah sorry, I don't typically deal with raw c strings like this (too used to std::string and such).

src/debugger/gdb-stub.c
```diff
@@ -433,7 +438,52 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+
+	unsigned index = 0;
+	for (index = 0; params[index] != ','; ++index) {
+		offset <<= 4;
+		offset |= _hex2int(&params[index], 1);
+	}
+
+	++index;
+	for (; index < strlen(params) - 3; ++index)
+	{
```
endrift, line 456:
```suggestion
	for (; index + 3 < strlen(params); ++index) {
```

src/debugger/gdb-stub.c
```diff
@@ -433,7 +438,52 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	}
+
+	memset(stub->outgoing, 0, GDB_STUB_MAX_LINE);
+	if (strlen(data) < length + offset) {
+		length = strlen(data) - offset;
+		stub->outgoing[0] = 'l';
+	} else {
+		stub->outgoing[0] = 'm';
+	}
+	strncpy(&stub->outgoing[1], &data[offset], length);
```
endrift, line 468: You should ensure that this doesn't overflow. The other side may not necessarily play nice.

src/debugger/gdb-stub.c
```diff
@@ -433,7 +438,52 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	size_t i;
+	for (i = 0; i < nBlocks; ++i) {
+		if (!(blocks[i].flags & mCORE_MEMORY_MAPPED)) {
+			continue;
+		}
+
+		char* type = blocks[i].flags & mCORE_MEMORY_WRITE ? "ram" : "rom";
+		index += snprintf(&memoryMap[index], GDB_STUB_MAX_LINE - index, "<memory type=\"%s\" start=\"0x%08x\" length=\"0x%08x\"/>", type, blocks[i].start, blocks[i].size);
+	}
+	strncpy(&memoryMap[index], "</memory-map>", 14);
```
endrift, line 486: Needs bounds checking.

src/debugger/gdb-stub.c
```diff
@@ -452,6 +502,12 @@ static void _processQReadCommand(struct GDBStub* stub, const char* message) {
 		strncpy(stub->outgoing, "m1", GDB_STUB_MAX_LINE - 4);
 	} else if (!strncmp("sThreadInfo#", message, 12)) {
 		strncpy(stub->outgoing, "l", GDB_STUB_MAX_LINE - 4);
+	} else if (!strncmp("Xfer:features:read:target.xml:", message, 30)) {
+		_processQXferCommand(stub, message + 30, TARGET_XML);
+	} else if (!strncmp("Xfer:memory-map:read::", message, 22)) {
+		if (strlen(stub->memoryMapXml) == 0)
+			_generateMemoryMapXml(stub, stub->memoryMapXml);
```
endrift, line 509:
```suggestion
		if (strlen(stub->memoryMapXml) == 0) {
			_generateMemoryMapXml(stub, stub->memoryMapXml);
		}
```

src/debugger/gdb-stub.c
```diff
@@ -710,6 +766,8 @@ bool GDBStubListen(struct GDBStub* stub, int port, const struct Address* bindAdd
 		goto cleanup;
 	}
 
+	memset(stub->memoryMapXml, 0, GDB_STUB_MAX_LINE);
```
endrift, line 769: This should be done in init, not here.

endrift (changes requested):
Still has memory errors. Please be careful.

src/debugger/gdb-stub.c
aldelaro5: I am not sure about this: moving it there means it's possible to open a connexion, then stop it, but keep the gdb server window open, then start a completely different game and then start a new connexion. This would mean the mapping could change, but GDB will be sent the previous one which is why I put it there because this code will only run when you open a sucessful connexion so we ensure the cached one is clear before we continue.

src/debugger/gdb-stub.c
aldelaro5: What should happen if there isn't enough in the buffer left to copy it? I understand the check that needs to happen, but not necessarily how to handle it because if that happens, then the buffer was too small to begin with.

aldelaro5:
Converted to draft because I found that ghidra has an issue with target descriptor in general, feel free to answer my 2 reviews comment for now, but I will hold off the merge until I can figure this out (since merging this will break ghidra and the issue is definitely not on our end because `info registers` still gives the same output)

aldelaro5:
I had to force push because I found that lldb can crash if you don't specify the layout of the registers explicility.

But in doing that, I also found that doing this allows us to not send dummy data on g packets because we explicility do not mention the presence of the fpr so both gdb and lldb can be happy.

src/debugger/gdb-stub.c
```diff
@@ -29,6 +29,30 @@ enum {
...
+                                "<reg name=\"r5\" bitsize=\"32\" type=\"uint32\"/>"
+                                "<reg name=\"r6\" bitsize=\"32\" type=\"uint32\"/>"
+                                "<reg name=\"r7\" bitsize=\"32\" type=\"uint32\"/>"
+                                "<reg name=\"r8\" bitsize=\"32\" type=\"uint32\"/>"
+                                "<reg name=\"r9\" bitsize=\"32\" type=\"uint32\"/>"
+                                "<reg name=\"r10\" bitsize=\"32\" type=\"uint32\"/>"
+                                "<reg name=\"r11\" bitsize=\"32\" type=\"uint32\"/>"
+                                "<reg name=\"r12\" bitsize=\"32\" type=\"uint32\"/>"
+                                "<reg name=\"sp\" bitsize=\"32\" type=\"data_ptr\"/>"
+                                "<reg name=\"lr\" bitsize=\"32\"/>"
```
endrift, line 50: This is sometimes used as a generic int, but sometimes not. Is there a strong reason for it to be labeled either way, or not at all?

src/debugger/gdb-stub.c
```diff
@@ -433,7 +447,55 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	unsigned length = 0;
+
+	unsigned index = 0;
+	for (index = 0; params[index] != ','; ++index) {
+		offset <<= 4;
+		offset |= _hex2int(&params[index], 1);
+	}
+
+	++index;
+	for (; index + 3 < strlen(params); ++index) {
```
endrift, line 464: You need to cache the result of strlen or else this loop becomes quadratic.

src/debugger/gdb-stub.c
```diff
@@ -433,7 +447,55 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+		length <<= 4;
+		length |= _hex2int(&params[index], 1);
+	}
+
+	if (length + 1 > GDB_STUB_MAX_LINE) {
+    length = GDB_STUB_MAX_LINE - 1;
```
endrift, line 470: Mixed indentation.

src/debugger/gdb-stub.c
endrift: If it's too small to begin with it should be truncated. Currently this will write off the end, which is even worse. While in theory neither should happen, it's better to fail-safe than fail-unsafe.

endrift (changes requested):
Still some nits and unresolved stuff.

src/debugger/gdb-stub.c
aldelaro5: These are just to tell the width of the registers. GDB's definition have the same things, but defining them explicitly allows for example to have GDB not try to check for the FPR because they aren't present.

So really, it's because that's how GDB does it vanilla.

src/debugger/gdb-stub.c
```diff
@@ -433,7 +447,62 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+
+	unsigned index = 0;
+	for (index = 0; params[index] != ','; ++index) {
+		offset <<= 4;
+		offset |= _hex2int(&params[index], 1);
+	}
+
+	++index;
+	int lengthStrParams = strlen(params);
+	for (; index + 3 < lengthStrParams; ++index) {
```
endrift, line 465: Nit: This name is very strange. I used paramsLength or similar in a previous recommendation, and I can recommend that again now.

src/debugger/gdb-stub.c
```diff
@@ -433,7 +447,62 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	}
+
+	memset(stub->outgoing, 0, GDB_STUB_MAX_LINE);
+	if (strlen(data) < length + offset) {
+		length = strlen(data) - offset;
+		stub->outgoing[0] = 'l';
+	} else {
+		stub->outgoing[0] = 'm';
+	}
+	strncpy(&stub->outgoing[1], &data[offset], length);
```
endrift, line 481: This is a very specific edge case, but: use `strlcpy` instead of `strncpy`; That will fix the edge case where the size of the buffer and the size of the data are equal--in the current case, the null terminator would get cropped out. strlcpy handles this properly.

src/debugger/gdb-stub.c
endrift: You appear to be correct. You could get away with just setting the first byte to 0, but ultimately it won't matter either way.

endrift (changes requested):
One extremely minor security thing that should be easily fixed, and one nit. Then it should be ready, as soon as I test it on my end.

src/debugger/gdb-stub.c
aldelaro5: done...but for some reason, I had to do + 1 on the length to make it not truncate the last character? that's odd.

endrift: That's a bad sign...means you're computing the maximum length wrong somewhere. I'll take a check tomorrow to see if I can figure it out.

src/debugger/gdb-stub.c
```diff
@@ -433,7 +448,62 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	index += strlen(memoryMap);
+	const struct mCoreMemoryBlock* blocks;
+	size_t nBlocks = stub->d.core->listMemoryBlocks(stub->d.core, &blocks);
+	size_t i;
+	for (i = 0; i < nBlocks; ++i) {
+		if (!(blocks[i].flags & mCORE_MEMORY_MAPPED)) {
+			continue;
+		}
+
+		char* type = blocks[i].flags & mCORE_MEMORY_WRITE ? "ram" : "rom";
```
endrift, line 497:
```suggestion
		const char* type = blocks[i].flags & mCORE_MEMORY_WRITE ? "ram" : "rom";
```

Also didn't you say that this breaks sw breakpoints in ROM?

src/debugger/gdb-stub.c
```diff
@@ -433,7 +448,62 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+		index += snprintf(&memoryMap[index], GDB_STUB_MAX_LINE - index, "<memory type=\"%s\" start=\"0x%08x\" length=\"0x%08x\"/>", type, blocks[i].start, blocks[i].size);
+	}
+	int amountLeft = GDB_STUB_MAX_LINE - index;
+	if (amountLeft < 14) {
+		strncpy(&memoryMap[index], "</memory-map>", amountLeft);
+	}
+	else {
+		strncpy(&memoryMap[index], "</memory-map>", 14);
+	}
```
endrift, line 506: This conditional is useless.

```suggestion
	strncpy(&memoryMap[index], "</memory-map>", amountLeft);
```

src/debugger/gdb-stub.c
```diff
@@ -433,7 +448,62 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	for (; index + 3 < paramsLength; ++index) {
+		length <<= 4;
+		length |= _hex2int(&params[index], 1);
+	}
+
+	if (length + 1 > GDB_STUB_MAX_LINE) {
+		length = GDB_STUB_MAX_LINE - 1;
+	}
+
+	memset(stub->outgoing, 0, GDB_STUB_MAX_LINE);
```
endrift, line 475: Since we overwrite the buffer here anyway, why is this needed?

src/debugger/gdb-stub.c
```diff
@@ -433,7 +448,62 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	++index;
+	unsigned int paramsLength = strlen(params);
+	for (; index + 3 < paramsLength; ++index) {
+		length <<= 4;
+		length |= _hex2int(&params[index], 1);
+	}
+
+	if (length + 1 > GDB_STUB_MAX_LINE) {
+		length = GDB_STUB_MAX_LINE - 1;
+	}
```
endrift, line 473: Doesn't this play poorly with the fact that there has to be a checksum at the end? Shouldn't you crop it shorter?

src/debugger/gdb-stub.c
```diff
@@ -433,7 +448,62 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+
+	memset(stub->outgoing, 0, GDB_STUB_MAX_LINE);
+	if (strlen(data) < length + offset) {
+		length = strlen(data) - offset;
+		stub->outgoing[0] = 'l';
+	} else {
+		stub->outgoing[0] = 'm';
+	}
+	strlcpy(&stub->outgoing[1], &data[offset], length + 1);
```
endrift, line 482: The + 1 being "needed" shows that you're calculating the length wrong.

endrift (changes requested):
Taking a closer look, I actually see...well, many issues.

src/debugger/gdb-stub.c
aldelaro5: It's more that gdb won't request sw because it is readonly memory.

The thing is that I really do not understand why it would matter if you consider that gdb can just guess the thumbness wrong and ask the wrong kind of breakpoint. So really, the question is more "why we allow sw in general?".

Though, by this logic, it could prevent to write memory in the rom which should be possible. This is why I was confused when you pointed out the rom shouldn't be listed as ram, CodaHighland more suggested to mark everything ram except the bios. When you declare rom here, you inform GDB that it's useless to try to write there, but considering it's a stub...it can write to stuff that would be read only right?

I would need clarity on what to do here tbh.

src/debugger/gdb-stub.c
aldelaro5: I debugged and I am actually not sure: GDB doesn't seem to count the null terminator as part of its requested length. So what strlcpy does is say GDB requests 651 bytes (which it does on my end), this will actually put 650 bytes then put the null terminator which makes it 651 length. I checked with a debugger and despite the fact I pass in a size of 651 (which is what GDB asked me), I end up with stub->outgoing being length 651 with a null terminator, but remember that the first byte doesn't count because it's m or l so I end up sending one byte short.

Am I correct? because the length is definitely calculated correctly at 651 when I get to this line, but this somehow copies 650 bytes instead of 651.

endrift: The discrepancy is the null terminator byte. `strlcpy` is supposed to take the size of the _buffer_ to ensure that after copying it is properly null terminated. If gdb requests a number of bytes and you use that directly then that means you're calculating the...number of bytes it's asking for, not the length of the buffer.

src/debugger/gdb-stub.c
```diff
@@ -433,7 +448,58 @@ static void _processQSupportedCommand(struct GDBStub* stub, const char* message)
...
+	index += strlen(memoryMap);
+	const struct mCoreMemoryBlock* blocks;
+	size_t nBlocks = stub->d.core->listMemoryBlocks(stub->d.core, &blocks);
+	size_t i;
+	for (i = 0; i < nBlocks; ++i) {
+		if (!(blocks[i].flags & mCORE_MEMORY_MAPPED)) {
+			continue;
+		}
+
+		const char* type = blocks[i].flags & mCORE_MEMORY_WRITE ? "ram" : "rom";
```
endrift, line 498: Since GDB can in fact patch memory on the ROM (but not the BIOS), the condition should be changed as such:

```suggestion
		const char* type = blocks[i].flags & (mCORE_MEMORY_WRITE | mCORE_MEMORY_WORM) ? "ram" : "rom";
```

PR #2400, opened 2021-12-30
GDB Stub: Allow to override GDB's handling of write watchpoints

aldelaro5:
This PR allows to override the GDB's handling of write watchpoints in 2 ways:

- Take control of the GDB's logic that detects if a write watchpoint changed or not as well as prevent a potentially unwanted step into
- The above, but with the addition to also break on nonmodifying watchpoints

The GDB logic part is done by hiding the truth to GDB: we send an S05 instead of a T05watch which forces GDB to stop and not do anything more than that. This can be desired because without this, it's not possible for GDB to know about mgba's savestates which means that say you do a savestate, break on a write watchpoint and then load that state causing the same write, GDB will not be able to detect this and ignore the request to stop (it will just request the inferior to continue). This logic is already properly handled in mgba so we just defer to it instead. Additionally, GDB assumes incorrectly that the break happens ON THE write instructions because normally, it would break there, step into and check if the value changed. mGBA actually breaks on the instruction AFTER so the end result is you end up 2 instructions after. This logic override simply allows us to have GDB not do that so we always end up on the instruction after the write.

For the nonmodifying writes, we simply change the internal type of watchpoint accordingly. This allows to bypass EITHER change logic which is not a feature that GDB supports, but is needed in a lot of reverse engineering cases.

The new options are introduced in the GDB server window, here's a preview of the new window:

Screenshot from 2022-02-09 20-28-02

### DIFF
--- a/src/platform/qt/GDBController.h
+++ b/src/platform/qt/GDBController.h
@@ -28,6 +28,7 @@ public:
 public slots:
 	void setPort(ushort port);
 	void setBindAddress(const Address&);
+	void setWatchpointsBehavior(int watchpointsBehaviorId);
 	void listen();
 
 signals:
@@ -41,6 +42,7 @@ private:
 
 	ushort m_port = 2345;
 	Address m_bindAddress;
+	enum GDBWatchpointsBehvaior m_watchpointsBehavior = GDB_WATCHPOINT_STANDARD_LOGIC;
 };
 
 }
--- a/src/platform/qt/GDBWindow.cpp
+++ b/src/platform/qt/GDBWindow.cpp
@@ -5,6 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 #include "GDBWindow.h"
 
+#include <QButtonGroup>
 #include <QGridLayout>
 #include <QGroupBox>
 #include <QLabel>
@@ -12,6 +13,7 @@
 #include <QMessageBox>
 #include <QPushButton>
 #include <QHBoxLayout>
+#include <QRadioButton>
 #include <QVBoxLayout>
 
 #include "GDBController.h"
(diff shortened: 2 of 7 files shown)


src/platform/qt/GDBController.h
```diff
@@ -28,6 +28,7 @@ Q_OBJECT
 public slots:
 	void setPort(ushort port);
 	void setBindAddress(const Address&);
+	void setWatchpointsBehavior(const int watchpointsBehaviorId);
```
endrift, line 31:
```suggestion
	void setWatchpointsBehavior(int watchpointsBehaviorId);
```

const does nothing useful here.

Though ideally this would be the actual enum, for signal/slot reasons I can see why you didn't do that. A lambda slot may be preferable.

src/platform/qt/GDBWindow.cpp
```diff
@@ -47,6 +49,28 @@ GDBWindow::GDBWindow(GDBController* controller, QWidget* parent)
...
+	m_watchpointsStandardRadio->setChecked(true);
+	watchpointsButtonGroup->addButton(m_watchpointsStandardRadio, 0);
+	watchpointsSettingsLayout->addWidget(m_watchpointsStandardRadio);
+
+	m_watchpointsOverrideLogicRadio = new QRadioButton("Override GDB's handling logic", watchpointsSettings);
+	watchpointsButtonGroup->addButton(m_watchpointsOverrideLogicRadio, 1);
+	watchpointsSettingsLayout->addWidget(m_watchpointsOverrideLogicRadio);
+
+	m_watchpointsOverrideLogicAnyWriteRadio = new QRadioButton("Override GDB's handling logic and\nbreak on nonmodifying writes", watchpointsSettings);
+	watchpointsButtonGroup->addButton(m_watchpointsOverrideLogicAnyWriteRadio, 2);
```
endrift, line 71: Can you not pass the enum values without casting?

endrift (commented):
Looks mostly good at the moment, but I'll want to take a closer look when it's not 4 AM and a headache is keeping me up.

src/platform/qt/GDBWindow.cpp
aldelaro5: I tried and it's not possible. Now I don't exactly know how you would do this with a lambda becasue I am connecting directly to the controller object meaning that if I don't cast on the controller side, I HAVE to cast on the window side, just to be sure, is the later what you would prefer? Cause then I do lambda slot and change the controller function to just be a normal setter.

endrift:
How about:
- Standard GDB
- Internal change detection
- Break on all writes

It's more concise, but I worry it won't get across the fact that they no longer show up as writepoint breaks to GDB and I'm not sure how to fix that without a bunch of extra text.

aldelaro5:
Changed the text and added the tr calls.

src/platform/qt/GDBWindow.cpp
ahigerd: I'd call this resolved now.

PR #2409, opened 2022-01-04
[RFC] CTest debugging registers

felixjones:
PR is mostly for discussion of this idea.
Motivation is using mGBA to write unit tests for GBA ROMs.

CTest is a very simple unit testing tool that's part of CMake https://gitlab.kitware.com/cmake/community/-/wikis/doc/ctest/Testing-With-CTest

The only requirement for CTest is that the executable returns with a code. This PR adds a crude way to call `exit` from a GBA ROM that quits mGBA with a given return code.
This is done by enabling mGBA debug, adding a CTest argument (see below), and then writing to ´REG_CTEST_EXIT´.

The `--ctest` argument is used to send string arguments to the GBA ROM for configuring a CTest.
`REG_CTEST_ARGC` contains the CTest argument counts, `REG_CTEST_ARGV` points to the CTest argv table (which all points within the buffer).

The horrible looking `parseArgsString` generates an `argv` table and packs as many args as possible from a given string into a buffer, returning the ´argc´.


endrift:
mGBA already has CTest tests, they just display the result visually and have the test suite executable (CInema) compare versus expected result.

If you want a generic way to pass arguments to a game, you shouldn't brand it CTEST. If you want CTest, you should explain why it needs to be a generic way to pass arguments to the game.

Further, I'm not fond of how this is implemented, but that could be beside the point depending on what you're trying to actually accomplish at the end of the day.

felixjones:
At the moment I'm having to execute mGBA on a timeout in a separate process, then REGEX match the output for my tests.
What I want to achieve is use mGBA for unit tests without having to do this timeout-process hack

The arguments parsing is to allow CTest parameters, right now I'm hijacking ROM saves to pass arguments as SRAM data, but whilst also very inflexible, this is a hack I can live with as it's the "wait for timeout" that is currently slowing down unit tests (and making them difficult to write).

How about this: a debug log flag that exits mGBA and returns either a zero or non-zero exit code?

endrift:
I'm pretty sure what you want is a custom frontend, not to hack something else onto regular-ass mGBA.

PR #2411, opened 2022-01-06
GBA code unit testing front-end

felixjones:
This is a testing front end specifically for unit testing GBA code, following the suggestion from @endrift in https://github.com/mgba-emu/mgba/pull/2409#issuecomment-1005319854

At the moment I have named this frontend "mtest", but this entire PR is WIP and open for feedback and suggestions (in particular, feature suggestions).
I will say that even this basic implementation has proven remarkably lean and perfect for my needs, it has sped up my unit testing by an order of magnitude and has made writing GBA unit tests an incredibly simple task.

Basic usage is: `mgba-mtest -S 0 -R r0 my_rom.gba`

- `-S SWI` sets what SWI to exit the test on. -S 0 will exit on SWI 0. You can probably pick an unused SWI if your test uses SWI 0.
- `-R REGISTER` sets what register to read when the exit condition is met (right now, that's only -S). r0-r15, sp, lr, pc and cpsr are valid strings for this.

A basic GBA test with tonc could look like:
```c
#include <tonc.h>

static void mtest_exit(int code) {
    swi_call(0);
}

int main() {
    mtest_exit(Div(9, 3) != 3);
}
```
With a crt0/syscalls.c that calls SWI on `exit()` the code required for a GBA test is even simpler than this as the return from main would signal the test end.

Using this with CTest could look something like this:
```cmake
enable_testing()
add_test(NAME exampleTest COMMAND mgba-mtest -S 0 -R r0 exampleRom.gba)
# -S 0 uses SWI 0 as the signal
# -R r0 uses r0 as the exit code
```

And because stdout is available us usual, a ctest can regex match any of the standard output.

As of this writing (2022-06-01) the current implementation in this PR by default has no way to actually complete if the -S argument is not used. Some feedback and ideas are required here (maybe default to exiting tests after 1 frame, so test writers are forced to be specific about their GBA tests?).

An aside: libmgba is remarkably easy to work with, I am incredibly impressed with how quickly it took me to get this working during the short moments I've had today to write this code.

(PR has some garbo commits, need squashing).


src/platform/test/mtest-main.c
```diff
@@ -0,0 +1,215 @@
...
+#define MTEST_USAGE \
+	"\nAdditional options:\n" \
+	"  -S SWI           Run until specified SWI call before exiting\n" \
+	"  -R REGISTER      Register to return as exit code\n" \
+
+ enum mRegisterExitCode {
+	 mRegister_NONE = -1,
+	 mRegister_GPRS = 0,
+	 mRegister_CPSR = 16
+ };
```
endrift, line 28:
```suggestion
 enum mRegisterExitCode {
	 mREGISTER_NONE = -1,
	 mREGISTER_GPRS = 0,
	 mREGISTER_CPSR = 16
 };
```

to match code style, though I have to wonder why anyone would put the exit code in the CPSR. Probably not worth support.

src/platform/test/mtest-main.c
```diff
@@ -0,0 +1,215 @@
...
+	mCoreConfigSetDefaultValue(&core->config, "idleOptimization", "remove");
+
+#ifdef M_CORE_GBA
+	if (core->platform(core) == mPLATFORM_GBA) {
+		((struct GBA*) core->board)->hardCrash = false;
+
+		_armSwi16 = ((struct GBA*) core->board)->cpu->irqh.swi16;
+		((struct GBA*) core->board)->cpu->irqh.swi16 = _mTestSwi16;
+		_armSwi32 = ((struct GBA*) core->board)->cpu->irqh.swi32;
+		((struct GBA*) core->board)->cpu->irqh.swi32 = _mTestSwi32;
```
endrift, line 96: Instead of hooking all SWIs and only using one, perhaps using SWI 03 (STOP) which already has a frontend-defined callback would be ideal here. See the `GBAStop` function in gba.c.

src/platform/test/mtest-main.c
```diff
@@ -0,0 +1,215 @@
...
+	freeArguments(&args);
+	mCoreConfigDeinit(&core->config);
+	core->deinit(core);
+
+	return cleanExit ? returnCode : 1;
+}
+
+static int _mTestRunloop(struct mCore* core, int exitSwiImmediate) {
+	do {
+		core->step(core);
```
endrift, line 149:
```suggestion
		core->runLoop(core);
```
Will be much faster, and an SWI call should always break out of the runloop, especially one to GBAStop.

endrift (changes requested):
Some quick minor stuff, but in general this seems like a better approach. Coming up with an exit condition for GB ROMs is probably a good idea too. Once the overall design idea in in place I can come back with a more detailed review.

src/platform/test/mtest-main.c
felixjones: I agree that GBAStop makes a perfect choice, however my own runtime uses SWI 0 to reset when exit is called. I've left in the option to override the SWI call, but if you absolutely disagree with this then I can adjust my runtime and remove this option here

LunarLambda:
The devkitARM C Runtime also calls SWI 0 when main() exits, so using SWI 0 + r0 (main return value) seems really useful to support. Maybe it's overkill to support arbitrary SWI and register combinations, but 0/r0 is reasonable.

Maybe a dedicated, unused SWI would be possible? I don't know how that would interact with using an official BIOS, though. Does mGBA still hook those?

Also, the original post mentions stdout being available. How does that work? Do you mean the log interface?

felixjones:
> Also, the original post mentions stdout being available. How does that work? Do you mean the log interface?

Yes, mGBA logging means zero extra work needed to be done for getting a unit test to write stdout, something that CTest specifically supports
So regardless of return code, with CTest I can regex for particular stdout patterns and pass/fail/skip on match/no match.

That being said, there's probably benefit to matching memory. Probably better to actually do those matches within this test platform, rather than stdout blobs of memory for regex'ing

LunarLambda:
That reminds, me I wanted to write an optimized `memcmp` for libseven.

That, or having matching `.iwram.bin` and similar files that contain the expected resulting memory image and similar.

something like `test.gba` -> `test.iwram.bin` or `./expected/iwram`

src/platform/test/mtest-main.c
```diff
@@ -0,0 +1,206 @@
...
+	UNUSED(signal);
+	_dispatchExiting = true;
+}
+
+static void _mTestSwi3Callback(struct mCore* core) {
+#ifdef M_CORE_GBA
+	_exitCode = ((struct GBA*) core->board)->cpu->regs.gprs[_returnCodeRegister];
+#endif
+	_dispatchExiting = true;
+}
```
endrift, line 164: Any reason this whole function isn't behind the ifdef?

src/platform/test/mtest-main.c
```diff
@@ -0,0 +1,206 @@
...
+		((struct GBA*) core->board)->hardCrash = false;
+
+		_exitSwiImmediate = mTestOpts.exitSwiImmediate;
+		_returnCodeRegister = mTestOpts.returnCodeRegister;
+
+		if (_exitSwiImmediate == 3) {
+			// Hook into SWI 3 (shutdown)
+			callbacks = calloc(1, sizeof(*callbacks));
+			callbacks->context = core;
+			callbacks->shutdown = (void(*)(void*)) _mTestSwi3Callback;
```
endrift, line 96: Any reason this is dynamically allocated? The struct isn't exactly very big.

endrift (approved):
Looks quite good, but I have two minor questions.

src/platform/test/mtest-main.c
felixjones: Probably in the mind-set of not consuming memory for GBA callbacks when running a test for GB core (which is dumb, we're hardly memory constrained).
Will remove.

felixjones:
@endrift the copyright notice on the source file has my name on it from when this was a branch intended to stay private. Is there a preferred copyright holder to attribute this to?

endrift:
I use my (legal) name: Jeffrey Pfau. But you are more than welcome to leave your name in there.

felixjones:
I noticed `addCoreCallbacks` copies the `struct mCoreCallbacks`, so I moved the allocation onto the stack for that.
I updated the copyright notice for the source file.

This is in a state where I am happy with the basic functionality and I think additional features (and more importantly, GB support) could come in as their own PRs.

src/platform/test/mtest-main.c
```diff
@@ -0,0 +1,202 @@
...
+	default:
+		return false;
+	}
+}
+
+static int _parseNamedRegister(const char* regStr) {
+	if (regStr[0] == 'r' || regStr[0] == 'R') {
+		++regStr;
+	}
+	return strtol(regStr, NULL, 10);
```
endrift, line 201: It would probably make sense to properly constrain this. Not that I'm too worried about some malicious actor here, but it is an obvious place to do so. Some error checking that it didn't fail to parse would probably be good too.

src/platform/test/CMakeLists.txt
```diff
@@ -46,3 +46,10 @@ if(BUILD_CINEMA)
 	add_test(cinema ${BINARY_NAME}-cinema -v)
 	install(TARGETS ${BINARY_NAME}-cinema DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT ${BINARY_NAME}-test)
 endif()
+
+if(BUILD_MTEST)
+	add_executable(${BINARY_NAME}-mtest ${CMAKE_CURRENT_SOURCE_DIR}/mtest-main.c)
```
endrift, line 51: Given that this would wind up with a binary named mgba-mtest, I wonder if the second m is redundant. Maybe a name like mgba-test-harness, or mgba-rom-tester? It's kind of verbose, but is pretty explicit about what it is as a result.

endrift (commented):
No real objections, but a few nits and questions about how things should ideally be done.

felixjones:
Renamed to mgba-rom-test so it is used like a verb on the command line.
Added constraints to both the -S and -R args. -S is from 0x00 to 0xFF, -R is from R0 to R15.

src/platform/test/rom-test-main.c
```diff
@@ -0,0 +1,217 @@
...
+
+#include <signal.h>
+
+#define ROM_TEST_OPTIONS "S:R:"
+#define ROM_TEST_USAGE \
+	"\nAdditional options:\n" \
+	"  -S SWI           Run until specified SWI call before exiting\n" \
+	"  -R REGISTER      General purpose register to return as exit code\n" \
+
+struct romTestOpts {
```
endrift, line 25:
```suggestion
struct RomTestOpts {
```

Sorry, I should have been more explicit; If m is being used as a prefix, the struct name is not initial caps, otherwise it is.

src/platform/test/rom-test-main.c
```diff
@@ -0,0 +1,217 @@
...
+		return _parseSwi(arg, &opts->exitSwiImmediate);
+	case 'R':
+		return _parseNamedRegister(arg, &opts->returnCodeRegister);
+	default:
+		return false;
+	}
+}
+
+static bool _parseSwi(const char* regStr, int* oSwi) {
+	long swi = strtol(regStr, NULL, 0);
```
endrift, line 198: It might be a good idea to make sure there isn't junk following the string, just for completeness sake. The errno alone is not sufficient given how strto* is specified.

src/platform/test/rom-test-main.c
```diff
@@ -0,0 +1,217 @@
...
+	*oSwi = swi;
+	return true;
+}
+
+static bool _parseNamedRegister(const char* regStr, unsigned int* oRegister) {
+	if (regStr[0] == 'r' || regStr[0] == 'R') {
+		++regStr;
+	}
+
+	unsigned long regId = strtoul(regStr, NULL, 10);
```
endrift, line 211: Ditto.

endrift (commented):
Two small things. Sorry this is taking so long.

src/platform/test/rom-test-main.c
```diff
@@ -0,0 +1,219 @@
...
+		((struct GBA*) core->board)->hardCrash = false;
+
+		_exitSwiImmediate = romTestOpts.exitSwiImmediate;
+		_returnCodeRegister = romTestOpts.returnCodeRegister;
+
+		if (_exitSwiImmediate == 3) {
+			// Hook into SWI 3 (shutdown)
+			struct mCoreCallbacks callbacks = {0};
+			callbacks.context = core;
+			callbacks.shutdown = (void(*)(void*)) _romTestSwi3Callback;
```
endrift, line 96: Why does this need to be cast? You can cast the void* to an mCore* inside of the callback implicitly (since it's C).

src/platform/test/rom-test-main.c
```diff
@@ -0,0 +1,219 @@
...
+		} else {
+			// Custom SWI hooks
+			_armSwi16 = ((struct GBA*) core->board)->cpu->irqh.swi16;
+			((struct GBA*) core->board)->cpu->irqh.swi16 = _romTestSwi16;
+			_armSwi32 = ((struct GBA*) core->board)->cpu->irqh.swi32;
+			((struct GBA*) core->board)->cpu->irqh.swi32 = _romTestSwi32;
+		}
+	}
+#endif
+
```
endrift, line 107: It might be good to add an "unsupported" for other here, otherwise it'll just run forever (or add another way to exit, such as a maximum frame count). Not a huge deal for the initial PR though. I can add this after it's merged.

endrift (approved):
Looks good, just some simple minor nits and two questions. It will be ready to merge afterwards.

src/platform/test/rom-test-main.c
felixjones: My IDE shouting at me to do explicit casts

felixjones:
I used an `#error` macro for the no core option, my rationale being that you have to explicitly compile rom-test, so if you're going that far then you should know that you need the GBA core (for now).

src/platform/test/rom-test-main.c
endrift: I...yes, sorry, I should have been more clear. You need the function type to match, but in the function itself you can just assign the void* argument to an mCore* variable implicitly.

endrift:
What I meant was that you can still compile it with GBA and GB support but if you pass a GB ROM to it you have to manually interrupt it. This isn't ideal.

PR #2418, opened 2022-01-10
MacOS fixes, kind of dodgy

kode54:
Please review this. Bless this mess. It needs some serious rewriting, but it mostly works. Unfortunately, due to the message pump mess, fast forward ends up getting stuck when running software mode, until cmd+tabbing away and back again.

### DIFF
--- a/include/mgba-util/platform/posix/threading.h
+++ b/include/mgba-util/platform/posix/threading.h
@@ -18,6 +18,13 @@ CXX_GUARD_START
 #include <OS.h>
 #endif
 
+#ifdef __APPLE__
+#include <sys/errno.h>
+#include <sys/time.h>
+#include <dispatch/dispatch.h>
+extern void (*__processEvents)();
+#endif
+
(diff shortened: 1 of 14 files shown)


endrift:
Can you tell me what this even does

kode54:
Sorry, I should have said what this does. This makes the GL display and GL enhancements functional under the 10.15-12.x SDKs, by serializing a bunch of GL calls through a message pump that gets processed on the main thread. It also changes a bunch of wait loops so they also process that message pump when waiting on the main thread. It's mostly a proof of concept.

endrift:
It sounds like a hacky workaround to defeat the intended design of this code instead of just...fixing the design so it doesn't break on macOS. Fixing it will be difficult but this is almost worse than no solution.

Also I just pushed a fix for the strlcpy thing that I'd committed-but-not-pushed-yet some when you first pointed it out so you can remove that.

kode54:
You're right, this is pretty bad, I guess.

PR #2452, opened 2022-02-18
Clean up rendering of logging grid checkboxes

ahigerd:
> <@endrift> @CodaHighland I just noticed that if you click on the checkbox...box in the logging settings grid, you get this selection frame for the text that isn't present...any idea how to get rid of it? 
> <@endrift> image
> <@endrift> you can see it's also taking up space that I don't need to on the right

Qt's item views aren't designed for this specific use case, but they provide accommodations for nonstandard appearances through QAbstractItemDelegate. This PR makes sure that the checkbox is centered within the cell and renders the keyboard focus in a more appropriate manner.

It also makes sure that the keyboard subfocus is updated appropriately, since clicking checkboxes in itemviews doesn't set the current item like clicking the text does.

### DIFF
--- a/src/platform/qt/CheckBoxDelegate.cpp
+++ b/src/platform/qt/CheckBoxDelegate.cpp
@@ -16,8 +16,7 @@ CheckBoxDelegate::CheckBoxDelegate(QObject* parent)
 	// initializers only
 }
 
-void CheckBoxDelegate::paint(QPainter* painter, const QStyleOptionViewItem& option, const QModelIndex& index) const
-{
+void CheckBoxDelegate::paint(QPainter* painter, const QStyleOptionViewItem& option, const QModelIndex& index) const {
 	QAbstractItemView* view = qobject_cast<QAbstractItemView*>(option.styleObject);
 	if (view && (index.flags() & Qt::ItemIsUserCheckable)) {
 		// Set up style options
--- a/src/platform/qt/CheckBoxDelegate.cpp
+++ b/src/platform/qt/CheckBoxDelegate.cpp
@@ -1,0 +1,50 @@
+/* Copyright (c) 2013-2022 Jeffrey Pfau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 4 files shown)


src/platform/qt/CheckBoxDelegate.cpp
```diff
@@ -0,0 +1,50 @@
...
+using namespace QGBA;
+
+CheckBoxDelegate::CheckBoxDelegate(QObject* parent)
+	: QStyledItemDelegate(parent)
+{
+	// initializers only
+}
+
+void CheckBoxDelegate::paint(QPainter* painter, const QStyleOptionViewItem& option, const QModelIndex& index) const
+{
```
endrift, line 20:
```suggestion
void CheckBoxDelegate::paint(QPainter* painter, const QStyleOptionViewItem& option, const QModelIndex& index) const {
```
Style nit

src/platform/qt/CheckBoxDelegate.h
```diff
@@ -0,0 +1,20 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#pragma once
+
+#include <QStyledItemDelegate>
+
+namespace QGBA {
+
+class CheckBoxDelegate : public QStyledItemDelegate
+{
```
endrift, line 13:
```suggestion
class CheckBoxDelegate : public QStyledItemDelegate {
```
Style nit

src/platform/qt/CheckBoxDelegate.h
```diff
@@ -0,0 +1,20 @@
...
+public:
```
endrift, line 14: No `Q_OBJECT`?

endrift (commented):
I'll take a closer look later, but there are some quick style nits first. Brace on a new line is only for constructors with initializers.

src/platform/qt/CheckBoxDelegate.h
ahigerd: Nope, don't need it. No signals, no slots, no qobject_cast, no QMetaType. The superclass's implementation takes care of everything in that regard.

Issue #2415, opened 2022-01-08
[Feature Request] Library mode always show filename


boomshroom:
I'm trying out the library mode, and I immediately noticed a potential issue. I have several copies of the same ROM, partly as backups, partly as multiple save files for the same game. When loading these into the library view, if they match a ROM in the database, then their filenames get overwritten with the game names. Since they have identical contents however, they all report as the same game making it easy to lose track of which ROM has which save file. I currently have 5 consecutive `Pokemon FireRed Version (USA)`s with no way to tell them apart. The fact that mGBA recognizes `.bak` files, but only some of them, as ROMs to load further worsens the issue.

![library](https://user-images.githubusercontent.com/628488/148629865-5578b255-b458-401e-bcdb-e43e3d417a44.png)

This is on version 0.9.3 on Linux.

PR #2481, opened 2022-03-27
Qt: Toggle to show filename or gamename in library mode

brunosxs:
The following PR implements the requested feature in #2415 .
It adds a new checkbox `Show Filename instead of ROM name in library view` in the Settings page, under Interface:

Screenshot_20220326_222632

Smoke tested in manjaro linux with QT 5.15.3
- Verified the option defaults to false.
- Verified the option saves and loads correctly
- Verified that the names change as expected depending on the state of the checkbox.

### DIFF
--- a/src/platform/qt/SettingsView.cpp
+++ b/src/platform/qt/SettingsView.cpp
@@ -488,6 +488,7 @@ void SettingsView::updateConfig() {
 	saveSetting("gba.forceGbp", m_ui.forceGbp);
 	saveSetting("vbaBugCompat", m_ui.vbaBugCompat);
 	saveSetting("updateAutoCheck", m_ui.updateAutoCheck);
+	saveSetting("showFilenameInLibrary", m_ui.showFilenameInLibrary);
 
 	if (m_ui.audioBufferSize->currentText().toInt() > 8192) {
 		m_ui.audioBufferSize->setCurrentText("8192");
@@ -710,6 +711,7 @@ void SettingsView::reloadConfig() {
 	loadSetting("gba.forceGbp", m_ui.forceGbp);
 	loadSetting("vbaBugCompat", m_ui.vbaBugCompat, true);
 	loadSetting("updateAutoCheck", m_ui.updateAutoCheck);
+	loadSetting("showFilenameInLibrary", m_ui.showFilenameInLibrary);
 
 	m_ui.libraryStyle->setCurrentIndex(loadSetting("libraryStyle").toInt());
 
--- a/src/platform/qt/SettingsView.ui
+++ b/src/platform/qt/SettingsView.ui
@@ -564,27 +564,34 @@
         </widget>
        </item>
        <item row="4" column="1">
+        <widget class="QCheckBox" name="showFilenameInLibrary">
+         <property name="text">
+          <string>Show filename instead of ROM name in library view</string>
+         </property>
+        </widget>
+       </item>
+       <item row="5" column="1">
         <widget class="QPushButton" name="clearCache">
          <property name="text">
           <string>Clear cache</string>
          </property>
         </widget>
        </item>
-       <item row="5" column="0" colspan="2">
+       <item row="6" column="0" colspan="2">
         <widget class="Line" name="line_8">
          <property name="orientation">
           <enum>Qt::Horizontal</enum>
          </property>
         </widget>
        </item>
-       <item row="6" column="1">
+       <item row="7" column="1">
         <widget class="QCheckBox" name="allowOpposingDirections">
          <property name="text">
           <string>Allow opposing input directions</string>
          </property>
         </widget>
        </item>
-       <item row="7" column="1">
+       <item row="8" column="1">
         <widget class="QCheckBox" name="suspendScreensaver">
          <property name="text">
           <string>Suspend screensaver</string>
@@ -594,14 +601,14 @@
          </property>
         </widget>
        </item>
-       <item row="8" column="0">
+       <item row="9" column="0">
         <widget class="QLabel" name="label_41">
          <property name="text">
           <string>When inactive:</string>
          </property>
         </widget>
        </item>
-       <item row="8" column="1">
+       <item row="9" column="1">
         <layout class="QHBoxLayout" name="horizontalLayout_21">
          <item>
           <widget class="QCheckBox" name="pauseOnFocusLost">
@@ -619,14 +626,14 @@
          </item>
         </layout>
        </item>
-       <item row="9" column="0">
+       <item row="10" column="0">
         <widget class="QLabel" name="label_42">
          <property name="text">
           <string>When minimized:</string>
          </property>
         </widget>
        </item>
-       <item row="9" column="1">
+       <item row="10" column="1">
         <layout class="QHBoxLayout" name="horizontalLayout_24">
          <item>
           <widget class="QCheckBox" name="pauseOnMinimize">
@@ -644,14 +651,14 @@
          </item>
         </layout>
        </item>
-       <item row="10" column="0" colspan="2">
+       <item row="11" column="0" colspan="2">
         <widget class="Line" name="line_17">
          <property name="orientation">
           <enum>Qt::Horizontal</enum>
          </property>
         </widget>
        </item>
-       <item row="11" column="1">
+       <item row="12" column="1">
         <widget class="QCheckBox" name="dynamicTitle">
          <property name="text">
           <string>Dynamically update window title</string>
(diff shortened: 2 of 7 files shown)


src/platform/qt/SettingsView.ui
```diff
@@ -97,7 +97,7 @@
    <item row="1" column="1">
     <widget class="QStackedWidget" name="stackedWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>1</number>
```
endrift, line 100: Please revert this line. It's a common mistake and I do it myself somewhat frequently.

src/platform/qt/SettingsView.ui
```diff
@@ -780,6 +780,16 @@
...
+       <item row="4" column="1">
+        <widget class="QCheckBox" name="showFilenameInLibrary">
+         <property name="toolTip">
+          <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;If the rom does not match mGBA's internal database, filename will be shown regardless if this option is unchecked.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+         </property>
+         <property name="text">
+          <string>Show filename instead of ROM name in library view</string>
+         </property>
+        </widget>
+       </item>
```
endrift, line 792: Please move these lines to above row 5 to ensure the tab order still works.

src/platform/qt/SettingsView.ui
```diff
@@ -780,6 +780,16 @@
          </property>
         </widget>
        </item>
+       <item row="4" column="1">
+        <widget class="QCheckBox" name="showFilenameInLibrary">
+         <property name="toolTip">
+          <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;If the rom does not match mGBA's internal database, filename will be shown regardless if this option is unchecked.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+         </property>
```
endrift, line 787: mGBA doesn't have tooltips for stuff yet, so let's keep that consistent for now at least.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -37,6 +37,13 @@ void AbstractGameList::updateEntry(const LibraryEntry& item) {
 void AbstractGameList::removeEntry(const QString& item) {
 	removeEntries({item});
 }
+void AbstractGameList::setShowFilenameInLibrary(bool newValue) {
```
endrift, line 40:
```suggestion
void AbstractGameList::setShowFilename(bool showFilename) {
```
Let's make it a little less verbose, and the parameter a bit more descriptive.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -37,6 +37,13 @@ void AbstractGameList::updateEntry(const LibraryEntry& item) {
...
+	if (showFilenameInLibrary == newValue) {
+		return;
+	}
```
endrift, line 43: Seems unneeded if this doesn't trigger anything other than an assignment.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -210,3 +217,16 @@ void LibraryController::loadDirectory(const QString& dir, bool recursive) {
 	mLibraryLoadDirectory(library.get(), dir.toUtf8().constData(), recursive);
 	m_libraryJob.testAndSetOrdered(libraryJob, -1);
 }
+void LibraryController::setShowFilenameInLibrary(bool newValue) {
+	if (newValue == showFilenameInLibrary) {
+		return;
+	}
+	showFilenameInLibrary = newValue;
+	AbstractGameList* newCurrentList = nullptr;
+	if (m_libraryGrid.get() != nullptr) {
```
endrift, line 226:
```suggestion
	if (m_libraryGrid) {
```

src/platform/qt/library/LibraryController.h
```diff
@@ -112,6 +117,7 @@ private slots:
 
 	std::unique_ptr<LibraryGrid> m_libraryGrid;
 	std::unique_ptr<LibraryTree> m_libraryTree;
+	bool showFilenameInLibrary = false;
```
endrift, line 120:
```suggestion
	bool m_showFilename = false;
```

src/platform/qt/library/LibraryController.h
```diff
@@ -65,8 +65,12 @@ class AbstractGameList {
 	virtual void addEntry(const LibraryEntry&);
 	virtual void updateEntry(const LibraryEntry&);
 	virtual void removeEntry(const QString&);
+	virtual void setShowFilenameInLibrary(bool newValue);
 
 	virtual QWidget* widget() = 0;
+protected:
+	bool showFilenameInLibrary = false;
```
endrift, line 72:
```suggestion
	bool m_showFilename = false;
```

Is this actually used anywhere? And why protected?

endrift (changes requested):
You don't appear to have updated LibraryTree. Moreover, the function names are more verbose than needed, and the private variable names don't match the style.

src/platform/qt/SettingsView.ui
brunosxs: I saw it didn't had but I added it because I felt necessary mainly because this particular option needs some clarification. And also to start a discussion on how to give more feedback to the user.

The behavior isn't clear to the end-user. When he sees "Name" in the library, convention says he expects this:
image
So, adding an option that explicitly changes this behavior, yet does not change under certain circunstances (like not having the database) and not making it clear to the user will lead to confusion. That is why I felt necessary to add a tooltip to make it clear that, if the rom cannot be found in the database, it will default to the filename regardless of the checkbox.

src/platform/qt/library/LibraryTree.cpp
```diff
@@ -140,7 +140,7 @@ void LibraryTree::updateEntry(const LibraryEntry& item) {
 	m_entries[item.fullpath] = item;
 
 	LibraryTreeItem* i = static_cast<LibraryTreeItem*>(m_items.value(item.fullpath));
-	i->setText(COL_NAME, item.displayTitle());
+	i->setText(COL_NAME, m_showFilename? item.filename : item.displayTitle());
```
endrift, line 143:
```suggestion
	i->setText(COL_NAME, m_showFilename ? item.filename : item.displayTitle());
```

Minor style nit

src/platform/qt/library/LibraryGrid.cpp
```diff
@@ -83,7 +87,11 @@ void LibraryGrid::updateEntries(const QList<LibraryEntry>& items) {
 
 void LibraryGrid::updateEntry(const LibraryEntry& item) {
 	QListWidgetItem* i = m_items.value(item.fullpath);
-	i->setText(item.displayTitle());
```
endrift, line 86: You can do the same ternary operation you do in LibraryGrid here too.

src/platform/qt/library/LibraryController.h
```diff
@@ -65,8 +65,12 @@ class AbstractGameList {
 	virtual void addEntry(const LibraryEntry&);
 	virtual void updateEntry(const LibraryEntry&);
 	virtual void removeEntry(const QString&);
+	virtual void setShowFilename(bool showFilename);
 
 	virtual QWidget* widget() = 0;
+protected:
+	bool m_showFilename = false;
+
```
endrift, line 73:
```suggestion

protected:
	bool m_showFilename = false;
```

Spacing nit.

src/platform/qt/SettingsView.cpp
```diff
@@ -710,6 +711,7 @@ void SettingsView::reloadConfig() {
 	loadSetting("gba.forceGbp", m_ui.forceGbp);
 	loadSetting("vbaBugCompat", m_ui.vbaBugCompat, true);
 	loadSetting("updateAutoCheck", m_ui.updateAutoCheck);
+	loadSetting("showFilenameInLibrary", m_ui.showFilenameInLibrary, false);
```
endrift, line 714:
```suggestion
	loadSetting("showFilenameInLibrary", m_ui.showFilenameInLibrary);
```
`false` is the default so you don't need to pass it.

endrift (changes requested):
Except for some minor nitpicks, the implementation looks fine. I just can't help but wonder if maybe an extra column would work better. The whole thing needs to be heavily revamped anyway though (to actually use a QAbstractItemModel instead of this...slow approach), so it shouldn't matter too much before then.

endrift (approved):

PR #2539, opened 2022-05-31
Update new translated Simplified Chinese entry

shinyoyo:
Review and updated language file.


endrift:
In the future, please use the weblate page (linked in the readme) to avoid accidentally duplicating work

shinyoyo:
OK，I'll do it

PR #2547, opened 2022-06-09
Scripting: Change behaviour related to absolute and relative paths.

nuive:
I added the following changes related to file management on Lua Scripts:

1. Deleted the code which removes the absolute path from the filename. This is because LUA saves which file is running each function, and with that you can do the following in order to get the folder containing the file which is currently running:
path_of_running_file = debug.getinfo(1, "S").source:sub(2):match("(.*[/\\])")

And if the emulator only gives the LUA Engine the filename, this code returns nil.
I looked into other ways to get the path of running script, but all I've found is related to DLL modules, and this change simplifies this task.

2. Added some lines extracted from VBA-Rerecording and modified them in order to make them compatible with all OS (not Windows only). change the working folder to the folder containing the running LUA Script. This allows the LUA engine to look into the running script's folder for other scripts or modules if the current script requires them. Without this change, all .lua files must be stored in mGBA's lua folder, and the DLLs in mGBA's own folder.

### DIFF
--- a/src/script/engines/lua.c
+++ b/src/script/engines/lua.c
@@ -476,24 +476,24 @@ bool _luaLoad(struct mScriptEngineContext* ctx, const char* filename, struct VFi
 		free(luaContext->lastError);
 		luaContext->lastError = NULL;
 	}
-	char name[80];
 	if (filename) {
+		char name[strlen(filename) + 2];
 		if (*filename == '*') {
 			snprintf(name, sizeof(name), "=%s", filename + 1);
 		} else {
-			const char* lastSlash = strrchr(filename, '/');
-			const char* lastBackslash = strrchr(filename, '\\');
-			if (lastSlash && lastBackslash) {
-				if (lastSlash > lastBackslash) {
-					filename = lastSlash + 1;
-				} else {
-					filename = lastBackslash + 1;
-				}
-			} else if (lastSlash) {
-				filename = lastSlash + 1;
-			} else if (lastBackslash) {
-				filename = lastBackslash + 1;
+			char  dir[strlen(filename)];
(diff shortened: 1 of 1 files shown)


endrift:
You cannot take code from VBA-rr and put it in mGBA. The licenses are not compatible.

endrift:
Should be handled (more) cleanly by https://github.com/mgba-emu/mgba/commit/6e196748d88b830747e930c65365c9452654ab23

nuive:
Sorry, but it doesn't fix any of the two problems.

1. Deleted the code which removes the absolute path from the filename.
I supose you do this for some reason. As i said on the first message, this breaks the possibility of getting the folder where the script is run from. In my case, I have a LUA script which depends on a "data" folder located in the same folder that the script. I supose a workaround for that would be:
-Create another .lua file called path.lua with the code
path_of_running_file = debug.getinfo(1, "S").source:sub(2):match("(.*[/\])")
-Add the following to the main script
require "path".
-Then it would be solved, because this path problem is only with the script file you choose to run, not with any other file this script needs.

2. Change the working folder to the folder containing the running LUA Script.
The solution of adding the folder to the DLL folders is good, but it doesn't work for that, because this is related to where LUA looks for modules, not MGBA. LUA has a set of folders where it can look for DLLs and one of them is . (current folder) so the only way I could find was making LUA run from the script folder.

endrift:
That commit definitely does stop stripping out the absolute path...I have no idea how you're checking it if you thinks it doesn't.

I will need to look into how Lua (it's not spelled LUA, it's not short for anything) looks for DLLs, but changing the working directory is not acceptable.

PR #2565, opened 2022-06-22
clear buffer list in UI before deleting buffers

ahigerd:
Resetting the scripting state crashes mGBA. This PR does the minimum necessary to make it not do that by clearing out the list of buffers in the UI before deleting any data structures.

This PR is mutually exclusive with https://github.com/mgba-emu/mgba/pull/2566 as both solve the same issue in different ways.


endrift:
Simpler though this may be, I think I prefer the model approach for the long run.

PR #2566, opened 2022-06-22
refactor the script buffer list into a model

ahigerd:
Resetting the scripting state crashes mGBA. This PR refactors the way the UI keeps track of the list of buffers by using a list model instead of a QListWidget, enabling the use of standard Qt MVC signals to avoid conflicts.

This PR is mutually exclusive with https://github.com/mgba-emu/mgba/pull/2565 as both solve the same issue in different ways.

### DIFF
--- a/src/platform/qt/ScriptingController.cpp
+++ b/src/platform/qt/ScriptingController.cpp
@@ -7,7 +7,6 @@
 
 #include "CoreController.h"
 #include "ScriptingTextBuffer.h"
-class QTextDocument;
 
 using namespace QGBA;
 
--- a/src/platform/qt/ScriptingController.cpp
+++ b/src/platform/qt/ScriptingController.cpp
@@ -7,11 +7,12 @@
 
 #include "CoreController.h"
 #include "ScriptingTextBuffer.h"
+#include "ScriptingTextBufferModel.h"
 
 using namespace QGBA;
 
 ScriptingController::ScriptingController(QObject* parent)
-	: QAbstractListModel(parent)
+	: QObject(parent)
 {
 	m_logger.p = this;
 	m_logger.log = [](mLogger* log, int, enum mLogLevel level, const char* format, va_list args) {
@@ -32,6 +33,9 @@ ScriptingController::ScriptingController(QObject* parent)
 			break;
 		}
 	};
+
+	m_bufferModel = new ScriptingTextBufferModel(this);
+	QObject::connect(m_bufferModel, &ScriptingTextBufferModel::textBufferCreated, this, &ScriptingController::textBufferCreated);
 
 	init();
 }
@@ -86,14 +90,8 @@ void ScriptingController::clearController() {
 }
 
 void ScriptingController::reset() {
-	beginResetModel();
-	QList<ScriptingTextBuffer*> toDelete = m_buffers;
-	m_buffers.clear();
-	endResetModel();
 	CoreController::Interrupter interrupter(m_controller);
-	for (ScriptingTextBuffer* buffer : toDelete) {
-		delete buffer;
-	}
+	m_bufferModel->reset();
 	mScriptContextDetachCore(&m_scriptContext);
 	mScriptContextDeinit(&m_scriptContext);
 	m_engines.clear();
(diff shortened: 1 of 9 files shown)


src/platform/qt/ScriptingController.cpp
```diff
@@ -7,11 +7,12 @@
 
 #include "CoreController.h"
 #include "ScriptingTextBuffer.h"
+class QTextDocument;
 
 using namespace QGBA;
 
 ScriptingController::ScriptingController(QObject* parent)
-	: QObject(parent)
+	: QAbstractListModel(parent)
```
endrift, line 15: I'm...really not sure how I feel about making the ScriptingController inherit from QAbstractListModel. But at the same time, splitting out a separate class also feels...strange.

endrift (changes requested):
Needs a bit of discussion/bikeshedding + 1 easy cleanup nit

src/platform/qt/ScriptingView.cpp
```diff
@@ -9,6 +9,7 @@
 #include "ConfigController.h"
 #include "ScriptingController.h"
 #include "ScriptingTextBuffer.h"
+#include <QtDebug>
```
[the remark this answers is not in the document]
ahigerd: Oops. I thought I got all of those. Thanks.

src/platform/qt/ScriptingController.cpp
```diff
@@ -7,11 +7,12 @@
 
 #include "CoreController.h"
 #include "ScriptingTextBuffer.h"
+class QTextDocument;
```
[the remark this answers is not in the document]
ahigerd: ... er. *frown* I meant to put that in the .h file (where there are no using directives), so that isn't even where I intended to put it. Oops.

src/platform/qt/ScriptingController.cpp
```diff
@@ -131,3 +138,32 @@ void ScriptingController::init() {
 		m_activeEngine = *m_engines.begin();
 	}
 }
+
+void ScriptingController::bufferNameChanged(const QString&) {
+	ScriptingTextBuffer* buffer = qobject_cast<ScriptingTextBuffer*>(sender());
```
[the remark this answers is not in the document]
ahigerd: There's `QSignalMapper` but it just wraps `sender()` for you. This is just how it's done unless you want to add a pointer to the sender to the signal itself.

It's not a fragile cast, thankfully. sender() is either `nullptr` or it's a QObject. And if you have a QObject, then `qobject_cast` is a fast, safe upcast that returns null if it's not the desired type.

src/platform/qt/ScriptingController.cpp
ahigerd: The same thought occurred to me while I was writing it. It felt a little bit strange from a conceptual level, but the alternative is a thin little wrapper class that mostly just proxies signals, which seemed to be a bit much. I'll leave it up to you if that's something you want to change.

src/platform/qt/ScriptingView.cpp
```diff
@@ -57,27 +60,20 @@ void ScriptingView::load() {
...
-		m_textBuffers.removeAll(buffer);
-		delete item;
-	});
-	m_ui.buffers->addItem(item);
-	m_ui.buffers->setCurrentItem(item);
+void ScriptingView::controllerReset() {
+	selectBuffer(QModelIndex());
+}
+
+void ScriptingView::bufferAdded(const QModelIndex&, int row, int) {
```
ahigerd, line 67: I really wish there was a better way to do _this_. This wrapper feels silly. But as far as I can tell, this is the best it can get.

endrift: Might I recommend a lambda?

ahigerd: I was on the fence about a lambda when I wrote it (I prefer avoiding lambdas that aren't closing over local variables, since you could have used a compile-time function instead of a run-time object), but if you're recommending it, then I think that tips the scales.

src/platform/qt/ScriptingController.cpp
endrift: I thought about this more, and I really don't like it. The model represents something arbitrarily associated with scripting, the text buffers, not scripts themselves. The moment something else comes up that might want to be stuck in a list, this won't work. For forward-thinking purposes, I think it needs to be refactored into a separate class.

ahigerd: That's entirely fair, and it does satisfy the uncertainty I had. Done.

src/platform/qt/ScriptingController.h
```diff
@@ -6,7 +6,7 @@
 #pragma once
 
 #include <QHash>
-#include <QObject>
+#include <QAbstractListModel>
```
endrift, line 9: More cleanup needed

src/platform/qt/ScriptingTextBufferModel.h
```diff
@@ -0,0 +1,45 @@
...
+	void textBufferCreated(ScriptingTextBuffer*);
+
+public slots:
+	void reset();
+
+private slots:
+	void bufferNameChanged(const QString&);
+
+private:
+	static mScriptTextBuffer* createTextBuffer(void* context);
```
endrift, line 40: I think I'd prefer if this is public instead of using friend classes. Since this isn't a library, we don't have to worry as much about API misuse.

endrift (commented):
Seems much better this time around, just two small nits.

src/platform/qt/ScriptingTextBufferModel.h
ahigerd: Picked a slightly different solution that should still satisfy.

PR #2570, opened 2022-06-28
rewrite Qt library frontend

ahigerd:
I've ripped out the existing implementation for the Qt library frontend and replaced it with a QAbstractItemModel. This should improve performance significantly. I've added some additional functionality along the way, like some reasonable behavior on window resizing and storing view configuration in the settings.

The grid view isn't implemented yet, obviously (but it wasn't available for selection in the settings, either), and I haven't removed LibraryGrid from the repository for future reference.


src/platform/qt/library/LibraryController.cpp
```diff
@@ -8,8 +8,13 @@
 
 #include "ConfigController.h"
 #include "GBAApp.h"
-#include "LibraryGrid.h"
-#include "LibraryTree.h"
+#include "LibraryModel.h"
+
+#include <QTreeView>
```
endrift, line 13: Nit: Please keep these in alphabetical order

src/platform/qt/library/LibraryController.cpp
```diff
@@ -37,13 +42,11 @@ void AbstractGameList::updateEntry(const LibraryEntry& item) {
...
 	removeEntries({item});
 }
-void AbstractGameList::setShowFilename(bool showFilename) {
-	m_showFilename = showFilename;
-}
 
 LibraryController::LibraryController(QWidget* parent, const QString& path, ConfigController* config)
 	: QStackedWidget(parent)
 	, m_config(config)
+	, m_currentView(nullptr)
```
endrift, line 49: Nit: I'd prefer initialization with a constant to be done in the header.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -55,14 +58,52 @@ LibraryController::LibraryController(QWidget* parent, const QString& path, Confi
...
 	mLibraryAttachGameDB(m_library.get(), GBAApp::app()->gameDB());
 
-	m_libraryTree = std::make_unique<LibraryTree>(this);
-	addWidget(m_libraryTree->widget());
-
-	m_libraryGrid = std::make_unique<LibraryGrid>(this);
-	addWidget(m_libraryGrid->widget());
+	m_libraryModel = new LibraryModel(this);
+
+	addWidget(m_treeView = new QTreeView(this));
```
endrift, line 63: Nit: No assignment in expressions.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -55,14 +58,52 @@ LibraryController::LibraryController(QWidget* parent, const QString& path, Confi
...
+	m_treeModel = new QSortFilterProxyModel(this);
+	m_treeModel->setSourceModel(m_libraryModel);
+	m_treeModel->setSortRole(Qt::EditRole);
+	m_treeView->setModel(m_treeModel);
+	m_treeView->setSortingEnabled(true);
+	m_treeView->setAlternatingRowColors(true);
+
+	addWidget(m_listView = new QListView(this));
```
endrift, line 71: Ditto

src/platform/qt/library/LibraryController.cpp
```diff
@@ -55,14 +58,52 @@ LibraryController::LibraryController(QWidget* parent, const QString& path, Confi
...
+	m_listModel = new QSortFilterProxyModel(this);
+	m_listModel->setSourceModel(m_libraryModel);
+	m_listModel->setSortRole(Qt::EditRole);
+	m_listView->setModel(m_listModel);
+
+	QObject::connect(m_treeView, &QAbstractItemView::activated, this, &LibraryController::startGame);
+	QObject::connect(m_listView, &QAbstractItemView::activated, this, &LibraryController::startGame);
+	QObject::connect(m_treeView->header(), &QHeaderView::sortIndicatorChanged, this, &LibraryController::sortChanged);
+
+	QTimer* expandThrottle = new QTimer(this);
```
endrift, line 81: This is...sort of a weird way to do this. I suppose it works, but...

Anyway, in general I'd make something like this a member variable, even if I don't access it after this function, just so that it lives in the same allocation. Not sure that makes sense in the case of d-pointer classes though.

src/platform/qt/library/LibraryController.h
```diff
@@ -15,11 +15,17 @@
 
 #include <mgba/core/library.h>
 
+class QTreeView;
```
endrift, line 18: Nit: Alphabetical order

src/platform/qt/library/LibraryController.h
```diff
@@ -15,11 +15,17 @@
...
+class QListView;
+class QAbstractItemView;
+class QSortFilterProxyModel;
+
 namespace QGBA {
 
 // Predefinitions
 class LibraryGrid;
 class LibraryTree;
+class LibraryModel;
```
endrift, line 28: Nit: These weren't in alphabetical order before, but you can fix that since you're here. Also, that comment seems useless.

src/platform/qt/library/LibraryController.h
```diff
@@ -47,16 +55,16 @@ struct LibraryEntry {
 	size_t filesize;
 	uint32_t crc32;
 
+	LibraryEntry& operator=(const LibraryEntry&) = default;
+	LibraryEntry& operator=(LibraryEntry&&) = default;
 	bool operator==(const LibraryEntry& other) const { return other.fullpath == fullpath; }
+
+	static inline uint64_t checkHash(size_t filesize, uint32_t crc32) { return (uint64_t(filesize) << 32) | crc32; }
```
endrift, line 62: Though this seems potentially a bit of a weird way to add additional integrity to the hash, I suppose it's not the worst. Especially given how bad CRC32 is. (I intend to add more checksums down the line.) However, you should probably at least saturate the file size to 32 bits, just on the offhand chance someone somehow adds a 4 GB file (even though no ROMs are that big). A combination of clz and popcount is probably more than sufficient, though admittedly slower. There are utility functions for both in mgba-util/math.h--there might not be a 64-bit version of them atm but I believe there is on the medusa branch that can be backported. A basic single-round LCG is probably also sufficient.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,422 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include "LibraryModel.h"
+
+#include "utils.h"
+
+#include <QSortFilterProxyModel>
+#include <QItemSelectionModel>
+#include <QApplication>
+#include <QDir>
+#include <algorithm>
```
endrift, line 15: Nit: Newline between Qt and stl includes, and alphabetical within each group

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,422 @@
...
+{
+	if (m_treeMode == tree) {
+		return;
+	}
+	beginResetModel();
+	m_treeMode = tree;
+	endResetModel();
+}
+
+bool LibraryModel::showFilename() const
```
endrift, line 103: Nit: No newline after the function signature (constructors are an exception).

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,422 @@
...
+{
+	return m_showFilename;
+}
+
+void LibraryModel::setShowFilename(bool show)
+{
```
endrift, line 109: Ditto

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,422 @@
...
+	if (items.isEmpty()) {
+		return;
+	} else if (m_treeMode) {
+		addEntriesTree(items);
+	} else {
+		addEntriesList(items);
+	}
+}
+
+void LibraryModel::addEntryInternal(const LibraryEntry& item)
```
endrift, line 148: Ditto

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,422 @@
...
+#include <QItemSelectionModel>
+#include <QApplication>
+#include <QDir>
+#include <algorithm>
+
+using namespace QGBA;
+
+namespace QGBA {
+
+struct SpanSet {
```
endrift, line 21: I wonder if some version of this (perhaps with templates) might be useful in qt/utils.h. We can think about that later, though.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -213,16 +266,46 @@ void LibraryController::loadDirectory(const QString& dir, bool recursive) {
...
+
+void LibraryController::resizeTreeView(bool expand) {
+	if (expand) {
+		m_treeView->expandAll();
 	}
-	if (m_libraryTree) {
-		m_libraryTree->setShowFilename(m_showFilename);
+
+	int viewportWidth = m_treeView->viewport()->width();
+	int totalWidth = m_treeView->header()->sectionSizeHint(LibraryModel::COL_CRC32);
```
endrift, line 289: We should probably add a COL_MAX, since there may be more columns later on.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,422 @@
...
+			entry = m_pathIndex[path][index.row()];
+		} else if (!index.parent().isValid() && index.row() < m_games.size()) {
+			entry = &m_games[index.row()];
+		}
+		if (entry) {
+			if (role == FullPathRole) {
+				return entry->fullpath;
+			}
+			switch (index.column()) {
+				case COL_NAME: return m_showFilename ? entry->filename : entry->displayTitle();
```
endrift, line 361: Nit: Newline between case labels and statements.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,422 @@
...
+	} else if (role == Qt::TextAlignmentRole) {
+		return index.column() == COL_SIZE ? (int)(Qt::AlignTrailing | Qt::AlignVCenter) : (int)(Qt::AlignLeading | Qt::AlignVCenter);
+	}
+	return QVariant();
+}
+
+QVariant LibraryModel::headerData(int section, Qt::Orientation orientation, int role) const {
+	if (orientation == Qt::Horizontal && role == Qt::DisplayRole) {
+		switch (section) {
+			case COL_NAME: return QApplication::translate("LibraryTree", "Name", nullptr);
```
endrift, line 377: Ditto

src/platform/qt/library/LibraryModel.h
```diff
@@ -0,0 +1,78 @@
+/* Copyright (c) 2014-2017 waddlesplash
```
endrift, line 1: Unless this has a lot in common with waddlesplash's previous version, it might make sense to remove this copyright. @waddlesplash, any opinion? Also, if you're not going to put yourself in the copyright, please put me.

endrift (changes requested):
Mostly just nits (though a lot of them), and a few bikeshedding questions. I trust your implementation skills for Qt models, so I didn't look over the implementation in too much depth.

src/platform/qt/library/LibraryModel.h
waddlesplash: I only glanced at the code, but it seems superficially similar at best, so I don't think I have much of a copyright claim to it indeed.

src/platform/qt/library/LibraryModel.cpp
ahigerd: Oops. I _knew_ that was the style guide, but I follow the Linux kernel style guide in basically everything else so I forgot ^^()

src/platform/qt/library/LibraryController.cpp
ahigerd: Agreed. I had toyed with the idea and then got distracted.

src/platform/qt/CMakeLists.txt
```diff
@@ -481,3 +480,23 @@ if(DISTBUILD AND NOT APPLE)
 		add_custom_command(TARGET ${BINARY_NAME}-qt POST_BUILD COMMAND "${STRIP}" "$<TARGET_FILE:${BINARY_NAME}-qt>")
 	endif()
 endif()
+
+if(BUILD_SUITE)
+	enable_testing()
+	find_package(${QT}Test)
+	if(${QT}Test_FOUND)
+		set(TEST_SRC
+			test/main.cpp
```
endrift, line 489: Nit: Alphabetical order

src/platform/qt/library/LibraryController.cpp
```diff
@@ -55,14 +57,53 @@ LibraryController::LibraryController(QWidget* parent, const QString& path, Confi
...
+	} else {
+		m_currentStyle = LibraryStyle::STYLE_TREE;
+	}
+	setViewStyle(libraryStyle);
 
-	m_currentStyle = LibraryStyle::STYLE_TREE; // Make sure setViewStyle does something
-	setViewStyle(LibraryStyle::STYLE_LIST);
+	QVariant librarySort = m_config->getQtOption("librarySort");
+	QVariant librarySortOrder = m_config->getQtOption("librarySortOrder");
+	if (librarySort.isNull()) {
```
endrift, line 99: Should check if it can be converted to int if not null

src/platform/qt/library/LibraryController.cpp
```diff
@@ -55,14 +57,53 @@ LibraryController::LibraryController(QWidget* parent, const QString& path, Confi
...
+		librarySort = 0;
+	}
+	if (librarySortOrder.isNull()) {
```
endrift, line 102: Should check if it can be converted to Qt::SortOrder if not null

src/platform/qt/library/LibraryController.h
```diff
@@ -47,16 +53,18 @@ struct LibraryEntry {
 	size_t filesize;
 	uint32_t crc32;
 
+	LibraryEntry& operator=(const LibraryEntry&) = default;
+	LibraryEntry& operator=(LibraryEntry&&) = default;
 	bool operator==(const LibraryEntry& other) const { return other.fullpath == fullpath; }
+
+	static inline uint64_t checkHash(size_t filesize, uint32_t crc32) {
```
endrift, line 60: You might want to add a const mLibraryEntry* overload for this and make the one that takes the specific types private so it can be changed easily in the future, should we decide the current inputs aren't enough.

src/platform/qt/test/library.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+using namespace QGBA;
+
+class LibraryModelTest : public QObject {
+Q_OBJECT
+
+private:
+	LibraryModel* model = nullptr;
+
+	int findGbaRow() {
```
endrift, line 28: _Extremely_ minor nit, but if you have to push anything might as well fix it: `findGBARow` is preferred here.

src/platform/qt/test/library.cpp
```diff
@@ -0,0 +1,200 @@
...
+				}
+			}
+		}
+		return QModelIndex();
+	}
+
+private slots:
+	void init() {
+		model = new LibraryModel();
+#if QT_VERSION >= QT_VERSION_CHECK(5, 11, 0)
```
endrift, line 118: Does this class work in Qt 5.10 or lower?

endrift (changes requested):
Some minor things since I've still glossed over the actual implementation fine details. I'll need to check how it works on my end before I can give approval anyway.

src/platform/qt/test/library.cpp
ahigerd: Noted. (That disagrees with the Qt style guide, which uses e.g. `QHttpHeader`, so this was going to be a clash of style guides one way or the other.)

src/platform/qt/CMakeLists.txt
ahigerd: How does sort order apply to files vs folders? Files first, folders first, sort strictly alphabetically?

endrift: Strictly alphabetically

src/platform/qt/library/LibraryController.h
endrift: Tunnel vision on the feature, not the fine details, I would wager.

endrift:
Unsure it's related, but I got this warning when testing: `QSortFilterProxyModel: index from wrong model passed to mapToSource`

src/platform/qt/test/suite.h
```diff
@@ -0,0 +1,23 @@
...
+struct TestRegistrar {
+	virtual int main(int argc, char** argv) = 0;
+};
+
+extern QList<TestRegistrar*> registeredTests;
+
+#define M_REGISTER_QT_TEST(Class) \
+	static struct Class ## Registrar : TestRegistrar { \
+		QTEST_MAIN(Class) \
+		Class ## Registrar() { registeredTests << this; } \
```
endrift, line 22: There appears to be an initialization order conflict on my end whereby this function gets called before `registeredTests`'s constructor gets called, I think. It's really not clear what's going on, but it's crashing with a null `this`:

```
#4  0x000055555557fb35 in QList<TestRegistrar*>::append (this=0x555555599510 <registeredTests>, t=@0x7fffffffd8d0: 0x5555555994f8 <registrar_LibraryModelTest>) at /usr/include/qt/QtCore/qlist.h:622
622         if (d->ref.isShared()) {
(gdb) p this->d
$8 = (QListData::Data *) 0x0
```

endrift: My suggestion is convert it into something that doesn't depend on constructors: a POD linked list.

Here's an example impl that works on my end and lets me run the tests:

```patch
diff --git a/src/platform/qt/test/main.cpp b/src/platform/qt/test/main.cpp
index 28a422415..2bf59edad 100644
--- a/src/platform/qt/test/main.cpp
+++ b/src/platform/qt/test/main.cpp
@@ -5,11 +5,11 @@
 #include "suite.h"

-QList<TestRegistrar*> registeredTests;
+TestRegistrar* TestRegistrar::first = nullptr;

 int main(int argc, char** argv) {
        int err = 0;
-       for (TestRegistrar* test : registeredTests) {
+       for (TestRegistrar* test = TestRegistrar::first; test; test = test->next) {
                int result = test->main(argc, argv);
                if (result && !err) {
                        err = result;
diff --git a/src/platform/qt/test/suite.h b/src/platform/qt/test/suite.h
index 6aae5f7e0..b5d40427f 100644
--- a/src/platform/qt/test/suite.h
+++ b/src/platform/qt/test/suite.h
@@ -12,12 +12,13 @@

 struct TestRegistrar {
        virtual int main(int argc, char** argv) = 0;
-};

-extern QList<TestRegistrar*> registeredTests;
+       static TestRegistrar* first;
+       TestRegistrar* next = nullptr;
+};

 #define M_REGISTER_QT_TEST(Class) \
        static struct Class ## Registrar : TestRegistrar { \
                QTEST_MAIN(Class) \
-               Class ## Registrar() { registeredTests << this; } \
+               Class ## Registrar() { next = first; first = this; } \
        } registrar_ ## Class;
```

endrift (changes requested):

src/platform/qt/test/suite.h
ahigerd: For documentation: You said you'd rather have one suite per executable, so this is moot and I'm just going to refactor this out.

ahigerd:
> Unsure it's related, but I got this warning when testing: `QSortFilterProxyModel: index from wrong model passed to mapToSource`

It's related. Checking, I see it too. Haven't pinpointed it yet. I'll hook up QAbstractItemModelTester and see if it sheds light on it.

src/platform/qt/library/LibraryController.cpp
```diff
@@ -120,6 +120,13 @@ void LibraryController::selectEntry(const QString& fullpath) {
 		return;
 	}
 	QModelIndex index = m_libraryModel->index(fullpath);
+
+	// If the model is proxied in the current view, map the index to the proxy
+	QAbstractProxyModel* proxy = qobject_cast<QAbstractProxyModel*>(m_currentView->model());
+	if (proxy) {
+		index = proxy->mapFromSource(index);
```
ahigerd, line 127: This was the problem. Fixed it!

ahigerd:
Rebased in order to pick up the platform icons.

ahigerd:
Possible addition: https://github.com/ahigerd/mgba/pull/1/files

I don't know if this is a good idea or not, so I'm keeping it separate until it can be reviewed.

src/platform/qt/CMakeLists.txt
```diff
@@ -481,3 +491,23 @@ if(DISTBUILD AND NOT APPLE)
...
+if(BUILD_SUITE)
+	enable_testing()
+	find_package(${QT}Test)
+	if(${QT}Test_FOUND)
+		get_property(ALL_TESTS DIRECTORY PROPERTY VARIABLES)
+		list(FILTER ALL_TESTS INCLUDE REGEX "^TEST_QT_.*_SRC$")
+		foreach(TEST_SRC ${ALL_TESTS})
+			string(REGEX REPLACE "^TEST_QT_(.*)_SRC$" "\\1" TEST_NAME ${TEST_SRC})
+			message("TEST_NAME=${TEST_NAME}")
+			message("TEST_SRC=${${TEST_SRC}}")
```
endrift, line 504: Please remove the debug code

src/platform/qt/library/LibraryController.cpp
```diff
@@ -213,16 +248,46 @@ void LibraryController::loadDirectory(const QString& dir, bool recursive) {
...
+}
+
+void LibraryController::resizeTreeView(bool expand) {
+	if (expand) {
+		m_treeView->expandAll();
 	}
-	if (m_libraryTree) {
-		m_libraryTree->setShowFilename(m_showFilename);
+
+	int viewportWidth = m_treeView->viewport()->width();
```
endrift, line 270: Not sure if it's still a problem, but I had issues with the column sizes at certain window sizes--I think if the window was too narrow. Worth checking different window sizes.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,424 @@
...
+	iconGBA.addFile(":/res/gba-icon-256.png", QSize(256, 256));
+	iconGBA.addFile(":/res/gba-icon-128.png");
+
+	QIcon iconGBC;
+	iconGBC.addFile(":/res/gbc-icon-128.png", QSize(128, 128));
+	iconGBC.addFile(":/res/gbc-icon-256.png", QSize(256, 256));
+
+	QIcon iconGB;
+	iconGB.addFile(":/res/gb-icon-128.png", QSize(128, 128));
+	iconGB.addFile(":/res/gb-icon-256.png", QSize(256, 256));
```
endrift, line 35: I've rendered some more sizes, but they look kinda crappy since the icon was designed to be viewed at 128x128. I'm probably gonna upload some specific sizes though; I don't want to add several dozen kB to the repo that may or may not be useful.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,424 @@
...
+	// These will silently and harmlessly fail if QSvgIconEngine isn't compiled in.
+	iconGBA.addFile(":/res/gba-icon.svg");
+	iconGBA.addFile(":/res/gbc-icon.svg");
+	iconGB.addFile(":/res/gb-icon.svg");
+	// iconNDS.addFile(":/res/nds-icon.svg");
+
+	m_icons[mPLATFORM_GBA] = iconGBA;
+	// m_icons[mPLATFORM_GBC] = iconGBC;
+	m_icons[mPLATFORM_GB] = iconGB;
+	// m_icons["NDS"] = iconNDS;
```
endrift, line 50: `mPLATFORM_DS`

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,424 @@
...
+					}
+					return entry->displayTitle(m_showFilename);
+				case COL_LOCATION:
+					return QDir::toNativeSeparators(entry->base);
+				case COL_PLATFORM:
+					return nicePlatformFormat(entry->platform);
+				case COL_SIZE:
+					return (role == Qt::DisplayRole) ? QVariant(niceSizeFormat(entry->filesize)) : QVariant(int(entry->filesize));
+				case COL_CRC32:
+					return (role == Qt::DisplayRole) ? QVariant(QStringLiteral("%0").arg(entry->crc32, 8, 16, QChar('0'))) : QVariant(entry->crc32);
```
endrift, line 368: Cases don't get indented.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,424 @@
...
+			case COL_NAME:
+				return QApplication::translate("LibraryTree", "Name", nullptr);
+			case COL_LOCATION:
+				return QApplication::translate("LibraryTree", "Location", nullptr);
+			case COL_PLATFORM:
+				return QApplication::translate("LibraryTree", "Platform", nullptr);
+			case COL_SIZE:
+				return QApplication::translate("LibraryTree", "Size", nullptr);
+			case COL_CRC32:
+				return QApplication::translate("LibraryTree", "CRC32", nullptr);
```
endrift, line 386: Cases don't get indented.

endrift (changes requested):
Some small nits and a question

src/platform/qt/library/LibraryModel.cpp
ahigerd: Depends on the resolution of the linked PR.

src/platform/qt/library/LibraryController.cpp
ahigerd: That should work? This puts a minimum usable size on the column so it'll just scroll horizontally if necessary.

endrift: I remembered what the bug was: if you start with the library not displayed, then enable it in settings, the resizing doesn't trigger. I had a small window so I thought that was related.

src/platform/qt/library/LibraryModel.cpp
ahigerd: *facepalm* I don't indent them in my typical style, either, but I didn't notice that the autoindent changed it when I was fixing spaces-vs-tabs. ^^()

src/platform/qt/library/LibraryController.cpp
ahigerd: Oh, interesting. Easy enough to fix.

src/platform/qt/library/LibraryModel.cpp
ahigerd: Went ahead and switched over to the data structure used by the GBC PR since it'll be more robust against future modifications.

ahigerd:
Oops. Forgot to add the new icons to the qrc.

endrift:
Can you rebase this on current master?

src/core/library.c
```diff
@@ -91,12 +96,38 @@ static void _bindConstraints(sqlite3_stmt* statement, const struct mLibraryEntry
 		sqlite3_bind_int(statement, useIndex, 1);
 		sqlite3_bind_int(statement, index, constraints->platform);
 	}
+
+	if (constraints->platformModels != M_LIBRARY_MODEL_UNKNOWN) {
+		index = sqlite3_bind_parameter_index(statement, ":models");
```
endrift, line 101: the `useIndex` equivalent as well as actually updating the CONSTRAINTS definition seem to be missing

src/platform/qt/Window.cpp
```diff
@@ -131,7 +131,7 @@ Window::Window(CoreManager* manager, ConfigController* config, int playerId, QWi
 	ConfigOption* showFilenameInLibrary = m_config->addOption("showFilenameInLibrary");
 	showFilenameInLibrary->connect([this](const QVariant& value) {
 			m_libraryView->setShowFilename(value.toBool());
-	}, this);
```
endrift, line 134: Ditto.

endrift (changes requested):
I don't have energy to review the rest of this atm (I should probably go to bed), but here's a review of the first 4 files for now, since there's stuff to change already. I'll try to get back to the rest soon.

src/platform/qt/utils.cpp
```diff
@@ -150,6 +161,47 @@ QString keyName(int key) {
 		return QObject::tr("Menu");
 	default:
 		return QKeySequence(key).toString(QKeySequence::NativeText);
+  }
```
endrift, line 164: Nit: Indentation style issue

endrift (commented):
I don't have the mental energy to go over the Qt changes at the moment, but barring a style nitpick, the other stuff looks good so far. I'll take a closer look in a few days and then hopefully merge it.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,418 @@
...
+	}
+	const LibraryEntry* entry = nullptr;
+	if (m_treeMode) {
+		if (!index.parent().isValid()) {
+			return folderData(index, role);
+		}
+		QString path = m_pathOrder[index.parent().row()];
+		entry = m_pathIndex[path][index.row()];
+	} else if (!index.parent().isValid() && index.row() < m_games.size()) {
+			entry = &m_games[index.row()];
```
ahigerd, line 343: Note to self: something's weird with the indentation here.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,417 @@
...
+	}
+	if (index.column() == COL_NAME) {
+		QString path = m_pathOrder[index.row()];
+		return path.section('/', -1);
+	}
+	return QVariant();
+}
+
+QVariant LibraryModel::data(const QModelIndex& index, int role) const {
+	if (role != Qt::DisplayRole &&
```
endrift, line 314: I think it might be a bit less repetitive to turn this into a switch/case with default being the early return and the valid fall-through cases just being a break statement.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,417 @@
...
+				return entry->displayTitle(m_showFilename);
+			case COL_LOCATION:
+				return QDir::toNativeSeparators(entry->base);
+			case COL_PLATFORM:
+				return nicePlatformFormat(entry->platform);
+			case COL_SIZE:
+				return (role == Qt::DisplayRole) ? QVariant(niceSizeFormat(entry->filesize)) : QVariant(int(entry->filesize));
+			case COL_CRC32:
+				return (role == Qt::DisplayRole) ? QVariant(QStringLiteral("%0").arg(entry->crc32, 8, 16, QChar('0'))) : QVariant(entry->crc32);
+		}
```
endrift, line 362: style nit: don't indent the case labels

src/platform/qt/library/LibraryModel.h
```diff
@@ -0,0 +1,84 @@
...
+Q_OBJECT
+
+public:
+	enum Columns {
+		COL_NAME = 0,
+		COL_LOCATION = 1,
+		COL_PLATFORM = 2,
+		COL_SIZE = 3,
+		COL_CRC32 = 4,
+		MAX_COLUMN = 4,
```
endrift, line 33: Generally speaking, I prefer using an exclusive bound because you don't need to manually assign the value that way. You'll need to update the <= to < in places of course, but I find it somewhat easier to maintain. Especially if we're gonna add columns.

src/platform/qt/utils.cpp
```diff
@@ -5,7 +5,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 #include "utils.h"
 
+#include <mgba/core/library.h>
+#include <mgba/gb/interface.h>
```
endrift, line 9:
```suggestion
#ifdef M_CORE_GB
#include <mgba/gb/interface.h>
#endif
```

src/platform/qt/utils.cpp
```diff
@@ -29,18 +33,25 @@ QString niceSizeFormat(size_t filesize) {
 	return unit.arg(size, 0, 'f', int(size * 10) % 10 ? 1 : 0);
 }
 
-QString nicePlatformFormat(mPlatform platform) {
+QString nicePlatformFormat(mPlatform platform, int validModels) {
 	switch (platform) {
 #ifdef M_CORE_GBA
 	case mPLATFORM_GBA:
-		return QObject::tr("GBA");
```
endrift, line 36: Why get rid of the translation?

endrift (commented):
A few minor notes so far, but mainly I think the dense code needs some better spacing to break up the logical blocks, and perhaps some comments explaining how the methods work. I'm pretty tired atm so my eyes kinda glossed over on some of the blocks of code.

One big thing is that now that SHA-1 stuff has been merged we'll probably want to figure out how to present that. Maybe remove the CRC32 column for now and figure out optional columns later, including CRC32 and SHA-1? Or leave it for a "more info" dialog when/if we get there.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,417 @@
...
+	if (m_treeMode) {
+		const LibraryEntry& entry = *m_games[pos];
+		return createIndex(m_pathIndex[entry.base].indexOf(&entry), 0, m_pathOrder.indexOf(entry.base));
+	}
+	return createIndex(pos, 0);
+}
+
+QModelIndex LibraryModel::index(int row, int column, const QModelIndex& parent) const {
+	if (!parent.isValid()) {
+		return createIndex(row, column, quintptr(0));
```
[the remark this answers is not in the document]
ahigerd: Because the function takes an integer type, not a pointer type.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,417 @@
...
+		return QVariant();
+	}
+	if (role == Qt::ToolTipRole && index.column() > COL_LOCATION) {
+		return QVariant();
+	}
+	if (role == Qt::DecorationRole && index.column() != COL_NAME) {
+		return QVariant();
+	}
+	if (role == Qt::TextAlignmentRole) {
+		return index.column() == COL_SIZE ? (int)(Qt::AlignTrailing | Qt::AlignVCenter) : (int)(Qt::AlignLeading | Qt::AlignVCenter);
```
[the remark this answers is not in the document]
ahigerd: While that's shorter, I'm not sure I consider it simpler? There's more places you have to look to be able to tell what the final value is going to be.

src/platform/qt/library/LibraryModel.cpp
```diff
@@ -0,0 +1,417 @@
...
+	removedRootSpans.sort(true);
+	for (const SpanSet::Span& span : removedRootSpans.spans) {
+		beginRemoveRows(QModelIndex(), span.left, span.right);
+		if (m_treeMode) {
+			for (int i = span.right; i >= span.left; i--) {
+				QString base = m_pathOrder.takeAt(i);
+				m_pathIndex.remove(base);
+			}
+		} else {
+			m_games.erase(m_games.begin() + span.left, m_games.begin() + span.right + 1);
```
ahigerd, line 236: Hmm... I think I need to add another case to the unit test, because I'm not sure it ever frees the memory in tree mode...

... _oh._ THAT'S what that empty loop I removed was for. I wonder how the loop contents got removed.

src/platform/qt/library/LibraryModel.h
ahigerd: I broadly agree but in terms of the implementation I have more reason to refer to the last column than to the number of columns; it's not just there for iteration. I'll make the switch if you insist but I'll be adding some `- 1`'s in places and removing `+ 1`'s in others.

src/platform/qt/utils.cpp
ahigerd: Because these are used directly as hash keys, so translating them would break it.

endrift (approved):
Can you fixup the git history so the fix commits are squashed? Keeping the backend and frontend commits separate is fine, I just don't want the whole history.

PR #2575, opened 2022-07-01
doc: Updated the README.md to include more important information for setting up Docker for building mGBA.

tommai78101:


README.md
```diff
@@ -41,7 +41,7 @@ Features
 - Community-provided translations for several languages via [Weblate](https://hosted.weblate.org/engage/mgba).
 - Many, many smaller things.
 
-#### Game Boy mappers
```
endrift, line 44: Please don't change the indent levels throughout this file.

README.md
```diff
@@ -120,15 +120,29 @@ Compiling
 
 Compiling requires using CMake 3.1 or newer. GCC and Clang are both known to work to compile mGBA, but Visual Studio 2013 and older are known not to work. Support for Visual Studio 2015 and newer is coming soon.
 
-#### Docker building
+## Docker building
 
-The recommended way to build for most platforms is to use Docker. Several Docker images are provided that contain the requisite toolchain and dependencies for building mGBA across several platforms.
+The recommended way to build for most platforms is to use Docker. Several Docker images are provided that contain the requisite toolchain and dependencies for building mGBA across several platforms. 
+
+First, you need to make sure you are within the mGBA repository root directory, and it is located in a directory where it's in a *nix path. Docker is known to be stringent on this restriction. (See issue #1985 for details.)
```
endrift, line 127: This restriction is specific to Docker for WIndows only. Also, for linking to issues (I should double check the issue here specifically), use `[#XXXX](https://mgba.io/i/XXXX)` format to make sure it links properly.

endrift (changes requested):
Looks mostly fine, but there are some specific issues you should address.

README.md
tommai78101: Oh this is cool! I'll update it. Thanks.

README.md
tommai78101: Will revert. Sorry.

README.md
```diff
@@ -40,9 +40,9 @@ Features
 - Cores available for RetroArch/Libretro and OpenEmu.
 - Community-provided translations for several languages via [Weblate](https://hosted.weblate.org/engage/mgba).
 - Many, many smaller things.
-
 #### Game Boy mappers
 
+
```
endrift, line 45: Looks like the newline got transposed by mistake.

README.md
```diff
@@ -122,13 +122,29 @@ Compiling requires using CMake 3.1 or newer. GCC and Clang are both known to wor
 
 #### Docker building
 
-The recommended way to build for most platforms is to use Docker. Several Docker images are provided that contain the requisite toolchain and dependencies for building mGBA across several platforms.
+The recommended way to build for most platforms is to use Docker. Several Docker images are provided that contain the requisite toolchain and dependencies for building mGBA across several platforms. 
+
+Note: If you are on Windows, you need to make sure you are within the mGBA repository root directory, and it is located in a directory where it's in a *nix path. Docker is known to be stringent on this restriction. (See issue [#1985](https://mgba.io/i/1985) for details.)
```
endrift, line 127: This is still not technically accurate. The problem is that by default, Docker is restricted to being able to load things out of your home directory (C:\Users\<User>), but this can be overridden. You might need to search around for more info, and it might have been changed in the meantime. Did you run into this issue specifically?

endrift: Alright, I might need to look into that more. That said, uh, the directory you listed isn't inside the home directory.

README.md
tommai78101: Sorry. Will fix this.

README.md
tommai78101: Ahhh, when I reverted some changes, it reverted the `/usr` Will fix.

endrift: I don't think you understand what I mean. It cares about the *Windows* home directory, not the "*NIX" home directory, which is a concept Docker for Windows has no idea about.

tommai78101: Wait, let me do some experiments first.

tommai78101: A bit of research shows:

- Docker Toolbox (or basically, Docker running on Windows systems older than Windows 10) has this limitation where you cannot run Docker images outside of the User home directory, unless you configure it or you put the image inside a VirtualBox container with a shared folder configured. Shared folders will map folder path A to folder path B. 

- Docker Desktop for Windows 10 and up have better native integration, and can run Docker images anywhere, so as long as you use `--mount` (Bind Mounts) or `--volume` (Volumes), the `source` and `target` folder destinations can be specified to be mapped as shared folders. This mounting is the same as mapping folder path A to folder path B.

https://forums.docker.com/t/mounting-folders-outside-the-users-folder-on-windows/65662/3

So it is wrong of me to assume that we need to put the mGBA checkout repository inside a "*nix" home directory, but rather, just specify the Docker command as mentioned in the README.md. It's only when you are using an older Windows system, do you then need to rely on VirtualBox shared folder configurations.

I have updated the README.md to reflect upon this new information.

README.md
```diff
@@ -122,13 +122,23 @@ Compiling requires using CMake 3.1 or newer. GCC and Clang are both known to wor
 
 #### Docker building
 
-The recommended way to build for most platforms is to use Docker. Several Docker images are provided that contain the requisite toolchain and dependencies for building mGBA across several platforms.
+The recommended way to build for most platforms is to use Docker. Several Docker images are provided that contain the requisite toolchain and dependencies for building mGBA across several platforms. 
+
+Note: If you are on Windows, you need to make sure your terminal session is inside the mGBA repository root directory before using Docker. (See issue [#1985](https://mgba.io/i/1985) for details.) If you are on an older Windows system, you may need to configure your Docker to use VirtualBox shared folders to correctly map your current `mgba` checkout directory to the Docker image's working directory.
+
+For example, on Windows, after cloning the repository, we will assume the `mgba` checkout directory is located at `X:\github\mgba`. What you want to do is to open up the terminal (Powershell or WSL) and navigate to the `mgba` root directory before proceeding:
```
endrift, line 129: I feel like this is still a misunderstanding of what's going on. I think what you're trying to explain here is just to change directory into the directory the checkout is in. This applies to whatever OS or shell you're on. That's just to get the `$PWD` working later on. It doesn't need to be `$PWD`, that was just to simplify the steps. But I feel like this is de-simplifying it now.

README.md
```diff
@@ -141,6 +151,8 @@ This will produce a `build-win32` directory with the build products. Replace `mg
 - mgba/windows:w32
 - mgba/windows:w64
 
+If you wished to speed up the build process, consider adding the flag `-e MAKEFLAGS=-j4` to do a parallel build for mGBA with 4 cores instead of one core, if your CPU supports it. Change the `4` to however many cores you have on your build machine.
```
endrift, line 154: Just say -jN where N is the desired number of cores. This is needlessly verbose now.

README.md
```diff
@@ -122,13 +122,23 @@ Compiling requires using CMake 3.1 or newer. GCC and Clang are both known to wor
...
+
+	cd X:\github\mgba
 
 To use a Docker image to build mGBA, simply run the following command while in the root of an mGBA checkout:
 
-	docker run --rm -t -v ${PWD}:/home/mgba/src mgba/windows:w32
+	docker run --rm -it -v ${PWD}:/home/mgba/src mgba/windows:w32
 
-This will produce a `build-win32` directory with the build products. Replace `mgba/windows:w32` with another Docker image for other platforms, which will produce a corresponding other directory. The following Docker images available on Docker Hub:
+Depending on which terminal you are running the command from and whether your current working path may contain spaces or special characters, you may need to change the `${PWD}` to `$PWD` or `"$PWD"` to run the command.
```
endrift, line 137: I don't think it's up to mGBA to explain how your shell works.

README.md
```diff
@@ -122,13 +122,23 @@ Compiling requires using CMake 3.1 or newer. GCC and Clang are both known to wor
...
+
+What we are doing is to create a shared folder, mapping the `${PWD}` to the target directory, `/home/mgba/src`, so it can be used as volumes in the Docker image, completely managed by Docker.
```
endrift, line 139: Is this information important? It's just explaining how Docker works. Which is again not up to mGBA.

endrift (commented):
I'm not really sure what this information is helpful for at this point.

README.md
tommai78101: Removing all those explanations. I supposed I didn't really understand what my dev machine's problem was, and why I sought out that Github issue in the first place to get my mGBA build up and working.

README.md
tommai78101: I agreed. I simplified it. Thanks for the feedback.

README.md
tommai78101: Removing it. I thought it might be helpful to explain a bit about the curly brackets, and why on different platforms and terminal types, you need to remove curly brackets. I thought that, without this explanation, people would just straight up copy/paste the command without modifying it.

endrift: To some extent you're right, but also this section won't stop them from copy/pasting it. Putting the most widely compatible version there is probably a good idea. I think that might be `"${PWD}"`.

README.md
```diff
@@ -122,13 +122,15 @@ Compiling requires using CMake 3.1 or newer. GCC and Clang are both known to wor
...
+
+Note: If you are on an older Windows system before Windows 10, you may need to configure your Docker to use VirtualBox shared folders to correctly map your current `mgba` checkout directory to the Docker image's working directory. (See issue [#1985](https://mgba.io/i/1985) for details.) 
 
 To use a Docker image to build mGBA, simply run the following command while in the root of an mGBA checkout:
 
-	docker run --rm -t -v ${PWD}:/home/mgba/src mgba/windows:w32
+	docker run --rm -it -v ${PWD}:/home/mgba/src mgba/windows:w32
 
-This will produce a `build-win32` directory with the build products. Replace `mgba/windows:w32` with another Docker image for other platforms, which will produce a corresponding other directory. The following Docker images available on Docker Hub:
+When running the Docker image, this will produce a `build-win32` directory with the build products. Replace `mgba/windows:w32` with another Docker image for other platforms, which will produce a corresponding other directory. The following Docker images available on Docker Hub:
```
endrift, line 133:
```suggestion
After starting the Docker container, it will produce a `build-win32` directory with the build products. Replace `mgba/windows:w32` with another Docker image for other platforms, which will produce a corresponding other directory. The following Docker images available on Docker Hub:
```

I think this phrasing sounds less stilted (and is more technically accurate to Docker terminology)

README.md
```diff
@@ -141,6 +143,8 @@ This will produce a `build-win32` directory with the build products. Replace `mg
 - mgba/windows:w32
 - mgba/windows:w64
 
+If you wished to speed up the build process, consider adding the flag `-e MAKEFLAGS=-jN` to do a parallel build for mGBA with `N` number of CPU cores.
```
endrift, line 146:
```suggestion
If you want to speed up the build process, consider adding the flag `-e MAKEFLAGS=-jN` to do a parallel build for mGBA with `N` number of CPU cores.
```

endrift (changes requested):
Mostly just some stilted phrasing issues now. I'll probably be fine with it after these.

Issue #2517, opened 2022-05-08
[Feature addition] ROM saving for GBA bootlegs


metroid-maniac:
I want to add support to mGBA for so-called batteryless bootleg GBA games which store persistent save data in the same region as the ROM.

#2189 and #2311 are reports of people using such bootleg ROMs. Patches to make these ROMs can be found [here](https://github.com/acocalypso/batteryless-patches).

I've made an attempt at implementing this feature on the fork [here](https://github.com/metroid-maniac/mgba). While it can run these ROMs with functional saving, there is still a lot of code cleanup that needs to be done. If this feature would be accepted I'd like to discuss how this should be fixed up before progressing, because the current approach will require some changes to the way ROMs are loaded.

GB/C bootlegs can also be found with this sort of patch, but as I don't have a way to dump or flash these bootlegs at the moment I am ignoring them.

endrift:
I would like something like this, but it's a bit tricky, as I'm sure you already know. I would need to look at your implementation. You probably already noticed the copy-on-write mechanic, but I was assuming the save could be written as some sort of overlay. Perhaps using XOR to make sure that only the difference between the base ROM and the save is written to disk.

metroid-maniac:
Right now I've added a new cartridge override type for rewritable ROM. I'm currently writing back the changes directly to the ROM file. I don't like this as a conclusive solution - it's just a stopgap to see if the flash state machines were implemented correctly enough for these games to load and save correctly. 
Since mGBA loads the ROM before overrides are applied this also leads to having to pessimistically open and map the ROM with read/write permissions. Storing the difference as an overlay in a different file would be better.

A solution which requires a minimal amount of changes would probably be to add rewritable ROMs as new save types. Then, the overlays can be loaded and saved from .sav files. The point of these bootlegs is that they do not have persistent storage otherwise, so it shouldn't be an issue to repurpose the .sav file.
Otherwise we can continue having rewritable ROM as a new override type and emit a new file for the overlay, as a XOR or a diff or something else.

endrift:
I've grabbed some bootlegs off of aliexpress that should get me some good examples of cartridges. I don't think I'm gonna get to this this release though.

PR #2582, opened 2022-07-07
GBA Save: Add support for 128kB SRAM and FlashROM saving (bootlegs)

metroid-maniac:
Add support for 128kB SRAM saving. Used by some Pokemon and Super Mario Advance 4 bootlegs. GBATA-style SRAM bank switching is also supported for support with SRAM-patched ROMs for older flashcarts.
Add support for rewritable ROM using flash commands. Used by "batteryless bootlegs" for saving.  ROM overlay is stored in the .sav file instead of SRAM, as these carts do not have batteries to retain SRAM contents persistently. SRAM is still used temporarily during gameplay.
Add some toggles to the Qt frontend to configure these parameters. 
Patches for batteryless ROMs can be found at https://github.com/acocalypso/batteryless-patches , Super Mario Advance 4 is of  interest as it also uses 128KB SRAM and will not save properly unless this is correctly emulated.

Relates to issue #2517


src/gba/overrides.c
```diff
@@ -247,6 +247,9 @@ bool GBAOverrideFind(const struct Configuration* config, struct GBACartridgeOver
 			} else if (strcasecmp(savetype, "SRAM512") == 0) {
 				found = true;
 				override->savetype = SAVEDATA_SRAM512;
+			} else if (strcasecmp(savetype, "SRAM1M)") == 0) {
```
endrift, line 250:
```suggestion
			} else if (strcasecmp(savetype, "SRAM1M") == 0) {
```

include/mgba/internal/gba/flashrom.h
```diff
@@ -0,0 +1,71 @@
...
+#include <mgba-util/common.h>
+#include <mgba-util/memory.h>
+
+CXX_GUARD_START
+
+enum FlashROMType {
+	FLASHROM_NONE,
+	FLASHROM_22XX,
+	FLASHROM_INTEL
+};
```
endrift, line 13: This should probably be in the Savedata type enum instead. Actually, most of or all of this should be rolled into the savedata struct.

include/mgba/internal/gba/flashrom.h
```diff
@@ -0,0 +1,71 @@
...
+	FLASHROM_INTEL_ERASE,
+	FLASHROM_INTEL_PROGRAM,
+	FLASHROM_INTEL_STATUS
+};
+
+enum {
+	FLASHROM_BLOCK_SIZE = 0x10000,
+};
+
+typedef uint8_t FlashROMBlock[FLASHROM_BLOCK_SIZE];
```
endrift, line 43: Typedefs for a sized array are probably a bad idea. Certainly confusing.

src/platform/qt/OverrideView.ui
```diff
@@ -204,6 +209,39 @@
...
+          </item>
+          <item row="3" column="0">
+           <widget class="QLabel">
+            <property name="text">
+             <string>FlashROM (bootlegs)</string>
+            </property>
+           </widget>
+          </item>
+          <item row="3" column="1">
+           <widget class="QComboBox" name="flashrom">
```
endrift, line 227: This should just be more fields in the save types.

endrift (changes requested):
I haven't managed to review much of the actual implementation yet, but to start with you should roll a lot of the novel structs and enums and fields into the existing savedata struct. The flash type does not need to be a separate enum, and a lot of stuff gets simplified if it's rolled into the existing structures.

Also, it crashed when I tried to use it. I'll look into why.

include/mgba/internal/gba/flashrom.h
metroid-maniac: I had a reason for separating it out. SMA4 needs 1M SRAM and will give warnings if 512K SRAM is present, but Pokemon Mystery Dungeon will corrupt its save if 1M SRAM is present and 512K SRAM is needed. Separating it out seemed reasonable.

In practice, I think there will only be 4 needed FlashROM configs for maximal compatibility (22xx 512K, 22xx 1M, 1M, Intel 512K, Intel 1M) so maybe it's better to just enumerate them in the normal save list.

metroid-maniac:
Going back to the drawing board with this one

Issue #905, opened 2017-10-12
Custom keyboard defaults for languages


ShalokShalom:
German keyboards suffer from the current default setting. In order to provide sane defaults for them, implement "custom keybindings for the different language settings"

endrift:
@rootfather Any input on this? I can try to provide an interface for it.

ShalokShalom:
^-^

rootfather:
@endrift indeed, the default keyboard layout is not optimized for German keyboards with a standard QWERTZ keyboard.

As an example, take the default mapping for the A and B button. Per default, the "B" button is mapped to the key Z and the A button is mapped to X. On a QWERTY keyboard, this works perfectly fine and is a good default setting. However, on a German keyboard that uses the QWERTZ layout, the "Y" and the "Z" keys are swapped compared to QWERTY which renders the default "A" and "B" mapping quite useless.

I agree with @ShalokShalom that an interface which provides custom defaults for the different keyboard layouts out there would be a nice addition.

depp:
I also experience this issue. Looks like #393 and #2020 are related. My thoughts:

* Fixing this issue is easy with the SDL front-end. SDL already provides a platform-independent way to map keys by keyboard location rather than character.
* Qt provides `QKeyEvent::nativeScanCode()` that provides the information we want.
* Unfortunately, Qt's `nativeScanCode()` is platform-specific.
* The scancodes for various platforms are known. SDL has tables, for example. It would require some work.
* To display the input configuration, you'd want to map the configured scancodes to key names, which is not straightforward. Qt does not provide this. There are some platform-specific techniques.
* It would be nice if this didn't break people's existing configurations.

PR #2605, opened 2022-07-30
SDL: Use scancodes for keyboard bindings (#905)

depp:
Core key bindings for SDL >= 2.0 are now independent of the keyboard
layout. QWERTY users get ZXAS, German users get YXAS, French layouts get
WXQS, etc. Existing customizations will be ignored, users with
customized bindings for SDL will get the new defaults and have to create
a new configuration.

Existing functionality is preserved for SDL 1.x.


endrift (commented):
I have some reservations about this change:

The first and foremost one is that it won't load custom SDLK bindings if there are any present, instead falling back to the new SDLS bindings.

Secondly, which OSes have you tested this on? I seem to recall trying something like this years ago, but running into issues with it on some OSes. It'll need broader testing.

Third, it might (might) be desirable to be able to select at runtime whether or not scancodes or keysyms are used. But I'm not sure.

I have one implementation nitpick, specifically that I use `#define DEF` and `#ifdef DEF` instead of `#define DEF 1` and `#if DEF` but that's a simple enough change once more prominent concerns are allayed or addressed.

Targeting this for 0.11 though since 0.10 is frozen for this kind of change.

depp:
> The first and foremost one is that it won't load custom SDLK bindings if there are any present, instead falling back to the new SDLS bindings.

You could do this with `SDL_GetScancodeFromKey()`, but it might take me some time to understand the configuration system well enough to make this change.

> Secondly, which OSes have you tested this on? I seem to recall trying something like this years ago, but running into issues with it on some OSes. It'll need broader testing.

I've only personally tested this code on Linux, but I've written similar code and tested it on Windows and macOS before. I can test on additional systems.

> Third, it might (might) be desirable to be able to select at runtime whether or not scancodes or keysyms are used. But I'm not sure.

I have a hard time seeing the benefit. SDL scancodes are portable. Keysyms are not portable, they are fragile and break on some systems. Some people even change keyboard layout while programs are running--and this would break any keysym layout.

endrift:
I know I'm *several* years late, but I am amenable to this change, but only with an SDLK -> SDLS migration path. I'm not sure offhand what the best way to detect if SDLS should be used/if a migration has already happened though. Perhaps a config key for "isMigrated" that's set when a migration happens.

PR #2611, opened 2022-08-07
Script bindings for sockets

ahigerd:
This PR provides an event-driven TCP socket API for mGBA scripting. Where the behaviors overlap, it roughly imitates the LuaSocket API, but due to the mGBA scripting core not being yieldable, it can't be a perfect match.

Also included are two sample scripts. Both scripts will log messages received over the socket to the console and will send an input viewer. One script will connect to a server; the other will open a server and listen for incoming connections.

Documentation still needs to be written.

### DIFF
--- a/include/mgba-util/socket.h
+++ b/include/mgba-util/socket.h
@@ -21,11 +21,14 @@ CXX_GUARD_START
 typedef SOCKET Socket;
 #else
 #ifdef GEKKO
+#define USE_GETHOSTBYNAME
 #include <network.h>
 #else
 #include <arpa/inet.h>
+#include <netdb.h>
 #include <netinet/in.h>
 #include <netinet/tcp.h>
+#include <sys/types.h>
 #include <sys/socket.h>
 #endif
 #include <errno.h>
@@ -191,6 +194,17 @@ static inline Socket SocketOpenTCP(int port, const struct Address* bindAddress) 
 	}
 
 	int err;
+
+	int enable = 1;
+#ifdef GEKKO
+	err = net_setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, &enable, sizeof(enable));
+#else
+	err = setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, &enable, sizeof(enable));
+#endif
+	if (err) {
+		return INVALID_SOCKET;
+	}
+
 	if (!bindAddress) {
 		struct sockaddr_in bindInfo;
 		memset(&bindInfo, 0, sizeof(bindInfo));
@@ -435,6 +449,62 @@ static inline int SocketPoll(size_t nSockets, Socket* reads, Socket* writes, Soc
 			++e;
 		}
 	}
+	return result;
+}
+
+static inline int SocketResolveHost(const char* addrString, struct Address* destAddress) {
+	int result = 0;
+#ifdef USE_GETHOSTBYNAME
+#warning Using gethostbyname() for hostname resolution is not threadsafe
+#ifdef GEKKO
+	struct hostent* host = net_gethostbyname(addrString);
+#else
+	struct hostent* host = gethostbyname(addrString);
+#endif
+	if (!host) {
+		return errno;
+	}
+	family = host->h_addrtype;
+	if (host->h_addrtype == AF_INET && host->h_length == 4) {
+		destAddress->version = IPV4;
+		destAddress->ipv4 = ntohl(*host->h_addr_list[0]);
+	} else if (host->h_addrtype == AF_INET6 && host->h_length == 16) {
+		destAddress->version = IPV6;
+		memcpy(destAddress->ipv6, host->h_addr_list[0], 16);
+	} else {
+		result = NO_DATA;
+	}
+#else
+	struct addrinfo* addr = NULL;
+	result = getaddrinfo(addrString, NULL, NULL, &addr);
+	if (result) {
+#ifndef _WIN32
+		if (result == EAI_SYSTEM) {
+			result = errno;
+		}
+#endif
+		goto error;
+	}
+	if (addr->ai_family == AF_INET && addr->ai_addrlen == sizeof(struct sockaddr_in)) {
+		struct sockaddr_in* addr4 = (struct sockaddr_in*) addr->ai_addr;
+		destAddress->version = IPV4;
+		destAddress->ipv4 = ntohl(addr4->sin_addr.s_addr);
+	} else if (addr->ai_family == AF_INET6 && addr->ai_addrlen == sizeof(struct sockaddr_in6)) {
+		struct sockaddr_in6* addr6 = (struct sockaddr_in6*) addr->ai_addr;
+		destAddress->version = IPV6;
+		memcpy(destAddress->ipv6, addr6->sin6_addr.s6_addr, 16);
+	} else {
+#ifdef _WIN32
+		result = WSANO_DATA;
+#else
+		result = EAI_NODATA;
+#endif
+	}
+error:
+	if (addr) {
+		freeaddrinfo(addr);
+	}
+#endif
 	return result;
 }
 
--- a/include/mgba/internal/script/socket.h
+++ b/include/mgba/internal/script/socket.h
@@ -1,0 +1,25 @@
+/* Copyright (c) 2013-2022 Jeffrey Pfau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#ifndef M_SCRIPT_SOCKET_H
+#define M_SCRIPT_SOCKET_H
+
+enum mSocketErrorCode {
+	mSCRIPT_SOCKERR_UNKNOWN_ERROR = -1,
+	mSCRIPT_SOCKERR_OK = 0,
+	mSCRIPT_SOCKERR_AGAIN,
+	mSCRIPT_SOCKERR_ADDRESS_IN_USE,
+	mSCRIPT_SOCKERR_CONNECTION_REFUSED,
+	mSCRIPT_SOCKERR_DENIED,
+	mSCRIPT_SOCKERR_FAILED,
+	mSCRIPT_SOCKERR_NETWORK_UNREACHABLE,
+	mSCRIPT_SOCKERR_NOT_FOUND,
+	mSCRIPT_SOCKERR_NO_DATA,
+	mSCRIPT_SOCKERR_OUT_OF_MEMORY,
+	mSCRIPT_SOCKERR_TIMEOUT,
+	mSCRIPT_SOCKERR_UNSUPPORTED,
+};
+
+#endif
--- a/include/mgba/script/context.h
+++ b/include/mgba/script/context.h
@@ -10,6 +10,7 @@
 
 CXX_GUARD_START
 
+#include <mgba/core/log.h>
 #include <mgba/script/types.h>
 #include <mgba-util/table.h>
 #include <mgba-util/vfs.h>
@@ -17,6 +18,8 @@ CXX_GUARD_START
 #define mSCRIPT_KV_PAIR(KEY, VALUE) { #KEY, VALUE }
 #define mSCRIPT_CONSTANT_PAIR(NS, CONST) { #CONST, mScriptValueCreateFromSInt(NS ## _ ## CONST) }
 #define mSCRIPT_KV_SENTINEL { NULL, NULL }
+
+mLOG_DECLARE_CATEGORY(SCRIPT);
 
 struct mScriptFrame;
 struct mScriptFunction;
@@ -83,6 +86,7 @@ struct mScriptValue* mScriptContextAccessWeakref(struct mScriptContext*, struct 
 void mScriptContextClearWeakref(struct mScriptContext*, uint32_t weakref);
 
 void mScriptContextAttachStdlib(struct mScriptContext* context);
+void mScriptContextAttachSocket(struct mScriptContext* context);
 void mScriptContextExportConstants(struct mScriptContext* context, const char* nspace, struct mScriptKVPair* constants);
 void mScriptContextExportNamespace(struct mScriptContext* context, const char* nspace, struct mScriptKVPair* value);
 
--- a/src/script/engines/lua.c
+++ b/src/script/engines/lua.c
@@ -5,6 +5,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 #include <mgba/internal/script/lua.h>
 
+#include <mgba/internal/script/socket.h>
+#include <mgba/script/context.h>
 #include <mgba/script/macros.h>
 #include <mgba-util/string.h>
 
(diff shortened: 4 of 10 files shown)


include/mgba-util/socket.h
```diff
@@ -438,6 +441,56 @@ static inline int SocketPoll(size_t nSockets, Socket* reads, Socket* writes, Soc
 	return result;
 }
 
+static inline int SocketResolveHost(const char* addrString, struct Address* destAddress) {
+	int result = 0;
+#ifdef USE_GETHOSTBYNAME
+	// TODO: This isn't threadsafe. Fix!
```
endrift, line 447: Perhaps make this a #warning, especially in the #else case.

include/mgba-util/socket.h
```diff
@@ -438,6 +441,56 @@ static inline int SocketPoll(size_t nSockets, Socket* reads, Socket* writes, Soc
...
+#ifdef GEKKO
+	struct hostent* host = net_gethostbyname(addrString);
+#else
+	struct hostent* host = gethostbyname(addrString);
+#endif
+	if (!host) {
+		result = errno;
+		goto error;
```
endrift, line 455: Since `error` is the end of the functionin this case, an early return is preferred.

src/core/scripting.c
```diff
@@ -818,7 +818,7 @@ mSCRIPT_DEFINE_STRUCT(mScriptTextBuffer)
 	mSCRIPT_DEFINE_STRUCT_METHOD(mScriptTextBuffer, clear)
 	mSCRIPT_DEFINE_DOCSTRING("Set the number of rows and columns")
 	mSCRIPT_DEFINE_STRUCT_METHOD(mScriptTextBuffer, setSize)
-	mSCRIPT_DEFINE_DOCSTRING("Set the posiiton of the cursor")
+	mSCRIPT_DEFINE_DOCSTRING("Set the position of the cursor")
```
endrift, line 821: Consider splitting this off and also the whitespace fixes into a separate PR, especially since those are the only changes for this file.

src/script/engines/lua.c
```diff
@@ -177,6 +266,11 @@ struct mScriptEngineContext* _luaCreate(struct mScriptEngine2* engine, struct mS
 	lua_getglobal(luaContext->lua, "require");
 	luaContext->require = luaL_ref(luaContext->lua, LUA_REGISTRYINDEX);
 
+	int status = luaL_dostring(luaContext->lua, _socketLuaSource);
+	if (status) {
+		printf("dostring: %s\n", lua_tostring(luaContext->lua, -1));
+	}
```
endrift, line 272: Don't use printf. I should check if there's an mLOG category for scripting and add it if not.

src/script/socket.c
```diff
@@ -0,0 +1,194 @@
...
+struct mScriptValue* _mScriptSocketCreate() {
+	struct mScriptSocket client = {
+		.socket = INVALID_SOCKET,
+		.error = 0,
+		.port = 0
+	};
+
+	struct mScriptValue* result = mScriptValueAlloc(mSCRIPT_TYPE_MS_S(mScriptSocket));
+	result->value.opaque = calloc(1, sizeof(struct mScriptSocket));
+	*(struct mScriptSocket*) result->value.opaque = client;
```
endrift, line 39: I believe C99 lets you do `*ptr = (struct mScriptSocket) { ... }`

src/script/socket.c
```diff
@@ -0,0 +1,194 @@
...
+	mSCRIPT_DEFINE_STRUCT_METHOD(mScriptSocket, select)
+	mSCRIPT_DEFINE_DOCSTRING("TODO")
+	mSCRIPT_DEFINE_STRUCT_MEMBER(mScriptSocket, S32, error)
+mSCRIPT_DEFINE_END;
+
+mSCRIPT_DEFINE_STRUCT_BINDING_DEFAULTS(mScriptSocket, listen)
+	mSCRIPT_S32(1)
+mSCRIPT_DEFINE_DEFAULTS_END;
+
+static struct _mScriptSocketError {
```
endrift, line 153: `const`

src/script/socket.c
```diff
@@ -0,0 +1,194 @@
...
+	{ EAI_NONAME, "not known" },
+	{ EAI_NODATA, "no data" },
+#else
+	{ -HOST_NOT_FOUND, "host not found" },
+	{ -TRY_AGAIN, "temporary failure" },
+	{ -NO_RECOVERY, "failed" },
+	{ -NO_DATA, "no data" },
+#endif
+};
+static int _mScriptSocketNumErrors = sizeof(_mScriptSocketErrors) / sizeof(struct _mScriptSocketError);
```
endrift, line 179: `const`

endrift (commented):
Just some nits for now. Will take a closer look later, hopefully today.

src/script/engines/lua.c
ahigerd: Hmm... You know, I was just intending to remove that. It was just there for debugging. But if you think mLOG is a better idea, then sure.

src/script/socket.c
ahigerd: I was just imitating how you'd written it elsewhere.

include/mgba-util/socket.h
ahigerd: It's not the end of the function. It calls `freeaddrinfo` first.

ahigerd: Oh wait. Wrong ifdef branch.

Huh. I thought I removed that one already. Must have rolled back.

src/script/engines/lua.c
ahigerd: _mLOG_CAT_SCRIPT btw

include/mgba-util/socket.h
ahigerd: Why would there be a warning in the #else case?

endrift: To denote that that case is strongly recommended against. It shouldn't be hit in any existing compilation path atm, so it'd be a good idea to tell people _don't hit this path_ if at all possible.

src/script/socket.c
endrift: Don't worry about it then. I can clean up minor tech debt later.

endrift:
Seems to be decent so far, but I'm curious what about it doesn't match, and uh, how the doc generator would handle it. I'm not really sure what the surface API would look like, to be honest. An example script is probably a good idea,

ahigerd:
LuaSocket's `receive` API also has some oddly specific behaviors: https://w3.impa.br/~diego/software/luasocket/tcp.html#receive -- `*l` is useful for line-based protocols, sure, but `*a` is asking for a headache thanks to blocking. It also supports a second parameter that's just prefixed onto the return value. I chose not to implement any of these.

Also, LuaSocket isn't event-driven, so there's no callback API at all.

As far as the doc generator is concerned... Yeah, I have no idea.

I sent an example script in Discord, but I'll reproduce it here:
```lua
s = socket.tcp()
s:on("received", function()
  while true do
    local p, err = s:receive(1024)
    if p then
      console:log(p:match("^(.-)%s*$"))
    else
      if err ~= "temporary failure" then
        console:error(err)
      end
      return
    end
  end
end)
if s:connect("127.0.0.1", 8888) then
  console:log("Connected")
else
  console:log("Failed to connect")
end
```
It connects to a server and echoes anything it receives to the console.

endrift:
I meant putting an example script _in the repo_. Also if you could build the doc generator and give me the yaml output I could see how it does.

src/script/socket.c
```diff
@@ -0,0 +1,224 @@
...
+		data->size = 0;
+		ssock->error = SocketError();
+	} else {
+		data->size = bytes;
+		ssock->error = 0;
+	}
+	return value;
+}
+
+// XXX: This is for prototyping. It's not a good idea.
```
endrift, line 120: Is this comment still accurate?

src/script/socket.c
```diff
@@ -0,0 +1,224 @@
...
+
+mSCRIPT_DEFINE_STRUCT_BINDING_DEFAULTS(mScriptSocket, listen)
+	mSCRIPT_S32(1)
+mSCRIPT_DEFINE_DEFAULTS_END;
+
+static const struct _mScriptSocketError {
+	int32_t err;
+	const char* message;
+} _mScriptSocketErrors[] = {
+	{ EAGAIN, "temporary failure" },
```
endrift, line 175: Should these be initial caps? I'm not used to seeing all lowercase error messages.

src/script/socket.c
```diff
@@ -0,0 +1,224 @@
...
+mSCRIPT_DECLARE_STRUCT_METHOD(mScriptSocket, W(mScriptSocket), accept, _mScriptSocketAccept, 0);
+mSCRIPT_DECLARE_STRUCT_METHOD(mScriptSocket, S32, open, _mScriptSocketOpen, 2, STR, address, U16, port);
+mSCRIPT_DECLARE_STRUCT_METHOD(mScriptSocket, S32, connect, _mScriptSocketConnect, 2, STR, address, U16, port);
+mSCRIPT_DECLARE_STRUCT_METHOD_WITH_DEFAULTS(mScriptSocket, S32, listen, _mScriptSocketListen, 1, U32, queueLength);
+mSCRIPT_DECLARE_STRUCT_METHOD(mScriptSocket, S32, send, _mScriptSocketSend, 1, STR, data);
+mSCRIPT_DECLARE_STRUCT_METHOD(mScriptSocket, WSTR, recv, _mScriptSocketRecv, 1, U32, maxBytes);
+mSCRIPT_DECLARE_STRUCT_METHOD(mScriptSocket, S32, select, _mScriptSocketSelectOne, 1, S64, timeoutMillis);
+
+mSCRIPT_DEFINE_STRUCT(mScriptSocket)
+	mSCRIPT_DEFINE_CLASS_DOCSTRING("TODO")
```
endrift, line 145: Should probably fill these in before merging.

src/script/socket.c
ahigerd: Largely, yes. It would be better to collect all of the active sockets and select on them in one go instead of having one select() call per active socket. It's not the worst thing in the world because I wouldn't expect one of these scripts to have a lot of sockets open, of course.

src/script/socket.c
ahigerd: It's convention from LuaSocket; they have to be exact strings because script code has to check for them.

src/script/socket.c
ahigerd: True.

ahigerd:
@endrift By the way, I noticed that resetting the scripting state leaves any open socket resources orphaned. We should probably figure something out for that...

Also we should discuss whether or not we want SO_REUSEADDR.

src/script/socket.c
ahigerd: Updated it.

src/script/socket.c
endrift: I wonder if the textual descriptions should be moved into the Lua implementation then, not the generic implementation that Lua sits on top of.

src/script/engines/lua.c
```diff
@@ -177,6 +325,20 @@ struct mScriptEngineContext* _luaCreate(struct mScriptEngine2* engine, struct mS
...
 	luaContext->require = luaL_ref(luaContext->lua, LUA_REGISTRYINDEX);
 
+	int status = luaL_dostring(luaContext->lua, _socketLuaSource);
+	if (status) {
+		mLOG(SCRIPT, ERROR, "Error in dostring while initializing sockets: %s\n", lua_tostring(luaContext->lua, -1));
+		lua_pop(luaContext->lua, 1);
+	} else {
+		lua_getglobal(luaContext->lua, "socket");
+		lua_getfield(luaContext->lua, -1, "ERRORS");
+		for (int i = 0; i < _mScriptSocketNumErrors; i++) {
```
endrift, line 335: No definitions in `for` loops, though it's technically allowed in C99 I don't do it in mGBA code. I may consider changing this at some point though, especially now that I'm starting to move mGBA to C11.

src/script/socket.c
```diff
@@ -0,0 +1,275 @@
...
+#endif
+};
+static const int _mScriptSocketNumErrorMappings = sizeof(_mScriptSocketErrorMappings) / sizeof(struct _mScriptSocketErrorMapping);
+
+static void _mScriptSocketSetError(struct mScriptSocket* ssock, int32_t err) {
+	if (!err) {
+		ssock->error = SOCKET_OK;
+		return;
+	}
+	for (int i = 0; i < _mScriptSocketNumErrorMappings; i++) {
```
endrift, line 67: Ditto.

include/mgba/internal/script/socket.h
```diff
@@ -0,0 +1,25 @@
...
+	SOCKET_ADDRESS_IN_USE,
+	SOCKET_CONNECTION_REFUSED,
+	SOCKET_DENIED,
+	SOCKET_FAILED,
+	SOCKET_NETWORK_UNREACHABLE,
+	SOCKET_NOT_FOUND,
+	SOCKET_NO_DATA,
+	SOCKET_OUT_OF_MEMORY,
+	SOCKET_TIMEOUT,
+	SOCKET_UNSUPPORTED,
```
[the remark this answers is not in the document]
ahigerd: I know that's the typical convention and that's what I did the first time, but then I discovered that no-`m` is the convention for namespaced constants exposed to the scripting engine.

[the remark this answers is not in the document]
ahigerd: Can change it to `SOCKERR` though.

src/script/socket.c
```diff
@@ -0,0 +1,275 @@
...
+		mSCRIPT_CONSTANT_PAIR(SOCKET, CONNECTION_REFUSED),
+		mSCRIPT_CONSTANT_PAIR(SOCKET, DENIED),
+		mSCRIPT_CONSTANT_PAIR(SOCKET, FAILED),
+		mSCRIPT_CONSTANT_PAIR(SOCKET, NETWORK_UNREACHABLE),
+		mSCRIPT_CONSTANT_PAIR(SOCKET, NOT_FOUND),
+		mSCRIPT_CONSTANT_PAIR(SOCKET, NO_DATA),
+		mSCRIPT_CONSTANT_PAIR(SOCKET, OUT_OF_MEMORY),
+		mSCRIPT_CONSTANT_PAIR(SOCKET, TIMEOUT),
+		mSCRIPT_CONSTANT_PAIR(SOCKET, UNSUPPORTED),
+		mSCRIPT_KV_SENTINEL
```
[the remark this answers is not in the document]
ahigerd: Whoops. Good catch.

src/script/socket.c
```diff
@@ -0,0 +1,275 @@
...
+};
+static const int _mScriptSocketNumErrorMappings = sizeof(_mScriptSocketErrorMappings) / sizeof(struct _mScriptSocketErrorMapping);
+
+static void _mScriptSocketSetError(struct mScriptSocket* ssock, int32_t err) {
+	if (!err) {
+		ssock->error = SOCKET_OK;
+		return;
+	}
+	for (int i = 0; i < _mScriptSocketNumErrorMappings; i++) {
+		if (_mScriptSocketErrorMappings[i].nativeError) {
```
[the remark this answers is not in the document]
ahigerd: ...... I could have sworn I typed something there. *facepalm*

src/script/engines/lua.c
```diff
@@ -5,6 +5,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 #include <mgba/internal/script/lua.h>
 
+#include <mgba/core/core.h>
```
endrift, line 8: What uses `core/core.h`?

src/script/engines/lua.c
```diff
@@ -17,6 +19,8 @@
 
 #define MAX_KEY_SIZE 128
 
+mLOG_DECLARE_CATEGORY(SCRIPT);
```
endrift, line 22: This is already done in `core/scripting.h`--though it should maybe be moved to `script/context.h` if you're using it here (there's a dependency graph between components, but I haven't actually explicitly defined it anywhere)

include/mgba/internal/script/socket.h
```diff
@@ -0,0 +1,25 @@
...
+	SOCKERR_ADDRESS_IN_USE,
+	SOCKERR_CONNECTION_REFUSED,
+	SOCKERR_DENIED,
+	SOCKERR_FAILED,
+	SOCKERR_NETWORK_UNREACHABLE,
+	SOCKERR_NOT_FOUND,
+	SOCKERR_NO_DATA,
+	SOCKERR_OUT_OF_MEMORY,
+	SOCKERR_TIMEOUT,
+	SOCKERR_UNSUPPORTED,
```
endrift, line 22: Though they're internal only so it probably doesn't matter that much, you can take a look at the PLATFORM exports to see how to export a constant that has a different name from its exported namespace.

endrift (commented):
A few last comments before merging.

src/script/engines/lua.c
ahigerd: Nothing, anymore, because I decided to hook the frame callback in Lua code instead of in C code. I forgot about the include.

include/mgba/internal/script/socket.h
ahigerd: Done

src/script/engines/lua.c
ahigerd: Moved.

ahigerd:
I switched the example scripts over to use `socket.ERRORS` instead of literal strings. As a quality-of-life tweak I also made the `socket.ERRORS` table respond to both numeric and string keys.

include/mgba/script/context.h
```diff
@@ -8,6 +8,8 @@
 
 #include <mgba-util/common.h>
 
+#include <mgba/core/log.h>
```
endrift, line 11: this goes after the CXX_GUARD

src/script/engines/lua.c
endrift: You should remove it here then.

src/script/socket.c
```diff
@@ -0,0 +1,277 @@
+/* Copyright (c) 2013-2022 Jeffrey Pfau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include <mgba/script/context.h>
+
+#include <mgba/core/core.h>
```
endrift, line 8: If tihs isn't used you should remove it.

endrift (changes requested):
Looks like some things only got half-addressed.

ahigerd:
The heck? I did that. O.o Did I manage to miss committing something?

endrift:
I'm ready to merge this unless you have any last-minute changes you want to make.

ahigerd:
No complaints from me!

PR #2620, opened 2022-08-11
Debugger: Add range watchpoints

bigfarts:
These are accessible via the following new CLI debugger commands:

 - rw: watchr minAddr maxAddr [cond]
 - r: watchr/r minAddr maxAddr [cond]
 - w: watchr/w minAddr maxAddr [cond]
 - c: watchr/c minAddr maxAddr [cond]

This also makes all watchpoints range watchpoints under the hood.

Preliminary benchmark results:

Time taken to run 10000 frames of Megaman Battle Network 1 (U) with a write watchpoint set at 0x02000000 in milliseconds, 10 runs each:

control (no watchpoint):
[4184, 4185, 4197, 4207, 4220, 4178, 4304, 4226, 4234, 4292]
mean = 4223, stdev = 43.95

old (single address watchpoint):
[4743, 4685, 4679, 4670, 4782, 4704, 4698, 4875, 4746, 4718]
mean = 4730, stdev = 61.67

new (range watchpoint):
[4683, 4691, 4693, 4706, 4782, 4674, 4746, 4768, 4770, 4776]
mean = 4728, stdev = 43.36

### DIFF
--- a/src/arm/debugger/memory-debugger.c
+++ b/src/arm/debugger/memory-debugger.c
@@ -92,12 +92,13 @@ CREATE_MULTIPLE_WATCHPOINT_SHIM(storeMultiple, WATCHPOINT_WRITE)
 CREATE_SHIM(setActiveRegion, void, (struct ARMCore* cpu, uint32_t address), address)
 
 static bool _checkWatchpoints(struct ARMDebugger* debugger, uint32_t address, struct mDebuggerEntryInfo* info, enum mWatchpointType type, uint32_t newValue, int width) {
-	--width;
 	struct mWatchpoint* watchpoint;
 	size_t i;
+	uint32_t minAddress = address & ~(width - 1);
+	uint32_t maxAddress = minAddress + width;
 	for (i = 0; i < mWatchpointListSize(&debugger->watchpoints); ++i) {
 		watchpoint = mWatchpointListGetPointer(&debugger->watchpoints, i);
-		if (!((watchpoint->address ^ address) & ~width) && watchpoint->type & type) {
+		if (watchpoint->type & type && watchpoint->minAddress < maxAddress && minAddress < watchpoint->maxAddress) {
 			if (watchpoint->condition) {
 				int32_t value;
 				int segment;
@@ -107,7 +108,7 @@ static bool _checkWatchpoints(struct ARMDebugger* debugger, uint32_t address, st
 			}
 
 			uint32_t oldValue;
-			switch (width + 1) {
+			switch (width) {
 			case 1:
 				oldValue = debugger->originalMemory.load8(debugger->cpu, address, 0);
 				break;
--- a/src/debugger/cli-debugger.c
+++ b/src/debugger/cli-debugger.c
@@ -60,6 +60,10 @@ static void _setReadWriteWatchpoint(struct CLIDebugger*, struct CLIDebugVector*)
 static void _setReadWatchpoint(struct CLIDebugger*, struct CLIDebugVector*);
 static void _setWriteWatchpoint(struct CLIDebugger*, struct CLIDebugVector*);
 static void _setWriteChangedWatchpoint(struct CLIDebugger*, struct CLIDebugVector*);
+static void _setReadWriteRangeWatchpoint(struct CLIDebugger*, struct CLIDebugVector*);
+static void _setReadRangeWatchpoint(struct CLIDebugger*, struct CLIDebugVector*);
+static void _setWriteRangeWatchpoint(struct CLIDebugger*, struct CLIDebugVector*);
+static void _setWriteChangedRangeWatchpoint(struct CLIDebugger*, struct CLIDebugVector*);
 static void _listWatchpoints(struct CLIDebugger*, struct CLIDebugVector*);
 static void _trace(struct CLIDebugger*, struct CLIDebugVector*);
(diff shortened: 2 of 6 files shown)


endrift:
Can you rebase this on master and then add this to the changelog under features?

"Debugger: Add range watchpoints"

src/debugger/cli-debugger.c
```diff
@@ -114,6 +118,10 @@ static struct CLIDebuggerCommandSummary _debuggerCommands[] = {
 	{ "watch/c", _setWriteChangedWatchpoint, "Is", "Set a change watchpoint" },
 	{ "watch/r", _setReadWatchpoint, "Is", "Set a read watchpoint" },
 	{ "watch/w", _setWriteWatchpoint, "Is", "Set a write watchpoint" },
+	{ "watchr", _setReadWriteRangeWatchpoint, "IIs", "Set a range watchpoint" },
+	{ "watchr/c", _setWriteChangedRangeWatchpoint, "IIs", "Set a change range watchpoint" },
+	{ "watchr/r", _setReadRangeWatchpoint, "IIs", "Set a read range watchpoint" },
+	{ "watchr/w", _setWriteRangeWatchpoint, "IIs", "Set a write range watchpoint" },
```
endrift, line 124: `watchr` is not self-documenting. Perhaps the more verbose name `watch-range` and add `watchr` as aliases in the alias table below.

src/arm/debugger/memory-debugger.c
```diff
@@ -91,13 +91,13 @@ CREATE_MULTIPLE_WATCHPOINT_SHIM(loadMultiple, WATCHPOINT_READ)
 CREATE_MULTIPLE_WATCHPOINT_SHIM(storeMultiple, WATCHPOINT_WRITE)
 CREATE_SHIM(setActiveRegion, void, (struct ARMCore* cpu, uint32_t address), address)
 
-static bool _checkWatchpoints(struct ARMDebugger* debugger, uint32_t address, struct mDebuggerEntryInfo* info, enum mWatchpointType type, uint32_t newValue, int width) {
-	--width;
+static bool _checkWatchpoints(struct ARMDebugger* debugger, uint32_t minAddress, struct mDebuggerEntryInfo* info, enum mWatchpointType type, uint32_t newValue, int width) {
```
endrift, line 94: I'm not really fond of renaming `address` to `minAddress`, especially since it's not even accurate once you factor in alignment. GBA doesn't have misaligned loads, so addressing a 32-bit value at 0x2 will start access at 0x0 and end at 0x3, not start at 0x2 and end at 0x5. Having a `minAddress` is fine, but it's not equivalent to `address` with misaligned loads. Not to mention the fact that it causes unnecessary and confusing code churn later in the function.

src/debugger/cli-debugger.c
```diff
@@ -717,9 +784,17 @@ static void _listWatchpoints(struct CLIDebugger* debugger, struct CLIDebugVector
 	for (i = 0; i < mWatchpointListSize(&watchpoints); ++i) {
 		struct mWatchpoint* watchpoint = mWatchpointListGetPointer(&watchpoints, i);
 		if (watchpoint->segment >= 0) {
-			debugger->backend->printf(debugger->backend, "%" PRIz "i: %02X:%X\n", watchpoint->id, watchpoint->segment, watchpoint->address);
+			if (watchpoint->maxAddress == watchpoint->minAddress + 1) {
+				debugger->backend->printf(debugger->backend, "%" PRIz "i: %02X:%X\n", watchpoint->id, watchpoint->segment, watchpoint->minAddress);
+			} else {
+				debugger->backend->printf(debugger->backend, "%" PRIz "i: [%02X:%X,%02X:%X)\n", watchpoint->id, watchpoint->segment, watchpoint->minAddress, watchpoint->segment, watchpoint->maxAddress);
```
endrift, line 790: Even though this notation is _technically correct_, people who aren't familiar with bounded and unbounded interval notation will find this confusing. And I expect that will be a significant percentage of the users using this feature. Not to mention the fact that the segment is the same on both. Perhaps just `%02X:%X-%X`, even though it's kinda wishy washy.

endrift (changes requested):
Some small issues, though all but one of them is nitpicking. Now that 0.10 is out though this can be merged once they're resolved.

src/arm/debugger/memory-debugger.c
```diff
@@ -92,12 +92,13 @@ CREATE_MULTIPLE_WATCHPOINT_SHIM(storeMultiple, WATCHPOINT_WRITE)
 CREATE_SHIM(setActiveRegion, void, (struct ARMCore* cpu, uint32_t address), address)
 
 static bool _checkWatchpoints(struct ARMDebugger* debugger, uint32_t address, struct mDebuggerEntryInfo* info, enum mWatchpointType type, uint32_t newValue, int width) {
-	--width;
 	struct mWatchpoint* watchpoint;
 	size_t i;
+	uint32_t minAddress = address / width * width;
```
endrift, line 97: oh god no division in hot code

It's always going to be a power of 2, so you can -1 to get the mask:

```C
address & ~(width - 1)
```

src/arm/debugger/memory-debugger.c
endrift: Yes, at least in the case of the GBA. But we don't care about platforms that support unaligned loads here.

endrift:
Thanks :+1:

PR #2648, opened 2022-09-01
Use reverse DNS for LInux data files

JakobDev:
According to the [official specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html), the file name of the .desktop file should be reverse DNS. Same goes for the icon files, but I couldn't find this part of the spec now. With this PR, the files gets installed under the correct name.


endrift:
Can you use `${PROJECT_NAME}` instead of `mGBA`? It'd be good for when medusa comes along.

JakobDev:
@endrift Done

PR #2661, opened 2022-09-19
hle-bios.s: Make initial registers match BIOS state

easyaspi314:
The GBA bios uses `lr` to branch to the entrypoint, not `r0`.


endrift (approved):
This looks like it _should_ be fine, but changing the BIOS at all can cause issues for save states. I'll need to test it before merging.

PR #2714, opened 2022-11-04
GBA: Verify ELF entrypoint against ROM header

LunarLambda:
Detect more ELF binaries as valid by checking if the ROM/Multiboot entry branches in the header branch to `e_entry`.

We do this by locating the program header containing the entry branch address, decoding it, checking if it's a branch or branch-with-link instruction, and checking if the branch's target address is `e_entry`.

Should fix detection of Multiboot ROMs built with devkitARM and gba-toolchain.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,10 +12,10 @@ Other fixes:
  - VFS: Fix minizip write returning 0 on success instead of size
 Misc:
  - GB Serialize: Add missing savestate support for MBC6 and NT (newer)
+ - GBA: Improve detection of valid ELF ROMs
+ - Qt: Keep track of current pslette preset name (fixes mgba.io/i/2680)
  - macOS: Add category to plist (closes mgba.io/i/2691)
  - macOS: Fix modern build with libepoxy (fixes mgba.io/i/2700)
- - Qt: Keep track of current pslette preset name (fixes mgba.io/i/2680)
- - GBA: Improve detection of valid ELF ROMs
 
 0.10.0: (2022-10-11)
 Features:
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,7 @@ Misc:
  - macOS: Add category to plist (closes mgba.io/i/2691)
  - macOS: Fix modern build with libepoxy (fixes mgba.io/i/2700)
  - Qt: Keep track of current pslette preset name (fixes mgba.io/i/2680)
+ - GBA: Improve detection of valid ELF ROMs
 
 0.10.0: (2022-10-11)
 Features:
--- a/CHANGES
+++ b/CHANGES
@@ -13,9 +13,9 @@ Other fixes:
 Misc:
  - GB Serialize: Add missing savestate support for MBC6 and NT (newer)
  - GBA: Improve detection of valid ELF ROMs
- - Qt: Keep track of current pslette preset name (fixes mgba.io/i/2680)
  - macOS: Add category to plist (closes mgba.io/i/2691)
  - macOS: Fix modern build with libepoxy (fixes mgba.io/i/2700)
+ - Qt: Keep track of current pslette preset name (fixes mgba.io/i/2680)
 
 0.10.0: (2022-10-11)
 Features:
--- a/include/mgba/internal/gba/gba.h
+++ b/include/mgba/internal/gba/gba.h
@@ -55,10 +55,6 @@ struct GBA;
 struct Patch;
 struct VFile;
 
-#ifdef USE_ELF
-struct ELF;
-#endif
-
 mLOG_DECLARE_CATEGORY(GBA);
 mLOG_DECLARE_CATEGORY(GBA_DEBUG);
 
@@ -159,6 +155,8 @@ void GBAStop(struct GBA* gba);
 void GBADebug(struct GBA* gba, uint16_t value);
 
 #ifdef USE_ELF
+struct ELF;
+
 bool GBAVerifyELFEntry(struct ELF* elf, uint32_t target);
 #endif
 
--- a/include/mgba/internal/gba/gba.h
+++ b/include/mgba/internal/gba/gba.h
@@ -19,10 +19,6 @@ CXX_GUARD_START
 #include <mgba/internal/gba/sio.h>
 #include <mgba/internal/gba/timer.h>
 
-#ifdef USE_ELF
-#include <mgba-util/elf-read.h>
-#endif
-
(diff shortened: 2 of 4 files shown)


src/gba/gba.c
```diff
@@ -595,6 +595,58 @@ void GBADebug(struct GBA* gba, uint16_t flags) {
...
+		Elf32_Phdr* phdr = ELFProgramHeadersGetPointer(&ph, i);
+		if (!phdr->p_filesz) {
+			continue;
+		}
+
+		size_t phdr_s = phdr->p_paddr;
+		size_t phdr_e = phdr_s + phdr->p_filesz;
+
+		// Does the segment contain our target address?
+		if ((target >= phdr_s) && ((target + 4) <= phdr_e)) {
```
endrift, line 618: You can reduce the indent level by inverting this check and making it a continue instead.

src/gba/gba.c
```diff
@@ -595,6 +595,58 @@ void GBADebug(struct GBA* gba, uint16_t flags) {
...
+			size_t esize;
+			char* bytes = ELFBytes(elf, &esize);
```
endrift, line 620:
```suggestion
			const char* bytes = ELFBytes(elf, &esize);
```

src/gba/gba.c
```diff
@@ -595,6 +595,58 @@ void GBADebug(struct GBA* gba, uint16_t flags) {
...
+
+			// File offset to what should be the rom entry instruction
+			size_t off = phdr->p_offset + (target - phdr_s);
```
endrift, line 623: The parens here don't affect the result and don't help with readability.

```suggestion
			size_t off = phdr->p_offset + target - phdr_s;
```

src/gba/gba.c
```diff
@@ -595,6 +595,58 @@ void GBADebug(struct GBA* gba, uint16_t flags) {
...
+			if (off >= esize) {
+				continue;
+			}
+
+			uint32_t opcode;
+			LOAD_32(opcode, 0, &bytes[off]);
+			struct ARMInstructionInfo info;
+			ARMDecodeARM(opcode, &info);
+
+			if ((info.branchType == ARM_BRANCH) || (info.branchType == ARM_BRANCH_LINKED)) {
```
endrift, line 634: See above about reducing the indent level.

src/gba/gba.c
```diff
@@ -595,6 +595,58 @@ void GBADebug(struct GBA* gba, uint16_t flags) {
...
+				uint32_t b_target = target + (info.op1.immediate + 8);
```
endrift, line 635: See above about paren usage.
```suggestion
				uint32_t b_target = target + info.op1.immediate + 8;
```

endrift (changes requested):
Should be good to merge after some nit cleanup. Would probably be a good idea to add a CHANGES entry too (under the misc section)

LunarLambda:
:+1:

CHANGES
```diff
@@ -15,6 +15,7 @@ Misc:
  - macOS: Add category to plist (closes mgba.io/i/2691)
  - macOS: Fix modern build with libepoxy (fixes mgba.io/i/2700)
  - Qt: Keep track of current pslette preset name (fixes mgba.io/i/2680)
+ - GBA: Improve detection of valid ELF ROMs
```
endrift, line 18: Please keep these sections alphabetized by category.

include/mgba/internal/gba/gba.h
```diff
@@ -55,6 +55,10 @@ struct GBA;
 struct Patch;
 struct VFile;
 
+#ifdef USE_ELF
+struct ELF;
```
endrift, line 59: This belongs in the ifdef below, right above the function declaration.

src/gba/gba.c
```diff
@@ -595,6 +595,62 @@ void GBADebug(struct GBA* gba, uint16_t flags) {
...
+		Elf32_Phdr* phdr = ELFProgramHeadersGetPointer(&ph, i);
+		if (!phdr->p_filesz) {
+			continue;
+		}
+
+		size_t phdr_s = phdr->p_paddr;
+		size_t phdr_e = phdr_s + phdr->p_filesz;
+
+		// Does the segment contain our target address?
+		if ((target < phdr_s) || ((target + 4) > phdr_e)) {
```
endrift, line 618:
```suggestion
		if (target < phdr_s || target + 4 > phdr_e) {
```
That sure was a bunch of parens. Same reasoning as the previous nit cleanup; sorry I missed it that time.

src/gba/gba.c
```diff
@@ -595,6 +595,62 @@ void GBADebug(struct GBA* gba, uint16_t flags) {
...
+		if (off >= esize) {
+			continue;
+		}
+
+		uint32_t opcode;
+		LOAD_32(opcode, 0, &bytes[off]);
+		struct ARMInstructionInfo info;
+		ARMDecodeARM(opcode, &info);
+
+		if ((info.branchType != ARM_BRANCH) && (info.branchType != ARM_BRANCH_LINKED)) {
```
endrift, line 637:
```suggestion
		if (info.branchType != ARM_BRANCH && info.branchType != ARM_BRANCH_LINKED) {
```
Same logic as above.

endrift (changes requested):
Some minor corrections to the last pass, as well as two paren nits I missed before. There shouldn't be any nits after this since I did a full pass to make sure I didn't miss any other paren style nits.

Generally speaking, variable names in mGBA are `camelCase` not `snake_case`, but I won't hold up this PR on naming those given these specific names (only one letter on one of the two "words"), but if you wanna fix those too I won't complain.

src/gba/gba.c
```diff
@@ -595,6 +595,62 @@ void GBADebug(struct GBA* gba, uint16_t flags) {
...
+
+		size_t esize;
+		const char* bytes = ELFBytes(elf, &esize);
+
+		if (off >= esize) {
+			continue;
+		}
+
+		uint32_t opcode;
+		LOAD_32(opcode, 0, &bytes[off]);
```
endrift, line 633:
```suggestion
		LOAD_32(opcode, off, bytes);
```
That's what that middle field is for :P

That said, an additional alignment check is a good idea, as unaligned 4-byte loads are unspecified on ARM7TDMI (and the way it actually works is weird and not worth emulating here--we can just reject an unaligned address out of hand since that would break in actual use(

endrift (approved):

PR #2721, opened 2022-11-11
Ability to load shader on launch from command line

danaYatsuta:
Using "--shader" option followed by shader folder, that shader can be loaded on launch.

My personal reason for implementing this was to launch some games with gba-color.shader and some not depending on whether the game oversaturates colors to counter original GBA's dark screen or not, but it may have other uses.

### DIFF
--- a/include/mgba/feature/commandline.h
+++ b/include/mgba/feature/commandline.h
@@ -20,6 +20,7 @@ struct mArguments {
 	char* cheatsFile;
 	char* savestate;
 	char* bios;
+	char* shader;
 	int logLevel;
 	int frameskip;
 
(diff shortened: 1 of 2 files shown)


endrift:
Since the SDL fronted doesn't support shaders yet, I should decide if I want to add support for that before merging this, or ask you to move the command line setting to the Qt subparser. The former probably makes more sense.

endrift:
I already took care of this on the Qt side, so this is obsolete now. Adding shader support to the SDL frontend is pending still, but it doesn't make sense to add shaders to the shared CLI parser since that's used for a bunch of things that don't support shaders too.

PR #2762, opened 2022-12-23
Qt: Add support to resize settings dialog vertically

momocaoo:
On a Steam Deck the settings dialog is, by default, taller than the screen. This makes it very difficult to adjust the settings as the dialog is not resizable and you can't get to the Apply button.

This change wraps the layout in a scroll area, letting you scroll over the content and resize the window vertically. I had to adjust the width of the QStackedWidget to a fixed size for this to work well, maybe someone that knows Qt better has a cleaner solution.

## before
Screenshot_20221223_200226

## after
Screenshot_20221223_200317

(I had to test with a vm so the resolution is smaller).

### DIFF
--- a/src/platform/qt/SettingsView.ui
+++ b/src/platform/qt/SettingsView.ui
@@ -10,8 +10,20 @@
     <height>797</height>
    </rect>
   </property>
+  <property name="minimumSize">
+   <size>
+    <width>885</width>
+    <height>200</height>
+   </size>
+  </property>
+  <property name="maximumSize">
+   <size>
+    <width>885</width>
+    <height>797</height>
+   </size>
+  </property>
   <property name="sizePolicy">
-   <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
+   <sizepolicy hsizetype="Fixed" vsizetype="Preferred">
     <horstretch>0</horstretch>
     <verstretch>0</verstretch>
    </sizepolicy>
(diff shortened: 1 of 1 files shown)


endrift:
I don't really want all of the pages to be scrollable--some already are (e.g. logging and shortcuts). It'd be better to pick and choose some. It's worth noting I've made the Game Boy page shorter in part to reduce the overall height.

momocaoo:
Alright. I totally understand. I will compile main on my Steam deck and see if it fits, maybe you already fixed it :)

endrift:
It's weird because my local compilation fits fine, but the Flatpak doesn't. Flatpak seems to have DPI issues.

momocaoo:
Could this be related? https://github.com/flatpak/flatpak/issues/2980

ahigerd:
@momocaoo To answer your question about Qt: Putting a QStackedWidget inside a QScrollWidget is the wrong way to do it for two reasons. The more obvious reason is that it breaks the geometry calculation, as you observed -- within a scroll area, there's no container to use for a size reference.

The less obvious reason is that it means that the scroll position is independent of the widget selection, so switching tabs will leave the interface scrolled down. Users expect the tabs to be the top-level component with scroll areas inside being independent of each other.

Within a QScrollWidget, the property you'd be looking for is `setWidgetResizable` in order to make it fit the contents to the window and only scroll if it's forced to. This will have the side effect of stretching the layout vertically if it's smaller than the window, so you would then need to add a stretch to the bottom of the layout to fill space.

endrift:
> Could this be related? [flatpak/flatpak#2980](https://github.com/flatpak/flatpak/issues/2980)

I'm not sure, but it's possible. I think the change I made to make the window slightly shorter was merged into 0.10.1; can you check if that works?

endrift:
Closing since this wasn't really the right approach, and I did end up making the window shorter in some capacity in the meantime.

PR #2839, opened 2023-02-24
Libretro: Fix undeclared constant

adamantike:
The `SIZE_CART_FLASH1M` constant was renamed to `GBA_SIZE_FLASH1M` in https://github.com/mgba-emu/mgba/commit/8545271e9ee275b9e733bbfa13195365b1fb82e1

These leftovers make the Libretro build fail, when running:
```
cmake -DBUILD_LIBRETRO=ON .. && make
```


endrift:
Huh, I thought I checked for all the constants that changed. Does this fix the build alone, or are there other remnants left over that need fixing too?

adamantike:
That was the only change I needed for it to correctly build!

endrift:
Good to know I didn't miss more :) Thanks for finding this.

adamantike:
Thank you! I can confirm it builds correctly now, and also fixes the installation of [`libretro-mgba-git`](https://aur.archlinux.org/packages/libretro-mgba-git) from the AUR.

PR #2882, opened 2023-04-12
Scripting: shim print()/warn() in Lua

ahigerd:


endrift:
At least for now you should manually tostring the input. This might be able to be removed later, but until then it's a good idea to make it work more generally.

src/script/engines/lua.c
```diff
@@ -1519,3 +1528,35 @@ static int _luaRequireShim(lua_State* lua) {
...
+
+	// The first upvalue is either "log" or "warn"
+	lua_getglobal(lua, "console");
+	lua_pushvalue(lua, lua_upvalueindex(1));
+	lua_gettable(lua, -2);
+
+	lua_insert(lua, 1);
+	lua_pop(lua, 1);
+
+	// If console:log is variadic and stringifies by itself:
```
endrift, line 1546:
```suggestion
	// TODO when console:log is variadic and stringifies by itself:
```
I actually do search the codebase for `TODO` sometimes.

ahigerd:
Fixed the comments, also made it more thorough about stringification.

PR #2883, opened 2023-04-14
Debugger: Send flags information for cpsr register

DavidSpickett:
There is a feature of target XML called flags. It allows you to describe what a register contains.

https://sourceware.org/gdb/onlinedocs/gdb/Target-Description-Format.html

GDB has supported this for a long time and I recently added support in LLDB:
https://github.com/llvm/llvm-project/commit/e07a421dd587f596b3b34ac2f79081402089f878

This change adds this flags information for the cpsr register of the ARM7TDMI. Based on the information in https://developer.arm.com/documentation/ddi0210/c/.

This is what it looks like when using GDB:
```
(gdb) info registers
r0             0x0                 0
<...>
cpsr           0x6000001f          [ Z C M=31 ]
```
And LLDB:
```
(lldb) register read cpsr
    cpsr = 0x6000001f
         = (N = 0, Z = 1, C = 1, V = 0, I = 0, F = 0, T = 0, M=31)
```

(the format is up to the debugger, lldb is a lot more verbose at the moment)

To enable this I have increased the GDB stub's outgoing buffer to 1400 bytes. The target XML is just above 130 bytes with the flags added.


src/debugger/gdb-stub.c
```diff
@@ -50,7 +50,21 @@ static const char* TARGET_XML = "<target version=\"1.0\">"
...
+                                "<field name=\"C\" start=\"29\" end=\"29\"/>"
+                                "<field name=\"V\" start=\"28\" end=\"28\"/>"
+                                "<field name=\"I\" start=\"7\" end=\"7\"/>"
+                                "<field name=\"F\" start=\"6\" end=\"6\"/>"
+                                "<field name=\"T\" start=\"5\" end=\"5\"/>"
+                                "<field name=\"M4\" start=\"4\" end=\"4\"/>"
+                                "<field name=\"M3\" start=\"3\" end=\"3\"/>"
+                                "<field name=\"M2\" start=\"2\" end=\"2\"/>"
+                                "<field name=\"M1\" start=\"1\" end=\"1\"/>"
+                                "<field name=\"M0\" start=\"0\" end=\"0\"/>"
```
endrift, line 65: Why is this split out? The M field functions as a single 5 bit value (or more accurately a single 4 bit value and an always-1 5th bit).

DavidSpickett: Because I read the documentation too literally. I've combined it into one "M" field.

DavidSpickett:
Thanks for the quick response!

However I think I have found a GDB bug while doing this, it prints this for the combined field:
```
cpsr           0x1f                [ M=0 ]
```
Marking this as draft while I track that down.

DavidSpickett:
The bug is https://sourceware.org/bugzilla/show_bug.cgi?id=28103, which was fixed in GDB-11.

For some reference, to be able to `apt install` that on Ubuntu you'd need to be using 22.04 (Jammy) which came out April 2022. Which is quite recent, and using an older version would give misleading results for "M" which we don't want.

I could do add the condition flags. They will be correct even with an older GDB and it is still an improvement over the raw value. What do you think?

Personally I wouldn't think this meant there was no M field at all, just that we weren't specifically showing it:
```
cpsr           0x6000001f          [ Z C ]
```
But I am completely biased on that because I know how the internals work.

endrift:
I think it's better to merge it as-is and let people upgrade gdb as time goes on, instead of having to add it in later.

Issue #1352, opened 2019-03-11
Request: shortcut to turn off cheats


Cibomatto2002:
Could we see shortcut to turn off cheats?

TamTrongNguyen:
Hello, my name is Tam Nguyen and I'm a student at UT taking the virtualization class cs360v. Can I request to work on this issue?

endrift:
This might be appropriate for a beginner issue, but it will probably take a bit more intimate understanding of some of the subsystems involved than may be desired for a beginner issue.

PR #2897, opened 2023-04-25
Add shortcut to disable all cheats

martinmurtiono:
This adds a button/shortcut to disable all cheats

https://github.com/mgba-emu/mgba/issues/1352

^ I am part of Tam's group in our virtualization class.

### DIFF
--- a/src/core/cheats.c
+++ b/src/core/cheats.c
@@ -812,5 +812,4 @@ void mCheatDisableAll(struct mCheatDevice* device) {
 			cheat->enabled = false;
 		}
 	}
-	return;
 }
--- a/src/core/cheats.c
+++ b/src/core/cheats.c
@@ -802,3 +802,15 @@ void mCheatDeviceDeinit(struct mCPUComponent* component) {
 		}
 	}
 }
+
+void mCheatDisableAll(struct mCheatDevice* device) {
+	size_t i;
+	for (i = 0; i < mCheatSetsSize(&device->cheats); ++i) {
+		struct mCheatSet* cheat = *mCheatSetsGetPointer(&device->cheats, i);
+		if (cheat->enabled) {
+			_unpatchROM(device, cheat);
+			cheat->enabled = false;
+		}
+	}
+	return;
+}
--- a/src/platform/qt/Window.cpp
+++ b/src/platform/qt/Window.cpp
@@ -1630,6 +1630,11 @@ void Window::setupMenu(QMenuBar* menubar) {
 	m_actions.addAction(tr("Game Pak sensors..."), "sensorWindow",  openNamedControllerTView<SensorView>(&m_sensorView, &m_inputController), "tools");
 
 	addGameAction(tr("&Cheats..."), "cheatsWindow", openControllerTView<CheatsView>(), "tools");
+	addGameAction(tr("&Disable cheats..."), "disableCheats", [this]() {
+		if (m_controller) {
+			mCheatDisableAll(m_controller->cheatDevice());
+		}
+	}, "tools");
 #ifdef ENABLE_SCRIPTING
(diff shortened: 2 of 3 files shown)


src/platform/qt/Window.cpp
```diff
@@ -1630,6 +1630,11 @@ void Window::setupMenu(QMenuBar* menubar) {
 	m_actions.addAction(tr("Game Pak sensors..."), "sensorWindow",  openNamedControllerTView<SensorView>(&m_sensorView, &m_inputController), "tools");
 
 	addGameAction(tr("&Cheats..."), "cheatsWindow", openControllerTView<CheatsView>(), "tools");
+	addGameAction(tr("&Disable cheats..."), "disableCheats", [this]() {
```
endrift, line 1633:
```suggestion
	addGameAction(tr("Disable all cheats"), "disableCheats", [this]() {
```

The "..." at the end indicates that it opens a new window, which this one doesn't.

src/core/cheats.c
```diff
@@ -802,3 +802,15 @@ void mCheatDeviceDeinit(struct mCPUComponent* component) {
...
+void mCheatDisableAll(struct mCheatDevice* device) {
+	size_t i;
+	for (i = 0; i < mCheatSetsSize(&device->cheats); ++i) {
+		struct mCheatSet* cheat = *mCheatSetsGetPointer(&device->cheats, i);
+		if (cheat->enabled) {
+			_unpatchROM(device, cheat);
+			cheat->enabled = false;
+		}
+	}
+	return;
```
endrift, line 815:
```suggestion
```

This line does nothing, so it should be removed.

endrift (changes requested):
Looks mostly fine, minus some nitpicks. However, I do worry that this cheats window won't notice that they've been disabled under them. Did you test that?

martinmurtiono:
Yes, the checkboxes get unticked for all cheats in the window.

endrift:
Though the original request by @cibomatto2002 was to "disable" cheats I'm unsure if he meant "turn the cheats individually off" (as was done here) or "temporarily disable the cheat subsystem entirely in a way that means if you re-enable it you don't have to manually re-enable the individual cheats you want enabled", and I would assume the latter would be more useful. Perhaps he could chime in.

martinmurtiono:
That's a good point. I might as well just make it a toggle as its more useful that way.

Cibomatto2002:
What I really wanted was a hot key to turn them all off at the same time it would be easier for the person that had a cheat on like infinite time.

martinmurtiono:
The current implementation turns all the cheats off like you described, but it might be more useful if it was a toggle to turn the cheats on/off. In that case, you can simply just press it once to disable all cheats, and for people who want an easy way to re-enable them, they have the option to repress the key and turn the cheats they previously had back on.

endrift:
Wanting to just mark all cheats as disabled instead of a "temporarily pause the cheat subsystem" is an incredibly niche use case and I don't think it's worthwhile. I'm going to close this PR until you get a version that adds the whole subsystem toggle.

martinmurtiono:
I'm not getting what you mean by "temporarily pause the cheat subsystem". Does saving the currently enabled cheats when disabling them then re-enabling the same cheats when you toggle cheats back on achieve this? Or did you have another method in mind?

martinmurtiono:
@endrift  When you mean the whole subsystem, do you mean to temporarily disable the cheat window and any cheats that were enabled? I feel like we can achieve the same function by toggling the most recently enabled cheats on/off. In this case Cibomatto2002 can still have his desired functionality of simply disabling all the cheats, and when you toggle it back on, the most recently enabled cheats will be re-enabled so users don't have to individually re-enable them, and the cheat window will still be available to edit.

Issue #2845, opened 2023-03-03
Add shortcut to reset and re-load the most recent script


40Cakes:
I'm learning LUA for mGBA at the moment and while developing I think it could speed up testing scripts by having a hotkey to reset and re-load the latest script without needing to open Tools > Scripting > File > Load recent script > latest file... every time. It would be especially helpful to quickly re-connect TCP sockets for example

Thank you 😊

sam-gupta-git:
My name is Sam and I am a student in the CS 360V Virtualization class at UT Austin. May I work on this issue?

endrift:
This has absolutely nothing to do with virtualization, but sure.

endrift:
Closed by #3266

PR #2901, opened 2023-04-26
Re-load most recently ran script 

TamTrongNguyen:
Adds a shortcut to re-load the most recently ran script 

https://github.com/mgba-emu/mgba/issues/2845

### DIFF
--- a/src/platform/qt/Window.cpp
+++ b/src/platform/qt/Window.cpp
@@ -1632,11 +1632,6 @@ void Window::setupMenu(QMenuBar* menubar) {
 	addGameAction(tr("&Cheats..."), "cheatsWindow", openControllerTView<CheatsView>(), "tools");
 #ifdef ENABLE_SCRIPTING
 	m_actions.addAction(tr("Scripting..."), "scripting", this, &Window::scriptingOpen, "tools");
-	
-	addGameAction(tr("Load recent script"), "loadscript", [this] {
-		m_mruFiles = m_config->getMRU(ConfigController::MRU::Script);
-		m_scripting->loadFile(m_mruFiles.at(0));
-	}, "tools");
 #endif
 
 	m_actions.addAction(tr("Create forwarder..."), "createForwarder", openTView<ForwarderView>(), "tools");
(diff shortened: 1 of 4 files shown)


src/platform/qt/Window.cpp
```diff
@@ -1632,6 +1632,11 @@ void Window::setupMenu(QMenuBar* menubar) {
 	addGameAction(tr("&Cheats..."), "cheatsWindow", openControllerTView<CheatsView>(), "tools");
 #ifdef ENABLE_SCRIPTING
 	m_actions.addAction(tr("Scripting..."), "scripting", this, &Window::scriptingOpen, "tools");
+	
+	addGameAction(tr("Load recent script"), "loadscript", [this] {
+		m_mruFiles = m_config->getMRU(ConfigController::MRU::Script);
```
endrift, line 1637: This overwrites the member variable `m_mruFiles`, which is _not_ what you want.

endrift (changes requested):
This needs to appear in the scripting window's menu, not the parent window's menu.

src/platform/qt/Window.cpp
TamTrongNguyen: I changed the location of the shortcut to appear in the scripting window but was not able to bind it to any keys. It does not overwrite m_mruFiles and does play the most recently run script in the scripting window

src/platform/qt/scripting/ScriptingView.cpp
```diff
@@ -62,10 +63,15 @@ void ScriptingView::load() {
 		if (!m_controller->loadFile(filename)) {
 			return;
 		}
+		setMRS(filename);
```
endrift, line 66: You don't need to store this separately, it's always going to be `m_mruFiles.at(0)`--unless `m_mruFiles.isEmpty()` is `true`, in which case there is no most recently loaded script.

src/platform/qt/scripting/ScriptingView.h
```diff
@@ -22,6 +22,7 @@ Q_OBJECT
 private slots:
 	void submitRepl();
 	void load();
+	void recent();
```
endrift, line 25: I'd recommend the name `loadRecent` instead, as `recent` isn't a verb and lacks clarity of the action in question

endrift (changes requested):

endrift:
Are you planning to finish this PR now that the semester is (presumably) over?

Issue #2830, opened 2023-02-15
Add shortcuts for increase and decrease fast-forward speed


pmbarrett314:
It would be useful to be able to bind a shortcut for incrementing or decrementing the fast-forward speed. My use-case is using mGBA on Steam Deck or with Steam Controller, where it would be nice to be able to bind the otherwise unused shoulder buttons or back paddles to give fine grained control of the fast-forward setting without having to use a point-and-click menu. For example if I'm fast-forwarding and having trouble moving a cursor to a specific point on a map, I could slow it down slightly so I don't overshoot.

endrift:
I've been tempted to add analog fast-forward support where you can pull an analog trigger a different amount for a different ratio. This is a bit different, but it sounds good too.

TamTrongNguyen:
Hello, my name is Tam Nguyen and I'm a student at UT taking the virtualization class cs360v. Can I request to work on this issue?

endrift:
This seems like a decent first issue, yeah. It has nothing to do with virtualization though.

sam-gupta-git:
^ I am also in Tam's group, and I have added this enhancement here #2903

PR #2903, opened 2023-04-26
Feature: shortcuts to increment fast forward speed

sam-gupta-git:
Adds shortcuts to increment/decrement fast forward speed.

Resolves the following issue:  #2830

### DIFF
--- a/src/platform/qt/Window.cpp
+++ b/src/platform/qt/Window.cpp
@@ -1430,7 +1430,7 @@ void Window::setupMenu(QMenuBar* menubar) {
 	m_config->updateOption("fastForwardRatio");
 
 	addGameAction(tr("Fast forward speed increment"), "fastForwardUp", [this] {
-		float newRatio = m_config->getOption("fastForwardRatio", newRatio).toFloat() + 1.0;
+		float newRatio = m_config->getOption("fastForwardRatio", 1.0f).toFloat() + 1.0f;
 		m_config->setOption("fastForwardRatio", QVariant(newRatio));
 	}, "emu");
 
--- a/src/platform/qt/Window.cpp
+++ b/src/platform/qt/Window.cpp
@@ -1436,7 +1436,7 @@ void Window::setupMenu(QMenuBar* menubar) {
 
 	addGameAction(tr("Fast forward speed decrement"), "fastForwardDown", [this] {
 		float newRatio = m_config->getOption("fastForwardRatio").toFloat() - 1.0f;
-		if (newRatio >= 2.0){
+		if (newRatio >= 2.0) {
 			m_config->setOption("fastForwardRatio", QVariant(newRatio));
 		}
 	}, "emu");
--- a/src/platform/qt/Window.cpp
+++ b/src/platform/qt/Window.cpp
@@ -1429,14 +1429,16 @@ void Window::setupMenu(QMenuBar* menubar) {
 	}
 	m_config->updateOption("fastForwardRatio");
 
-	addGameAction(tr("Fast forward speed increment"), "fastForwardUp", [this] {
+	addGameAction(tr("Increase fast forward speed"), "fastForwardUp", [this] {
 		float newRatio = m_config->getOption("fastForwardRatio", 1.0f).toFloat() + 1.0f;
-		m_config->setOption("fastForwardRatio", QVariant(newRatio));
+		if (newRatio >= 3.0f) {
+			m_config->setOption("fastForwardRatio", QVariant(newRatio));
+		}
 	}, "emu");
 
-	addGameAction(tr("Fast forward speed decrement"), "fastForwardDown", [this] {
+	addGameAction(tr("Decrease fast forward speed"), "fastForwardDown", [this] {
 		float newRatio = m_config->getOption("fastForwardRatio").toFloat() - 1.0f;
-		if (newRatio >= 2.0) {
+		if (newRatio >= 2.0f) {
 			m_config->setOption("fastForwardRatio", QVariant(newRatio));
 		}
 	}, "emu");
--- a/src/platform/qt/Window.cpp
+++ b/src/platform/qt/Window.cpp
@@ -1429,6 +1429,20 @@ void Window::setupMenu(QMenuBar* menubar) {
 	}
 	m_config->updateOption("fastForwardRatio");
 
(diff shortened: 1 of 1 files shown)


src/platform/qt/Window.cpp
```diff
@@ -1429,6 +1429,20 @@ void Window::setupMenu(QMenuBar* menubar) {
 	}
 	m_config->updateOption("fastForwardRatio");
 
+	addGameAction(tr("Fast forward speed increment"), "fastForwardUp", [this] {
+		float oldRatio = 0.0;
+		oldRatio = m_config->getOption("fastForwardRatio", oldRatio).toFloat();
+		float newRatio = oldRatio + 1.0;
+		m_config->setOption("fastForwardRatio", QVariant(newRatio));
```
endrift, line 1436: You can do this in two lines by using only one variable and adding `1.0f` inline. There's no need for two variables.

Also did you test this? I don't know if this would actually work since it doesn't reload the configuration.

src/platform/qt/Window.cpp
```diff
@@ -1429,6 +1429,20 @@ void Window::setupMenu(QMenuBar* menubar) {
...
+	}, "emu");
+
+	addGameAction(tr("Fast forward speed decrement"), "fastForwardDown", [this] {
+		float oldRatio = 0.0;
+		oldRatio = m_config->getOption("fastForwardRatio", oldRatio).toFloat();
+		float newRatio = oldRatio - 1.0;
+		m_config->setOption("fastForwardRatio", QVariant(newRatio));
```
endrift, line 1443: Ditto, but also you need to clamp this to be a `2.0f` minimum.

endrift (changes requested):

src/platform/qt/Window.cpp
sam-gupta-git: I reduced the number of lines to zero. Yes, I tested it and it works.

sam-gupta-git: I reduced the number of lines to two. Yes, I tested it and it works.

src/platform/qt/Window.cpp
```diff
@@ -1429,6 +1429,18 @@ void Window::setupMenu(QMenuBar* menubar) {
 	}
 	m_config->updateOption("fastForwardRatio");
 
+	addGameAction(tr("Fast forward speed increment"), "fastForwardUp", [this] {
+		float newRatio = m_config->getOption("fastForwardRatio", newRatio).toFloat() + 1.0;
```
endrift, line 1433:
```suggestion
		float newRatio = m_config->getOption("fastForwardRatio", 1.0f).toFloat() + 1.0f;
```

src/platform/qt/Window.cpp
sam-gupta-git: Changed it to only use one variable, and added the clamp.

src/platform/qt/Window.cpp
```diff
@@ -1429,6 +1429,18 @@ void Window::setupMenu(QMenuBar* menubar) {
...
+		m_config->setOption("fastForwardRatio", QVariant(newRatio));
+	}, "emu");
+
+	addGameAction(tr("Fast forward speed decrement"), "fastForwardDown", [this] {
+		float newRatio = m_config->getOption("fastForwardRatio", newRatio).toFloat() - 1.0;
```
endrift, line 1438:
```suggestion
		float newRatio = m_config->getOption("fastForwardRatio").toFloat() - 1.0f;
```

src/platform/qt/Window.cpp
```diff
@@ -1429,6 +1429,18 @@ void Window::setupMenu(QMenuBar* menubar) {
...
+		if (newRatio >= 2.0){
```
endrift, line 1439:
```suggestion
		if (newRatio >= 2.0) {
```

endrift (changes requested):

src/platform/qt/Window.cpp
```diff
@@ -1429,6 +1429,18 @@ void Window::setupMenu(QMenuBar* menubar) {
 	}
 	m_config->updateOption("fastForwardRatio");
 
+	addGameAction(tr("Fast forward speed increment"), "fastForwardUp", [this] {
+		float newRatio = m_config->getOption("fastForwardRatio", 1.0f).toFloat() + 1.0f;
+		m_config->setOption("fastForwardRatio", QVariant(newRatio));
```
endrift, line 1434: You also need to clamp this to make sure you're not adding 1 to -1, going from unbounded to 0x, which would break a lot. You can do by checking if the new ratio is at least 3.0f.

src/platform/qt/Window.cpp
```diff
@@ -1429,6 +1429,18 @@ void Window::setupMenu(QMenuBar* menubar) {
 	}
 	m_config->updateOption("fastForwardRatio");
 
+	addGameAction(tr("Fast forward speed increment"), "fastForwardUp", [this] {
```
endrift, line 1432:
```suggestion
	addGameAction(tr("Increase fast forward speed"), "fastForwardUp", [this] {
```

src/platform/qt/Window.cpp
```diff
@@ -1429,6 +1429,18 @@ void Window::setupMenu(QMenuBar* menubar) {
...
+		float newRatio = m_config->getOption("fastForwardRatio", 1.0f).toFloat() + 1.0f;
+		m_config->setOption("fastForwardRatio", QVariant(newRatio));
+	}, "emu");
+
+	addGameAction(tr("Fast forward speed decrement"), "fastForwardDown", [this] {
```
endrift, line 1437:
```suggestion
	addGameAction(tr("Decrease fast forward speed"), "fastForwardDown", [this] {
```

endrift (changes requested):
I was going to merge it and then change the wording afterwards (the old wording is kinda stilted), but I found one more issue. Once this is addressed I can merge it. You might also want to add "Qt: Add shortcuts for increasing and decreasing fast forward speed" or similar to the changelog (Misc section, since it's a minor usability feature).

src/platform/qt/Window.cpp
sam-gupta-git: Added the clamp and changed the wording.

PR #2905, opened 2023-04-27
Add support for exporting the list of memory search results to CSV for post-processing

### LINKED ISSUE #2215: [Suggestions] Allow us to copy a memory searches results
I'm trying to create some post-processing with the memory search.
I could create a script to compare results, but I can't copy the memory search list in any capacity (even an export to CSV would be fine)

danielputexas:
This adds a simple button which allows the user to choose a file, and then writes to it in CSV format the search results table exactly as it is shown.

Fixes #2215 .

### DIFF
--- a/src/platform/qt/MemorySearch.cpp
+++ b/src/platform/qt/MemorySearch.cpp
@@ -11,6 +11,9 @@
 #include "CoreController.h"
 #include "MemoryView.h"
 
+#include <QFileDialog>
+#include <QTextStream>
+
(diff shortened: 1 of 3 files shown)


endrift:
I don't see the logic in presenting less-parsable data, even if it is what's presenting the UI. Verbatim matching is usually _not_ what people want in a CSV.

endrift:
Closing, since I never got a reply as to the logic.

PR #2929, opened 2023-05-21
[UI bug] mGBA doesn't update savestate screenshots until you move the cursor over other savestates

### LINKED ISSUE #2183: [Switch][UI bug] mGBA doesn't update savestate screenshots until you move the cursor over other savestates
Steps to reproduce:
1. Create a savestate in slot #1
2. Do something to change the screen (like moving or opening a menu) and overwrite that savestate

If you open the save state menu again, the screenshot the emulator shows it's from the previous savestate, not the current one. It won't show the actual screenshot until you move the cursor over the other savestates slots (there's no need of loading them, only show their preview images).

Tested in the slightly outdated 0.10-6982-29115 build,  Atmosphère 0.19.2 and firmware 12.0.1

Utagai:
Fixes #2183.

Hi, this is my first PR and @endrift may remember me asking a question or two on the Discord server.

> :warning: I haven't touched C in about... 5 years. And this is my first time making changes to this codebase. Please let me know if I'm doing anything wrong or if I'm making any style mistakes etc.

The change itself is quite small but I'll explain it anyways: extend the `SCREENSHOT_*` enums in the `gui-runner.c` file to include a `SCREENSHOT_STALE`. This is distinct from `SCREENSHOT_INVALID` because we do not want to just draw the current frame onto the background, we want to draw the actual save state screenshot PNG. This is a case that I considered to be equivalent to the `!= SCREENSHOT_INVALID` case in terms of handling, since we basically just want to reload the new PNG image bytes into the background buffer.

_P.S._ `CONTRIBUTING.md` states that my commit messages should be tagged with their relevant components, but I am not familiar enough with this codebase to be know which components I'm touching. I didn't see any components in the document referring to `mGUI`.

### DIFF
--- a/src/feature/gui/gui-runner.c
+++ b/src/feature/gui/gui-runner.c
@@ -658,8 +658,9 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
 				break;
 			case RUNNER_SAVE_STATE:
 				struct mGUIBackground* gbaBackground = (struct mGUIBackground*) stateSaveMenu.background;
-				gbaBackground->screenshotId = (item->data.v.u >> 16) | SCREENSHOT_STALE;
-				mCoreSaveState(runner->core, item->data.v.u >> 16, SAVESTATE_SCREENSHOT | SAVESTATE_SAVEDATA | SAVESTATE_RTC | SAVESTATE_METADATA);
+				unsigned stateId = item->data.v.u >> 16;
+				gbaBackground->screenshotId = stateId | SCREENSHOT_STALE;
+				mCoreSaveState(runner->core, stateId, SAVESTATE_SCREENSHOT | SAVESTATE_SAVEDATA | SAVESTATE_RTC | SAVESTATE_METADATA);
 				break;
 			case RUNNER_LOAD_STATE:
 				mCoreLoadState(runner->core, item->data.v.u >> 16, SAVESTATE_SCREENSHOT | SAVESTATE_RTC);
--- a/src/feature/gui/gui-runner.c
+++ b/src/feature/gui/gui-runner.c
@@ -658,11 +658,10 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
 				break;
 			case RUNNER_SAVE_STATE:
 				struct mGUIBackground* gbaBackground = (struct mGUIBackground*) stateSaveMenu.background;
-				unsigned stateId = item->data.v.u >> 16;
 				// If we are saving state, then the screenshot stored for the state previously should no longer be considered up-to-date.
 				// Therefore, mark it as stale, so that at draw time, we will re-draw to the new save state's screenshot.
-				gbaBackground->screenshotId = stateId | SCREENSHOT_STALE;
-				mCoreSaveState(runner->core, stateId, SAVESTATE_SCREENSHOT | SAVESTATE_SAVEDATA | SAVESTATE_RTC | SAVESTATE_METADATA);
+				gbaBackground->screenshotId |= SCREENSHOT_STALE;
+				mCoreSaveState(runner->core, item->data.v.u >> 16, SAVESTATE_SCREENSHOT | SAVESTATE_SAVEDATA | SAVESTATE_RTC | SAVESTATE_METADATA);
 				break;
 			case RUNNER_LOAD_STATE:
 				mCoreLoadState(runner->core, item->data.v.u >> 16, SAVESTATE_SCREENSHOT | SAVESTATE_RTC);
--- a/src/feature/gui/gui-runner.c
+++ b/src/feature/gui/gui-runner.c
@@ -51,7 +51,6 @@ enum {
 enum {
 	SCREENSHOT_VALID = 0x10000,
 	SCREENSHOT_INVALID = 0x20000,
-	SCREENSHOT_STALE = 0x30000,
 };
 
 static const struct mInputPlatformInfo _mGUIKeyInfo = {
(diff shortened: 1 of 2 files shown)


src/feature/gui/gui-runner.c
```diff
@@ -656,7 +657,10 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
 				runner->core->reset(runner->core);
 				break;
 			case RUNNER_SAVE_STATE:
-				mCoreSaveState(runner->core, item->data.v.u >> 16, SAVESTATE_SCREENSHOT | SAVESTATE_SAVEDATA | SAVESTATE_RTC | SAVESTATE_METADATA);
+				struct mGUIBackground* gbaBackground = (struct mGUIBackground*) stateSaveMenu.background;
+				unsigned stateId = item->data.v.u >> 16;
+				gbaBackground->screenshotId = stateId | SCREENSHOT_STALE;
```
Utagai, line 662: I originally had an `if` check here that would check if the `screenshotId` was referring to the `stateId` in question. However, I believe it is impossible for it not to because we would have [updated `screenshotId`](https://github.com/mgba-emu/mgba/blob/master/src/feature/gui/gui-runner.c#L164-L169) in the draw loop when we [changed state IDs](https://github.com/mgba-emu/mgba/blob/master/src/feature/gui/gui-runner.c#L130)... does this sound correct?

src/feature/gui/gui-runner.c
```diff
@@ -656,7 +657,12 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
 				runner->core->reset(runner->core);
 				break;
 			case RUNNER_SAVE_STATE:
-				mCoreSaveState(runner->core, item->data.v.u >> 16, SAVESTATE_SCREENSHOT | SAVESTATE_SAVEDATA | SAVESTATE_RTC | SAVESTATE_METADATA);
+				struct mGUIBackground* gbaBackground = (struct mGUIBackground*) stateSaveMenu.background;
```
endrift, line 660: Don't define a variable in a case statement without declaring a block (putting the whole thing in `{}`)--I generally just avoid declaring variables if I can avoid it, or hoist the declaration. It'll get optimized anyway.

src/feature/gui/gui-runner.c
```diff
@@ -51,6 +51,7 @@ enum {
 enum {
 	SCREENSHOT_VALID = 0x10000,
 	SCREENSHOT_INVALID = 0x20000,
+	SCREENSHOT_STALE = 0x30000,
```
endrift, line 54: You don't need a new value for this, you can just use INVALID.

src/feature/gui/gui-runner.c
endrift: That does, yeah, in which case you can just |= the value with the flag instead of wholesale assigning it--which avoids the need for the `stateId` variable entirely.

endrift (changes requested):
You also need to update `_tryAutosave` to mark the screenshot as invalid if the ID is 0.

src/feature/gui/gui-runner.c
Utagai: Yeah, I also saw that I could use `INVALID`, especially cause `SCREENSHOT_VALID | SCREENSHOT_INVALID` is the same as `SCREENSHOT_STALE`. I figured I'd be 'cleaner' and use a dedicated enum...

I can make the change to just use `INVALID` though, no problem.

src/feature/gui/gui-runner.c
Utagai: Yeah, [I figured it would get optimized](https://github.com/mgba-emu/mgba/pull/2929/commits/5e3e71df3e5ff03ed63d7bf6234451a9c7ee48c1) anyways! I'll just remove it! If I consider your idea in [this comment](https://github.com/mgba-emu/mgba/pull/2929#discussion_r1199835438), then I think we can just remove the need for the variable entirely.

endrift: I saw that, I was just agreeing haha. That type of optimization is specifically called common subexpression elimination, and is pretty basic.

Utagai: Did not know about its official name, TIL! Thanks for the link!

src/feature/gui/gui-runner.c
```diff
@@ -656,6 +656,10 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
 				runner->core->reset(runner->core);
 				break;
 			case RUNNER_SAVE_STATE:
+				struct mGUIBackground* gbaBackground = (struct mGUIBackground*) stateSaveMenu.background;
+				// If we are saving state, then the screenshot stored for the state previously should no longer be considered up-to-date.
+				// Therefore, mark it as stale, so that at draw time, we will re-draw to the new save state's screenshot.
+				gbaBackground->screenshotId |= SCREENSHOT_INVALID;
```
endrift, line 662: You didn't remove this variable declaration, which can just be merged into the statement anyway.

```suggestion
				// If we are saving state, then the screenshot stored for the state previously should no longer be considered up-to-date.
				// Therefore, mark it as stale, so that at draw time, we will re-draw to the new save state's screenshot.
				((struct mGUIBackground*) stateSaveMenu.background)->screenshotId |= SCREENSHOT_INVALID;
```

endrift (changes requested):
You missed my comment about _tryAutosave it seems.

Utagai:
Oops... my bad for the mistakes there! On it.

src/feature/gui/gui-runner.c
```diff
@@ -656,6 +656,9 @@ void mGUIRun(struct mGUIRunner* runner, const char* path) {
 				runner->core->reset(runner->core);
 				break;
 			case RUNNER_SAVE_STATE:
+				// If we are saving state, then the screenshot stored for the state previously should no longer be considered up-to-date.
+				// Therefore, mark it as stale, so that at draw time, we will re-draw to the new save state's screenshot.
```
endrift, line 660: This phrasing is a bit funky. Not super important, but I might change it a bit like so:
```suggestion
				// Therefore, mark it as stale so that at draw time we load the new save state's screenshot.
```

endrift (approved):
One last thing: since this is a user-reported bug with visible changes, please add an entry to the CHANGES file in the "Other fixes" section, using the mGUI prefix. Changes are sorted alphabetically by section then by when they were added, so if there are other mGUI changes it goes at the bottom of those, but otherwise it goes in alphabetically between the other entries.

Utagai:
Thanks! Hopefully I updated the `CHANGES` file correctly. :sweat_smile:

PR #2931, opened 2023-05-22
Mark video encoding options as untranslatable

### LINKED ISSUE #2927: Cannot select certain encoders when trying to record with non english locale
When using mGBA with a non English locale for recording, it's impossible to select certain encoders from the drop down menu (they become immediately red):

grafik

When the correct english term is manually typed in, it still works:

grafik

shenef:
Also used more descriptive names for when no video or audio codec is selected.
I don't have a build environment so these changes are untested.
Closes #2927

### DIFF
--- a/src/platform/qt/VideoView.cpp
+++ b/src/platform/qt/VideoView.cpp
@@ -271,7 +271,7 @@ void VideoView::setFilename(const QString& fname) {
 void VideoView::setAudioCodec(const QString& codec) {
 	free(m_audioCodecCstr);
 	m_audioCodec = sanitizeCodec(codec, s_acodecMap);
-	if (m_audioCodec == "none") {
+	if (m_audioCodec == "no audio") {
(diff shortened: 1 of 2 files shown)


endrift:
Please don't introduce an unrelated change in the PR, especially when the change itself is subjective. The latter case makes it harder to judge whether or not it should be merged at all, whereas if it's simply absent it's a lot easier to review and merge.

shenef:
Thx for the feedback, I wasn't sure if I should include it but went for it thinking I could just revert it if necessary.
But now I realize that that looks quite messy in the commit history.
If you want me to, I'll revert it or recreate the PR without it.

endrift:
You can force-push to the branch by specifying the target as `f371fd6d425a04fb992fc87f1a9220ce11ac802e:master` to roll it back one commit.

endrift:
Since this still has an unrelated change, I've just cherry picked the single commit.

PR #2956, opened 2023-06-19
Memory search and memory view rework

KurobaM:
Memory Search:
- limit displayed results & change ui layout
- change guess type -> ability to pick multiple number types
- support signed number, common encoding for text & byte array
- results caching & show previous results
- allow searching within specific memory region
- double click row to open memory view & limit number of MemoryView children to 1

Memory View:
- allow changing text encoding of stringVal
- display all int8, uint8, int16, uin16, int32, uint32 interpretation of value at cursor instead of switching between alignment

### DIFF
--- a/src/script/stdlib.c
+++ b/src/script/stdlib.c
@@ -199,7 +199,7 @@ void mScriptContextAttachStdlib(struct mScriptContext* context) {
 
 	mScriptContextSetDocstring(context, "system", "Information about the system the script is running under");
 	mScriptContextSetDocstring(context, "system.version", "The current version of this build of the program");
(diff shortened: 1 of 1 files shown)


endrift:
You can't really make sweeping changes like this without even asking first if any of them are welcome. Some may be, some may not be. Further, doing ALL of this in one commit makes it effectively impossible to merge due to the fact that if it breaks something, bisecting will lead to a monolithic change--narrowing down what in it change broke it is very difficult.

Please at the very least break it down into several commits with one change at a time. If we only end up wanting some of them, that will also make it much, much easier to select those ones.

PR #2975, opened 2023-07-07
Update the macOS icon to match the "Big Sur" icon style

aslilac:
Hey there!

I took a stab at designing an icon that looked a bit more at home on macOS. I didn't change the GBA model at all, just placed it over a shaped background with a little bit of a checkerboard texture. If you guys like it, I'd love to see it become the official one!

Also happy to make small tweaks to it, or provide more sizes of the PNG.


endrift:
I have no intention of changing the logo or icon due to the current whims of Apple, which will probably change again in the next 5 years anyway. Just because Apple's design language has squircles now doesn't mean third party icons need them.

aslilac:
Well, thanks for taking your frustration out on me then. 😅

PR #3024, opened 2023-09-17
Scripting: add signed/unsigned read/write functions

ahigerd:
After discussion with Icesythe7, I decided to just take the initiative to add signed versions of the memory functions.


endrift:
So the reason I haven't merged this--or really taken care of any of the other signed/unsigned discrepancies is bikeshedding. I'm not sure how I want everything to be handled, but I want it consistent across the API. I don't really like the idea of signed/unsigned versions of every read, but the alternative is unifying them as _probably_ unsigned and adding a utility function for converting between them.

PR #3042, opened 2023-10-08
Debugger: Fix GDB no-ack mode

leo60228:
[No-ack mode](https://sourceware.org/gdb/onlinedocs/gdb/Packet-Acknowledgment.html) is an optimization in the GDB protocol when running over a reliable transport mechanism like TCP. It's already implemented in the GDB stub, but the stub doesn't advertise support for it, and it's broken by an off-by-one in a comparison. These fixes make it work as intended.


endrift (approved):
Looks good, just need to wait on getting the builders back up before I merge (probably tomorrow)

PR #3069, opened 2023-11-17
GBA default RTC hardware

felixjones:
And also perform the [Everdrive save type detection ](https://krikzz.com/pub/support/everdrive-gba/OS/changelist.txt)

> If ROM isn't exists in bram-db then first char of ROM ID can be used as save memory selector:
1 - EEPROM
2 - SRAM
3 - FLASH-64
4 - FLASH-128
Number should be stored as char, eg 1 equal to 0x31 in hex format

---

# Rationale

For developing homebrew with ELF binaries on mGBA: there is no mechanism for overriding the RTC other than forcing the game code to something like a Pokemon game.

Not all toolchains provide the ability for setting a game code for an ELF file (devkitArm and gba-toolchain, for example), so users are out of luck by default.

Modern flashcarts and other emulators assume the RTC is present.

As for the Everdrive stuff, the community settled on that standard, so it was a "might as well while I'm here" kind of thing. Willing to drop that commit.


endrift:
I can't merge the RTC on by default for a handful of reasons, but the biggest one being that I would need to make sure the RTC is lazily initialized so it doesn't write the additional footer at the end of _every_ save file. This could also effect games that do buggy writes to that area, so it'd ideally only be for hacked games, but that's hard without all of no-intro being imported. Further, if we're gonna do this for RTC, we'd want to do it for rumble too.

I'm thinking ideally it might be an option to add to the settings screen for "include cartridge hardware in all games" or something similar.

I'd be willing to merge the other commit as-is though.

src/gba/overrides.c
```diff
@@ -424,6 +424,23 @@ void GBAOverrideApplyDefaults(struct GBA* gba, const struct Configuration* overr
...
+				break;
+			case '2':
+				override.savetype = SAVEDATA_SRAM;
+				break;
+			case '3':
+				override.savetype = SAVEDATA_FLASH512;
+				break;
+			case '4':
+				override.savetype = SAVEDATA_FLASH1M;
+				break;
```
endrift, line 441: I'd recommend a `default: return;` here so it doesn't call GBAOverrideApply if it's not one of those four, but otherwise looks good.

endrift (changes requested):

src/gba/overrides.c
endrift: Ah actually it's not compiling in the CI. I might have changed the constants to start with GBA_* in the meantime.

PR #3083, opened 2023-11-28
Add Goodboy Galaxy to overrides

DanTheMan827:
To enable rumble


endrift:
Traditionally I don't accept overrides for homebrew, but this is actually a commercial title so I'll make an exception.

PR #3090, opened 2023-12-19
Qt: Fix generic icon on Wayland

FireSpike0:
The method used to set the window icon (`setWindowIcon`) does not work anymore on Wayland which results in a generic Wayland icon. To set the icon on Wayland, programs should make use of the `setDesktopFileName` method which hints the graphical shell to use the icon specified in the referenced desktop file. This icon is used for the taskbar and the window border / decorations. The described bug occurs especially when using Plasma with Wayland, but other shells using Wayland may be affected too.

The added method was introduced in Qt 5.7 which eventually raises the minimum required version of Qt.

Testing the new commit on Unix resulted in a working icon when using Plasma regardless of Wayland or X11.

If I've missed anything or should fix something, just ask and I'll try.

### DIFF
--- a/src/platform/qt/main.cpp
+++ b/src/platform/qt/main.cpp
@@ -104,6 +104,10 @@ int main(int argc, char* argv[]) {
 	QApplication::setWindowIcon(QIcon(":/res/mgba-256.png"));
 #endif
 
+#ifdef Q_OS_UNIX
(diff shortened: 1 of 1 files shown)


endrift:
Thanks, I was wondering what was going on.

PR #3097, opened 2023-12-30
Res: Add gbc-lcd shader, imitates GBC LCD subpixel arrangement with an optional backlight effect

pineapplemachine:
I wrote a shader that attempts to reproduce the GBC LCD subpixel pattern. It also supports a backlight bleed effect similar to the existing ags001 shader. I used ags001 as a starting point, but otherwise this is new work and not adapted from any other shader. I wrote it primarily for my own use, but I'm guessing it may be of interest to others too.

The license included with the shader is the Unlicense, i.e. the shader is intended to be public domain.

Some credit belongs to this resource: https://gbcc.dev/technology/

Screenshots:

Settings (it doesn't seem like mGBA currently has any easy way to put the options in an intended order?):

image
image

1x (not really intended to be used this way, though I'm sure that if you were really determined you could mess with the default settings enough to get something interesting):
image
image

2x:
image
image

2x (with settings tweaked to better suit the small size, "Subpixel glow size" set to 0.255 instead of default ):
image
image

3x:
image
image

4x:
image
image

8x:
image
image
image
image
image

Zoomed in on an 8x screen, to show the subpixels:
image

Screenshot of subpixels scaled up with the glow/luminance effect toned down, to show size and shape (achieved by changing the "Subpixel scale" and "Subpixel glow size" settings)
image

### DIFF
--- a/res/shaders/gbc-lcd.shader/gbc-lcd-light.fs
+++ b/res/shaders/gbc-lcd.shader/gbc-lcd-light.fs
@@ -56,8 +56,10 @@ uniform vec2 ReflectionDistance;
 
 #define M_PI 3.1415926535897932384626433832795
 
-// Helper to compute backlight bleed intensity
-// for a texCoord input.
+/**
+ * Helper to compute backlight bleed intensity
+ * for a texCoord input.
+ */
 float getLightIntensity(vec2 coord) {
     vec2 coordCentered = coord - vec2(0.5, 0.5);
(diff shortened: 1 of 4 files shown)


pineapplemachine:
I may have gotten a bit zealous and did some significant changes to improve the subpixel accuracy 😅

Please don't merge this just yet, I will update the PR in the near future with the improved shader and new screenshots

PR #3100, opened 2024-01-01
Fix #3078  issue

### LINKED ISSUE #3078: Selected shader not applying
Note: the second issue below turned out to be user error.

Hello! I've found 2-ish issues with regards to settings not being set properly when mGBA starts. They may be related, maybe not.

Some system info: I'm running EndeavourOS (Arch-based Linux distro), I have an AMD Ryzen 5 3600 and an AMD Radeon RX 5700 XT.

First: when I start mGBA, the shader is not loaded, even though it is present in the config files. I can apply the shader and save the settings to the config files, but the shader will not be loaded next time I open mGBA. This occurs on the development build I am using (0.11-8252-d75d26bcc via the AUR) but not on the release build I am using (0.10.2 via the Arch Linux extra repository (the package is versioned as 0.10.2-2)).

~~Second: when I launch a GB-only game, I'm pretty sure the game uses the default GBC color palette until I hit "Apply" on the Settings page, even though I have the GB-only model set to DMG and I have the GB palette set to DMG Green, with "GBC color palette if available" checked. This appears to be the same issue as #2377, and this happens on both builds I mentioned in the first bug, not just the development build.~~

I have attached a video of what this looks like, and a bug report from each of the builds I mentioned. I recorded the bugs with the development build, but the process I did before generating the bug report for the release build was identical.

[mGBA Shader + Palette Bug Report.zip](https://github.com/mgba-emu/mgba/files/13461653/mGBA.Shader.%2B.Palette.Bug.Report.zip)
[mGBA Shader + Palette Bug Report (release build).zip](https://github.com/mgba-emu/mgba/files/13461670/mGBA.Shader.%2B.Palette.Bug.Report.release.build.zip)

https://github.com/mgba-emu/mgba/assets/9244116/f574e1f7-ebba-4982-b537-f60eb15202b2

luc-git:
Fix the issue #3078

### DIFF
--- a/src/platform/qt/Display.cpp
+++ b/src/platform/qt/Display.cpp
@@ -116,18 +116,16 @@ void QGBA::Display::configure(ConfigController* config) {
 	config->updateOption("videoSync");
 #if defined(BUILD_GL) || defined(BUILD_GLES2) || defined(BUILD_GLES3)
 	if (opts->shader && supportsShaders()) {
-		auto loadshader = [this](struct VDir* shader) {
-			setShaders(shader);
-			shader->close(shader);
-		};
 		struct VDir* shader = VDirOpenArchive(opts->shader);
 		if (shader) {
-			loadshader(shader);
+			setShaders(shader);
+			shader->close(shader);
 			return;
 		}
 		shader = VDirOpen(opts->shader);
 		if (shader) {
-			loadshader(shader);
+			setShaders(shader);
+			shader->close(shader);
 		}
 	}
 #endif
--- a/src/platform/qt/Display.cpp
+++ b/src/platform/qt/Display.cpp
@@ -117,16 +117,14 @@ void QGBA::Display::configure(ConfigController* config) {
 #if defined(BUILD_GL) || defined(BUILD_GLES2) || defined(BUILD_GLES3)
 	if (opts->shader && supportsShaders()) {
 		struct VDir* shader = VDirOpenArchive(opts->shader);
-		if (shader) {
-			setShaders(shader);
-			shader->close(shader);
-			return;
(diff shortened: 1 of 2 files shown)


src/platform/qt/Display.cpp
```diff
@@ -116,10 +116,18 @@ void QGBA::Display::configure(ConfigController* config) {
 	config->updateOption("videoSync");
 #if defined(BUILD_GL) || defined(BUILD_GLES2) || defined(BUILD_GLES3)
 	if (opts->shader && supportsShaders()) {
-		struct VDir* shader = VDirOpen(opts->shader);
-		if (shader) {
+		auto loadshader = [this](struct VDir* shader) {
```
endrift, line 119: It's just two lines. Please don't use a closure for this.

endrift (changes requested):

src/platform/qt/Display.cpp
```diff
@@ -116,11 +116,15 @@ void QGBA::Display::configure(ConfigController* config) {
...
-			shader->close(shader);
+		struct VDir* shader = VDirOpenArchive(opts->shader);
+		if (!shader) {
+			shader = VDirOpen(opts->shader);
+			if (!shader) {
+				return;
+			}
 		}
+		setShaders(shader);
+		shader->close(shader);
```
endrift, line 127:
```suggestion
		struct VDir* shader = VDirOpenArchive(opts->shader);
		if (!shader) {
			shader = VDirOpen(opts->shader);
		}
		if (shader) {
			setShaders(shader);
			shader->close(shader);
		}
```

endrift (approved):

PR #3113, opened 2024-01-20
GDB Debugger default to loopback address

Martosis:
The current version of mGBA binds the GDB server to 0.0.0.0 which makes it publicly accessible. This allows an outside attacker to connect to mGBA's GDB server and execute arbitrary code. Although the Qt window allows for changing the bind address, there is no way (as far as I am aware) to automatically change the bind address when running mGBA from the command line with the argument --gdb. These commits change the default bind address from 0.0.0.0 to the loopback address 127.0.0.1.


endrift:
While this change isn't necessarily a bad idea, I think you're mistaken about how dangerous it is. The executed code is run in an emulated context. They'd have to find a way to break out of that for it to be actually arbitrary code.

I'll need to review this in a bit to make sure it's implemented right.

PR #3117, opened 2024-01-21
core: Add support for specifying an arbitrary portable directory.

Steveice10:
Adds support for specifying an arbitrary portable directory path inside the `portable.ini` file. The path can be either absolute or relative to the executable directory. This allows for the user to consolidate all of the mGBA config and other user data into a single directory wherever they want.

For example, the `portable.ini` could look like this to use a directory named "Data" next to the executable:
```ini
[Portable]
Path=Data
```

Or it could look like this to use a fixed directory on the user's desktop:
```ini
[Portable]
Path=/home/user/Desktop/mgba
```

The PR also includes updates to the Qt UI to support specifying a path when you choose to make the installation portable:
Screenshot 2024-01-21 at 4 37 07 PM

### DIFF
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -198,38 +198,46 @@ bool mCoreConfigSaveVFile(const struct mCoreConfig* config, struct VFile* vf) {
 	return ConfigurationWriteVFile(&config->configTable, vf);
 }
 
-void mCoreConfigMakePortable(const struct mCoreConfig* config) {
+void mCoreConfigMakePortable(const struct mCoreConfig* config, const char* path) {
 	struct VFile* portable = NULL;
+	struct Configuration portableConfig;
 	char out[PATH_MAX];
-	mCoreConfigPortablePath(out, sizeof(out));
+	mCoreConfigPortableIniPath(out, sizeof(out));
 	if (!out[0]) {
 		// Cannot be made portable
 		return;
 	}
-	portable = VFileOpen(out, O_WRONLY | O_CREAT);
+
+	ConfigurationInit(&portableConfig);
+
+	portable = VFileOpen(out, O_RDONLY);
 	if (portable) {
+		ConfigurationReadVFile(&portableConfig, portable);
 		portable->close(portable);
-		mCoreConfigSave(config);
 	}
+
+	if (path && path[0]) {
+		ConfigurationSetValue(&portableConfig, "portable", "path", path);
+	} else {
+		ConfigurationClearValue(&portableConfig, "portable", "path");
+	}
+
+	portable = VFileOpen(out, O_WRONLY | O_CREAT | O_TRUNC);
+	if (portable) {
+		ConfigurationWriteVFile(&portableConfig, portable);
+		portable->close(portable);
+	}
+
+	ConfigurationDeinit(&portableConfig);
+	mCoreConfigSave(config);
 }
 
 void mCoreConfigDirectory(char* out, size_t outLength) {
-	struct VFile* portable;
 	char portableDir[PATH_MAX];
 	mCoreConfigPortablePath(portableDir, sizeof(portableDir));
 	if (portableDir[0]) {
-		portable = VFileOpen(portableDir, O_RDONLY);
-		if (portable) {
-			portable->close(portable);
-			if (outLength < PATH_MAX) {
-				char outTmp[PATH_MAX];
-				separatePath(portableDir, outTmp, NULL, NULL);
-				strlcpy(out, outTmp, outLength);
-			} else {
-				separatePath(portableDir, out, NULL, NULL);
-			}
-			return;
-		}
+		strlcpy(out, portableDir, outLength);
+		return;
 	}
 #ifdef _WIN32
 	WCHAR wpath[MAX_PATH];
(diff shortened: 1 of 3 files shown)


endrift:
I'll need to take a closer look at this before I can review it properly, but I'd thought about implementing something like this that just pulled the config path out of a key that's actually *in* the portable.ini as opposed to just having the file's existence be a marker.

Steveice10:
> I'd thought about implementing something like this that just pulled the config path out of a key that's actually _in_ the portable.ini as opposed to just having the file's existence be a marker.

That's a much better idea, updated the PR to do just that. Let me know what you think of the implementation as I'm still fairly new to this codebase.

src/core/config.c
```diff
@@ -198,38 +198,34 @@ bool mCoreConfigSaveVFile(const struct mCoreConfig* config, struct VFile* vf) {
...
 		// Cannot be made portable
 		return;
 	}
 	portable = VFileOpen(out, O_WRONLY | O_CREAT);
 	if (portable) {
+		if (path && path[0]) {
+			struct Configuration portableConfig;
+			ConfigurationInit(&portableConfig);
+			ConfigurationSetValue(&portableConfig, "Portable", "Path", path);
+			ConfigurationWriteVFile(&portableConfig, portable);
```
endrift, line 215: You leak the Configuration here. You need to call ConfigurationDeinit

src/core/config.c
```diff
@@ -308,18 +304,48 @@ void mCoreConfigPortablePath(char* out, size_t outLength) {
...
+			if (outLength < PATH_MAX) {
+				char outTmp[PATH_MAX];
+				separatePath(portableIniPath, outTmp, NULL, NULL);
+				strlcpy(out, outTmp, outLength);
+			} else {
+				separatePath(portableIniPath, out, NULL, NULL);
+			}
+
+			struct Configuration portableConfig;
+			ConfigurationInit(&portableConfig);
```
endrift, line 325: You also leak this Configuration.

src/core/config.c
```diff
@@ -308,18 +304,48 @@ void mCoreConfigPortablePath(char* out, size_t outLength) {
...
+			if (ConfigurationReadVFile(&portableConfig, portableIni)) {
+				const char* path = ConfigurationGetValue(&portableConfig, "Portable", "Path");
```
endrift, line 327:
```suggestion
				const char* path = ConfigurationGetValue(&portableConfig, "portable", "path");
```

src/core/config.c
```diff
@@ -198,38 +198,34 @@ bool mCoreConfigSaveVFile(const struct mCoreConfig* config, struct VFile* vf) {
...
 		return;
 	}
 	portable = VFileOpen(out, O_WRONLY | O_CREAT);
 	if (portable) {
+		if (path && path[0]) {
+			struct Configuration portableConfig;
+			ConfigurationInit(&portableConfig);
+			ConfigurationSetValue(&portableConfig, "Portable", "Path", path);
+			ConfigurationWriteVFile(&portableConfig, portable);
+		}
```
endrift, line 216: You should have an else clause that resets the value to an empty string, otherwise it won't overwrite a previous setting, if applicable.

src/platform/qt/ConfigController.cpp
```diff
@@ -365,8 +365,8 @@ void ConfigController::write() {
 	mCoreConfigMap(&m_config, &m_opts);
 }
 
-void ConfigController::makePortable() {
-	mCoreConfigMakePortable(&m_config);
+void ConfigController::makePortable(const char* path) {
+	mCoreConfigMakePortable(&m_config, path);
```
endrift, line 369: For this PR can you hold off on the Qt changes and just pass NULL? We can do a followup PR with Qt changes so the core changes can go in first. It might take a few rounds of bikeshedding the Qt changes for a UI I like.

endrift (changes requested):
Minus some minor fixes, the core changes look fine. This is pretty much the change I had toyed with, but never got around to implementing it since I wasn't sure if it'd be _useful_.

src/platform/qt/ConfigController.cpp
Steveice10: Removed, will submit as a separate PR after.

Steveice10:
Addressed the review comments.

src/core/config.c
```diff
@@ -308,18 +308,49 @@ void mCoreConfigPortablePath(char* out, size_t outLength) {
...
+	struct VFile* portableIni;
+	char portableIniPath[PATH_MAX];
+	mCoreConfigPortableIniPath(portableIniPath, sizeof(portableIniPath));
+	out[0] = '\0';
+	if (portableIniPath[0]) {
+		portableIni = VFileOpen(portableIniPath, O_RDONLY);
+		if (portableIni) {
+			// Start with the path that the portable.ini file exists in.
+			if (outLength < PATH_MAX) {
+				char outTmp[PATH_MAX];
```
endrift, line 321: It'd be cleaner (and probably less risky of accidental off-by-one for the null terminator)  if you hosted this and used snprintf in the part you use strncat.

endrift (changes requested):
Some minor nits, but otherwise good.

src/core/config.c
```diff
@@ -198,38 +198,38 @@ bool mCoreConfigSaveVFile(const struct mCoreConfig* config, struct VFile* vf) {
...
+		if (path && path[0]) {
+			struct Configuration portableConfig;
+			ConfigurationInit(&portableConfig);
+			ConfigurationSetValue(&portableConfig, "portable", "path", path);
+			ConfigurationWriteVFile(&portableConfig, portable);
+			ConfigurationDeinit(&portableConfig);
+		} else {
+			// For an empty path, reset the file to also be empty.
+			portable->truncate(portable, 0);
+		}
```
endrift, line 220: For future proofing, I'd prefer if you loaded the old file and just update or delete that key.

endrift (changes requested):

PR #3124, opened 2024-01-28
Add a `--script` option to load LUA scripts automatically on startup

Eijebong:


src/platform/qt/ConfigController.cpp
```diff
@@ -153,7 +156,12 @@ ConfigController::ConfigController(QObject* parent)
 	m_subparsers[1].usage = "Frontend options:\n"
 	    "  --ecard FILE  Scan an e-Reader card in the first loaded game\n"
 	    "                Can be passed multiple times for multiple cards\n"
-	    "  --mb FILE     Boot a multiboot image with FILE inserted into the ROM slot";
+	    "  --mb FILE     Boot a multiboot image with FILE inserted into the ROM slot"
+#ifdef ENABLE_SCRIPTING
+	    "  --script FILE Lua script to load on start"
```
endrift, line 161:
```suggestion
	    "\n  --script FILE Script file to load on start"
```

This needs a newline at the beginning so it doesn't get concatenated to the previous line. Additionally, scripting languages beyond Lua are planned, so let's not be specific here.

src/platform/qt/ConfigController.cpp
```diff
@@ -153,7 +156,12 @@ ConfigController::ConfigController(QObject* parent)
...
+#endif
+	    "";
```
endrift, line 163:
```suggestion
	    ;
```

src/platform/qt/ConfigController.cpp
```diff
@@ -171,6 +179,15 @@ ConfigController::ConfigController(QObject* parent)
 			self->m_argvOptions[optionName] = QString::fromUtf8(arg);
 			return true;
 		}
+#ifdef ENABLE_SCRIPTING
+		if (optionName == QLatin1String("script")) {
+			QString scriptName = QString::fromUtf8(arg);
+			QList<QString> currentList = self->m_argvOptions[optionName].toStringList();
+			currentList.push_back(scriptName);
+			self->m_argvOptions[optionName] = QVariant(currentList);
```
endrift, line 187: Please follow the same convention as the ecard block so the option can be specified multiple times.

src/platform/qt/Window.cpp
```diff
@@ -2184,6 +2184,18 @@ void Window::setController(CoreController* controller, const QString& fname) {
 		m_controller->setPaused(true);
 		m_pendingPause = false;
 	}
+
+#ifdef ENABLE_SCRIPTING
+	if (!m_scripting) {
+		QList<QString> scripts = m_config->getArgvOption("script").toStringList();
```
endrift, line 2190:
```suggestion
		QStringList scripts = m_config->getArgvOption("script").toStringList();
```

endrift (changes requested):
Mostly good, though I've been thinking about adding this to the base command line argument list so it can be used in the SDL and some other versions.

PR #3128, opened 2024-01-30
qt: Add text input for specifying the portable directory.

Steveice10:
Follow up to https://github.com/mgba-emu/mgba/pull/3117 with the GUI component, for separate review. This adds a more user-friendly way to decide where the portable directory should be. Input can be either absolute or relative; for example the user could just put `data` and the directory would be named `data` next to the `portable.ini`.

Screenshot:
Screenshot 2024-01-21 at 4 37 07 PM


JakobDev:
Why not just use a File Picker?

Steveice10:
> Why not just use a File Picker?

I don’t think a file picker would support the relative path case very well.

PR #3144, opened 2024-02-16
Add a button to add dir to the library

luc-git:
It's not very clear on how to add a directory to the library. So this PR adds a button which is only visible while library is enabled and when there is no directory in the library.

Problem: button is not centered at all


ahigerd:
You're going to need to rework this to be compatible with https://github.com/mgba-emu/mgba/pull/2570

ahigerd:
Also, do fix the positioning. :)

PR #3151, opened 2024-02-27
fix win32 build

oltolm:
This should fix the win32 build. I don't have the problem because I use MSYS2.


endrift:
Finally got around to testing this. It is insufficient in fixing the compilation errors, as `SHGetKnownFolderPath` appears to still be missing. Further, threading.h is still broken due to the use of ConditionVariable and InitOnceExecuteOnce

endrift:
I've pushed my own version of the fixes, which are more widely applicable per its library use case. Let me know if it introduces any problems on MinGW and I can look directly.

PR #3172, opened 2024-03-25
Fix typo in spanish translation file

VMRuiz:
I have fixed a typo in the word "esportar," which should be "exportar." I couldn't find any other similar typos. However, there are some translations that don't feel quite right. For instance, "Frame" is typically not translated in Spanish, so seeing it translated as "Cuadro," although technically correct, feels unusual.

Please let me know if you would like me to address this in a new PR as well.


endrift:
I appreciate checking over the translation, but due to potential merge conflicts I only accept translation updates via [Weblate](https://hosted.weblate.org/projects/mgba/mgba-qt/). I've been worried about stilted translations, so fixing those is also appreciated.

PR #3174, opened 2024-03-28
fix hang in AudioDevice

oltolm:
So I tried the latest version e3edca1f4168c1eee36cd213ef8d5e91fd35e400 and games do not boot they just hang. This is my attempt to fix it. I don't know if it is correct, but it actually works. Games boot and are playable.

### DIFF
--- a/src/platform/qt/AudioDevice.cpp
+++ b/src/platform/qt/AudioDevice.cpp
@@ -64,11 +64,15 @@ qint64 AudioDevice::writeData(const char*, qint64) {
 }
 
 bool AudioDevice::atEnd() const {
+	return !bytesAvailable();
+}
+
+qint64 AudioDevice::bytesAvailable() const {
 	if (!m_context->core) {
 		return true;
 	}
 	mCoreSyncLockAudio(&m_context->impl->sync);
-	bool available = blip_samples_avail(m_context->core->getAudioChannel(m_context->core, 0)) == 0;
+	int available = blip_samples_avail(m_context->core->getAudioChannel(m_context->core, 0));
 	mCoreSyncUnlockAudio(&m_context->impl->sync);
-	return available;
+	return available * sizeof(mStereoSample);
 }
--- a/src/platform/qt/AudioDevice.h
+++ b/src/platform/qt/AudioDevice.h
@@ -21,6 +21,7 @@ public:
 	void setInput(mCoreThread* input);
 	void setFormat(const QAudioFormat& format);
 	bool atEnd() const override;
+	qint64 bytesAvailable() const override;
 
 protected:
 	virtual qint64 readData(char* data, qint64 maxSize) override;


src/platform/qt/AudioDevice.cpp
```diff
@@ -63,12 +63,17 @@ qint64 AudioDevice::writeData(const char*, qint64) {
...
+	return bytesAvailable() != 0;
+}
+
+qint64 AudioDevice::bytesAvailable() const {
 	mCoreSyncLockAudio(&m_context->impl->sync);
-	bool available = blip_samples_avail(m_context->core->getAudioChannel(m_context->core, 0)) == 0;
+	int n = blip_samples_avail(m_context->core->getAudioChannel(m_context->core, 0));
 	mCoreSyncUnlockAudio(&m_context->impl->sync);
-	return available;
+	return n * sizeof(mStereoSample);
```
endrift, line 78: Nit: leave the variable named `available` even if you change the type--this is the name I use in other places for the same value.

src/platform/qt/AudioDevice.h
```diff
@@ -20,7 +20,9 @@ Q_OBJECT
 
 	void setInput(mCoreThread* input);
 	void setFormat(const QAudioFormat& format);
-	bool atEnd() const override;
+	virtual bool isSequential() const override;
+	virtual bool atEnd() const override;
+	virtual qint64 bytesAvailable() const override;
```
endrift, line 25: Nit: `virtual` is redundant when `override` is applied, so I've not been putting it in new code.

```suggestion
	bool isSequential() const override;
	bool atEnd() const override;
	qint64 bytesAvailable() const override;
```

endrift (commented):
If this works locally I'll merge it, but I was having trouble with it stalling out and never resuming, e.g. if I pause or reset a game. The solution I committed is definitely hacky, but it seemed to work on my end.

endrift:
Unfortunately it does not seem to work on my machine and hangs instead.

src/platform/qt/AudioDevice.cpp
```diff
@@ -63,12 +63,17 @@ qint64 AudioDevice::writeData(const char*, qint64) {
...
 
+bool AudioDevice::isSequential() const {
+	return true;
+}
+
 bool AudioDevice::atEnd() const {
-	if (!m_context->core) {
-		return true;
-	}
+	return bytesAvailable() != 0;
```
endrift, line 71: This check appears to be backwards.

```suggestion
	return !bytesAvailable();
```

endrift (commented):
If I make this one change and *don't* revert the hack, it works on my end. Does that still work on your end?

endrift:
I think I forgot to mention that this doesn't work if I revert my hack. It looks like I only said that your version doesn't break the hack if the hack is left in place.

Which audio backend are you using for Qt Multimedia? I think your change may only fix it in some. In the build I have for Linux it uses PulseAudio, which seems like it will just stall out regardless and doesn't care about if it's sequential or not, hence why I added the hack initially. Your fix may fix other backends I didn't test with.

oltolm:
I use the Qt6 FFMPEG backend on Windows. On Linux there are two backends: FFMPEG and gstreamer. You are probably using gstreamer.

I re-reverted the hack and with my change it works for me.

oltolm:
This commit https://github.com/mgba-emu/mgba/commit/f51cb153d109f842fdc5325715be97a71fcacd28 broke Qt Multimedia audio backend again. It either hangs or the audio sounds broken.

PR #3201, opened 2024-05-01
fix shortcut handling

oltolm:
Setting shortcuts does not work properly for me. If I want to assign `Ctrl+F` to some action I press `Ctrl` and it shows "Control" then I press `F` and it shows `F` and finally it assigns `F`. But that's not what I want, I want `Ctrl+F` to be shown and assigned. This PR fixes that.

### DIFF
--- a/src/platform/qt/KeyEditor.cpp
+++ b/src/platform/qt/KeyEditor.cpp
@@ -41,7 +41,7 @@ void KeyEditor::setValue(int key) {
 	if (m_button) {
 		updateButtonText();
 	} else {
-		if (key < 0) {
+		if (key == Qt::Key_unknown) {
 			setText(tr("---"));
 		} else {
 			setText(keyName(key));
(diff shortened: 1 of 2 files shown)


endrift:
So while this does fix a bug, it seems to introduce a new one: when I press Ctrl on its own it says I've pressed Ctrl-Control, which is...well, confusing at the very least. The point of the old version was that if the only thing that was pressed was a modifier it didn't show up that way, but clearly the code rotted and broke at some point.

oltolm:
Yes, I noticed that too, but IMO it does not matter because you can not assign `Ctrl` alone and the code actually works. But I can try to fix it.

endrift:
At the very least it's confusing. But I do know players who want to be able to bind Ctrl or Shift for the keybindings, which isn't currently possible. Unfortunately, I don't Qt can manage left vs right but having it bindable at all even if it's not differentiable is still preferable to not being able to be bound at all.

oltolm:
I changed the logic so that when you press control alone you get `Ctrl+` instead of `Ctrl+control`. Now it's not confusing anymore. I don't think it's possible to make Qt accept shortcuts like `control` alone.

waddlesplash:
Why not just use the standard class `QKeySequenceEdit` here? I guess this class was probably added when Qt 4 was still being supported (QKeySequenceEdit was introduced in 5.2), so, why not replace it?

endrift:
mGBA never supported Qt 4, but I think it did support pre-5.2 at some point. I'd have to look into it, but I think I had a reason.

PR #3205, opened 2024-05-11
update readme in zh_cn

evrins:
previous translation in zh_cn is good but is kind of outdated

sync with current readme and fix minor erorrs


endrift:
Thanks!

PR #3224, opened 2024-06-03
Qt: Ignore rewind while loading/saving states

yeah-its-gloria:
I dunno if I did it right but I've managed to rewind while reloading states by accident (well, at first, it's sorta a guilty pleasure now ) so much that I figured this might be worth Fixing™️


endrift:
oh yeah this uhh this seems like a good idea

PR #3231, opened 2024-06-09
GBA Memory: Support the Chinese version of Digimon Sapphire by Vast Fame

RibShark:
The Chinese version of Digimon Sapphire by Vast Fame uses a third set of address reordering modes. This adds those modes and uses the CRC of the ROM to determine which to use. The header of the Chinese version is identical to the English version, so I couldn't use the header name to determine the mode unlike with George Sango.

### DIFF
--- a/include/mgba/internal/gba/cart/vfame.h
+++ b/include/mgba/internal/gba/cart/vfame.h
@@ -13,11 +13,13 @@
 
 CXX_GUARD_START
 
+#define DIGIMON_SAPPHIRE_CHINESE_CRC32 0x793A328F
+
 enum GBAVFameCartType {
 	VFAME_NO = 0,
 	VFAME_STANDARD = 1,
-	VFAME_ALTERNATE = 2,
-	VFAME_GEORGE = 3
+	VFAME_GEORGE = 2,
+	VFAME_ALTERNATE = 3,
 };
 
 struct GBAVFameCart {
@@ -29,7 +31,7 @@ struct GBAVFameCart {
 };
 
 void GBAVFameInit(struct GBAVFameCart* cart);
-void GBAVFameDetect(struct GBAVFameCart* cart, uint32_t* rom, size_t romSize);
+void GBAVFameDetect(struct GBAVFameCart* cart, uint32_t* rom, size_t romSize, uint32_t crc32);
 void GBAVFameSramWrite(struct GBAVFameCart* cart, uint32_t address, uint8_t value, uint8_t* sramData);
 uint32_t GBAVFameModifyRomAddress(struct GBAVFameCart* cart, uint32_t address, size_t romSize);
 uint32_t GBAVFameGetPatternValue(uint32_t address, int bits);
--- a/include/mgba/internal/gba/cart/vfame.h
+++ b/include/mgba/internal/gba/cart/vfame.h
@@ -16,7 +16,8 @@ CXX_GUARD_START
 enum GBAVFameCartType {
 	VFAME_NO = 0,
 	VFAME_STANDARD = 1,
-	VFAME_GEORGE = 2
+	VFAME_ALTERNATE = 2,
+	VFAME_GEORGE = 3
 };
 
 struct GBAVFameCart {
--- a/src/gba/cart/vfame.c
+++ b/src/gba/cart/vfame.c
@@ -7,11 +7,17 @@
 
 #include <mgba/internal/gba/gba.h>
 #include <mgba/internal/gba/memory.h>
+#include <mgba-util/crc32.h>
 
 static const uint8_t ADDRESS_REORDERING[4][16] = {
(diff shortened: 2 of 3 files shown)


include/mgba/internal/gba/cart/vfame.h
```diff
@@ -16,7 +16,8 @@ CXX_GUARD_START
 enum GBAVFameCartType {
 	VFAME_NO = 0,
 	VFAME_STANDARD = 1,
-	VFAME_GEORGE = 2
+	VFAME_ALTERNATE = 2,
+	VFAME_GEORGE = 3
```
endrift, line 20: Should probably keep these in the original order and add ALTERNATE as 3. It doesn't really matter TOO much, but is there a strong reason not to?

Also ending the terminal entry with a comma helps reduce code churn when new entries are added.

src/gba/cart/vfame.c
```diff
@@ -65,13 +76,18 @@ void GBAVFameDetect(struct GBAVFameCart* cart, uint32_t* rom, size_t romSize) {
...
+	// These games additionally operates with a different set of SRAM modes
+	// Their initialisation seems to be identical so the difference must be in the cart HW itself
 	// Other undumped games may have similar differences
 	if (memcmp("George Sango", &((struct GBACartridge*) rom)->title, 12) == 0) {
 		cart->cartType = VFAME_GEORGE;
 		mLOG(GBA_MEM, INFO, "George mode");
 	}
+	// Chinese version of Digimon Sapphire; header is identical to the English version which uses the normal reordering
+	// so we have to use some other way to detect it
+	else if (doCrc32(rom, romSize) == 0x793A328F) {
```
endrift, line 88: This value is already cached in gba->romCrc32. Plus, you should probably move the literal to a #define or enum.

Also, mGBA style has the else if on the same line as the closing bracket. Put the comment inside the block when necessary.

src/gba/cart/vfame.c
```diff
@@ -65,13 +76,18 @@ void GBAVFameDetect(struct GBAVFameCart* cart, uint32_t* rom, size_t romSize) {
...
+		cart->cartType = VFAME_STANDARD;
```
endrift, line 89: Did you mean to assign ALTERNATE here? Otherwise, it's never assigned.

src/gba/cart/vfame.c
```diff
@@ -256,6 +272,8 @@ static uint32_t _modifySramAddress(enum GBAVFameCartType type, uint32_t address,
 		return address;
 	} else if (type == VFAME_GEORGE) {
 		return _reorderBits(address, ADDRESS_REORDERING_GEORGE[mode - 1], 16);
+	} else if (type == VFAME_ALTERNATE) {
+		return _reorderBits(address, ADDRESS_REORDERING_ALTERNATE[mode - 1], 16);
 	} else {
 		return _reorderBits(address, ADDRESS_REORDERING[mode - 1], 16);
 	}
```
endrift, line 279: If it's gonna have more than two values with the same semantics, it's probably a good idea to change this block into a LUT of the orders, indexed by the type

src/gba/cart/vfame.c
```diff
@@ -266,6 +284,8 @@ static int8_t _modifySramValue(enum GBAVFameCartType type, uint8_t value, int mo
 	if (reorderType != 0) {
 		if (type == VFAME_GEORGE) {
 			value = _reorderBits(value, VALUE_REORDERING_GEORGE[reorderType - 1], 8);
+		} else if (type == VFAME_ALTERNATE) {
+			value = _reorderBits(value, VALUE_REORDERING_ALTERNATE[reorderType - 1], 8);
 		} else {
 			value = _reorderBits(value, VALUE_REORDERING[reorderType - 1], 8);
 		}
```
endrift, line 291: Same comment here about a LUT

endrift (changes requested):
cc @tzlion

src/gba/cart/vfame.c
RibShark: oops, my bad there, yeah I did

RibShark:
On further inspection of the different modes, I have come to the conclusion that they are not random and must be generated in some way.

For the value reordering, for any given mode the reordered bit must not be in its original position, and any odd numbered bits must remain odd and vice versa. Furthermore it seems that applying the mode 2 transformation to the mode 1 set always produces the mode 3 set.

The address reordering is somewhat similar. Ignoring bit 15, every 5 bit positions must have some permutation of the following sets that they are reordered to: (0, 7, 14), (1, 5, 12), (2, 8, 11), (3, 9, 13), (4, 6, 10). This is repeated two more times for the middle five bit positions and the lower five bit positions.

These patterns make me feel that there has to be some mathematical way to generate these that would be cleaner than just using a LUT, but I can't work out exactly how.

tzlion:
Yeah honestly I also thought there _should_ be a mathematical way to do it when I originally implemented this but I could never figure it out either

endrift:
Alright, we can worry about that later then.

PR #3240, opened 2024-06-22
Visual studio debug enhancements

chrisepstein:
MSVC rename of library, to avoid conflicting pdb files, to allow debug of library from visual studio.
Git ignore MSVS output/build directories
Added VCPKG to manifest to allow manifest implementation of packages using VCPKG


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.6)
```
endrift, line 1: This is not related to this PR and should be reverted

CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ if(NOT MSVC)
 else()
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_CRT_SECURE_NO_WARNINGS /wd4003 /wd4244 /wd4146 /wd4267 /Zc:preprocessor-")
 	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_CRT_SECURE_NO_WARNINGS /wd4003 /wd4244 /wd4146 /wd4267 /Zc:preprocessor-")
+	#MSVC debug does not like both .exe and shared .dll to be named the same, as it prevents loading of debug symbols
+ set(BINARY_NAME mgbalib)
```
endrift, line 47: Your indentation does not match the rest of the file. This is also not the only thing BINARY_NAME is  be used for, so you'll need to narrow it down with the actual library file.

CMakeLists.txt
```diff
@@ -210,7 +212,7 @@ endif()
 set(OS_DEFINES)
 if(WIN32)
 	set(WIN32_VERSION "${LIB_VERSION_MAJOR},${LIB_VERSION_MINOR},${LIB_VERSION_PATCH}")
-	set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
+ set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
```
endrift, line 215: Don't randomly change indentation.

CMakePresets.json
```diff
@@ -1,21 +1,30 @@
 {
-    "version": 3,
```
endrift, line 2: Don't randomly reindent entire files. It adds unnecessary churn that makes traversing git history and merges difficult.

vcpkg.json
```diff
@@ -0,0 +1,27 @@
...
+   ]
+  },
+  "libepoxy",
+  "libpng",
+  "libzip",
+  "lua",
+  "sdl2",
+  "sqlite3",
+  "qt5-base",
+  "liblzma",
```
endrift, line 18: This won't even be picked up. mGBA is hardcoded to use the internal liblzma.

chrisepstein:
need to fix appveyor using classic vcpkg install

CMakePresets.json
```diff
@@ -1,21 +1,30 @@
...
+    "CMAKE_EXPORT_COMPILE_COMMANDS": "ON"
+   }
+  },
+  {
+   "name": "debug",
+   "generator": "Ninja",
+   "cacheVariables": {
+    "CMAKE_BUILD_TYPE": "Debug",
+    "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
+    "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
```
endrift, line 26: Don't just add this to the new config and not any of the old ones. These should be kept in sync.

endrift (changes requested):
This PR has some pretty significant issues, many of which seem to stem from a lack of care for existing conventions. I can't even begin to think about merging it until those are fixed.

PR #3266, opened 2024-07-29
Load recent script

anthonydo8196:
Finishing up #2845

Added your recommended changes on Tam's commits. 

I believe the reason why we originally stored the most recently used script separately was because we noticed that if you loaded a script when you went on the Load recent script menu, it wouldn't update the mru list. But I believe I fixed that by calling appendMRU again in updateMRU. Not sure if I needed to add an if but I did just to be safe so that it would only call appendMRU if the file was successfully loaded.

Also should the name for button be something other than Load recent? It shares the same words as Load recent script

scriptingmenu


src/platform/qt/scripting/ScriptingView.cpp
```diff
@@ -39,6 +39,7 @@ ScriptingView::ScriptingView(ScriptingController* controller, ConfigController*
 
 	connect(m_ui.buffers->selectionModel(), &QItemSelectionModel::currentChanged, this, &ScriptingView::selectBuffer);
 	connect(m_ui.load, &QAction::triggered, this, &ScriptingView::load);
+	connect(m_ui.loadRecent, &QAction::triggered, this, &ScriptingView::loadRecent);
```
endrift, line 42: You should make sure this is disabled if the MRU is empty.

src/platform/qt/scripting/ScriptingView.ui
```diff
@@ -108,6 +109,11 @@
    <property name="text">
     <string>Load script...</string>
    </property>
+  </action>
+    <action name="loadRecent">
+   <property name="text">
+    <string>&amp;Load recent</string>
```
endrift, line 115: The name is a bit confusing. Perhaps "Load most recent"?

src/platform/qt/scripting/ScriptingView.cpp
```diff
@@ -105,7 +110,9 @@ void ScriptingView::updateMRU() {
 	m_ui.mru->clear();
 	for (const auto& fname : m_mruFiles) {
 		m_ui.mru->addAction(fname, [this, fname]() {
-			m_controller->loadFile(fname);
+			if(m_controller->loadFile(fname)) {
```
endrift, line 113:
```suggestion
			if (m_controller->loadFile(fname)) {
```

I had wondered why this didn't work, but didn't look into it. Thanks for fixing that!

endrift (changes requested):
Some minor nits. Looks good otherwise.

anthonydo8196:
Added all the changes.  

It looks like this when the MRU list is empty. 
loadActionDisabled
Then once the user adds a script the action gets enabled and works properly.

src/platform/qt/scripting/ScriptingView.cpp
```diff
@@ -105,7 +110,19 @@ void ScriptingView::updateMRU() {
...
+	checkEmptyMRU();
 }
+
+
+void ScriptingView::checkEmptyMRU() {
+	if(m_mruFiles.isEmpty()) {
+		m_ui.loadMostRecent->setEnabled(false);
+	} else {
+		m_ui.loadMostRecent->setEnabled(true);
+	}
```
endrift, line 127:
```suggestion
	m_ui.loadMostRecent->setEnabled(!m_mruFiles.isEmpty());
```

endrift (approved):
Looks good, pending that one change.

src/platform/qt/scripting/ScriptingView.cpp
anthonydo8196: Ah that is much simpler  and cleaner.

PR #3314, opened 2024-10-12
Qt: handle newlines in TextBuffer print()

ahigerd:
User "Wombat" observed that printing strings containing newlines into a text buffer would word-wrap as if the newlines weren't there. Turns out, ScriptingTextBuffer::print() just wasn't handling newlines at all. This PR causes print() to split the input on newlines and add each section as a separate line in the document.

### DIFF
--- a/src/platform/qt/scripting/ScriptingTextBuffer.cpp
+++ b/src/platform/qt/scripting/ScriptingTextBuffer.cpp
@@ -51,26 +51,25 @@ void ScriptingTextBuffer::setBufferName(const QString& name) {
 
 void ScriptingTextBuffer::print(const QString& text) {
 	QMutexLocker locker(&m_mutex);
-	QString split(text);
-	m_shim.cursor.beginEditBlock();
-	while (m_shim.cursor.positionInBlock() + split.length() > m_dims.width()) {
-		int cut = m_dims.width() - m_shim.cursor.positionInBlock();
+	for (QString split : text.split('\n')) {
+		while (m_shim.cursor.positionInBlock() + split.length() > m_dims.width()) {
+			int cut = m_dims.width() - m_shim.cursor.positionInBlock();
+			if (!m_shim.cursor.atBlockEnd()) {
+				m_shim.cursor.movePosition(QTextCursor::EndOfBlock, QTextCursor::KeepAnchor);
+			}
+			m_shim.cursor.insertText(split.left(cut));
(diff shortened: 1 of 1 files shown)


src/platform/qt/scripting/ScriptingTextBuffer.cpp
```diff
@@ -51,26 +51,27 @@ void ScriptingTextBuffer::setBufferName(const QString& name) {
...
 void ScriptingTextBuffer::print(const QString& text) {
 	QMutexLocker locker(&m_mutex);
-	QString split(text);
-	m_shim.cursor.beginEditBlock();
-	while (m_shim.cursor.positionInBlock() + split.length() > m_dims.width()) {
-		int cut = m_dims.width() - m_shim.cursor.positionInBlock();
-		if (!m_shim.cursor.atBlockEnd()) {
-			m_shim.cursor.movePosition(QTextCursor::EndOfBlock, QTextCursor::KeepAnchor);
+	for (QString split : text.split('\n')) {
+		m_shim.cursor.beginEditBlock();
```
endrift, line 55: Shouldn't this be outside the loop?

ahigerd: I don't _think_ so? We want to create a new block for each line being inserted, don't we?

ahigerd: Hm. Actually, reading the docs, why do you even need this? You don't support undo/redo in text buffers and that's all this does.

endrift: I was using this to try to make the whole edit block "atomic" to prevent it from doing any work in the middle until it's committed. I have no idea if it works that way. Feel free to remove it if it doesn't.

ahigerd: It doesn't work that way. It makes the edit block act as a single unit from an undo/redo perspective, but the thing that makes a series of edits atomic in terms of external updates is the event loop -- the various QTextDocument functions set a dirty flag on the associated QTextEdit to prompt it to redo the layout next time control returns to the event loop, much like how multiple calls to `QWidget::update()` only trigger a single `paintEvent`.

ahigerd: Rebased and pushed.

PR #3329, opened 2024-10-30
Qt: Ensure that the correct OpenGL variant is version checked

Jan200101:
Without this the code compared the desired OpenGLES profile version against the core profile version which in my case succeeded and later on failed when trying to create a context.

Setting the renderable type ensures the correct OpenGL variant is checked.

glxinfo output:
```
    Max core profile version: 4.6
    Max compat profile version: 4.6
    Max GLES1 profile version: 1.1
    Max GLES[23] profile version: 3.2
```


Jan200101:
after checking again this is clearly wrong

PR #3333, opened 2024-11-04
Fix various macOS related problems

yeah-its-gloria:
command q addictions go crazy sometimes

- Disable a duplicate libraries warning that isn't needed
- Quit using QApplication::quit instead of QWidget::close to prevent closing inside a dialog from crashing the application
- Allow Qt 6 builds to use std::filesystem if the proper macOS SDK version is present
- Stop looking for QtMultimedia plugins if Qt 6 is being used for macOS builds

### DIFF
--- a/src/platform/qt/Window.cpp
+++ b/src/platform/qt/Window.cpp
@@ -1422,7 +1422,7 @@ void Window::setupMenu(QMenuBar* menubar) {
 #endif
 
 	m_actions.addAction(tr("About..."), "about", openTView<AboutScreen>(), "file")->setRole(Action::Role::ABOUT);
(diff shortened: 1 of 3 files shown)


src/platform/qt/Window.cpp
```diff
@@ -1422,7 +1422,7 @@ void Window::setupMenu(QMenuBar* menubar) {
 #endif
 
 	m_actions.addAction(tr("About..."), "about", openTView<AboutScreen>(), "file")->setRole(Action::Role::ABOUT);
-	m_actions.addAction(tr("E&xit"), "quit", static_cast<QWidget*>(this), &QWidget::close, "file", QKeySequence::Quit)->setRole(Action::Role::QUIT);
+	m_actions.addAction(tr("E&xit"), "quit", &QApplication::quit, "file", QKeySequence::Quit)->setRole(Action::Role::QUIT);
```
ahigerd, line 1425: This gets a +1 from me, because I noticed that the Exit command doesn't actually exit mGBA if you have multiplayer windows open.

PR #2649, opened 2022-09-01
Add AppStream

JakobDev:
This PR adds a [AppStream](https://www.freedesktop.org/software/appstream/docs/) file for Linux. AppStream contains the data for Software centers on Linux (e.g.g Gnome Software or KDE Discover) like Description, Screenshots, License, Links etc. If a App provides no AppStream file, the data from the .desktop files are used, which is missing a lot of things of course.

This AppStream is currently missing some screenshots, as I can't find some good. Please provide the Link to at least one Screenshot, so I can add them before merging.

PR #3340, opened 2024-11-12
Add AppStream Metainfo generation

Jan200101:
Alternative to #2649

Uses the same metainfo file but fills out the release tags with git tags (when possible)
links to the git tags aren't included but are trivial to be added back

I've put the logic directly where the desktop file install is done, unsure if this is the best place to put it.

<details>
<summary> appstreamcli validate qt/io.mgba.mGBA.metainfo.xml </summary>

I: io.mgba.mGBA:6: developer-name-tag-deprecated
I: io.mgba.mGBA:~: developer-info-missing

✔ Validation was successful: infos: 2, pedantic: 1
</details>


endrift:
I am amenable to this PR, but I think first I need a better understanding of how AppStream is used. For example, I notice GBA stuff is listed in that, but GB/C is missing. Will that pose an impediment to actually using it?

Jan200101:
I didn't author the metainfo file but it seems to be just copied from the desktop file

the provides mediatype fields are intended to provide the same MimeType in the desktop file, if you want I can add the common mimetypes for gameboy and gbc games to both the desktop and the appstream file: `application/x-gameboy-rom;application/x-gameboy-color-rom;application/x-gameboy-isx`

AppStream is generally intended to be a distro independent format for providing metadata about a project, such as links to various resources, screenshots, version info, etc.

Its most commonly used in store type applications like Gnome Software or Discover and is also used by flathubs web interface

https://flathub.org/apps/io.mgba.mGBA
https://github.com/flathub/io.mgba.mGBA/blob/master/io.mgba.mGBA.metainfo.xml

endrift:
I added some of those to the .desktop file, but I excluded the isx one as that seems to be a specific format that mGBA doesn't recognize. If you could resync with the .desktop file I'll merge this.

src/platform/qt/CMakeLists.txt
```diff
@@ -471,6 +471,35 @@ set(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}" PARENT_SCOPE)
...
+			list(APPEND RELEASE_LIST "<release version=\"${TAG_REF}\" date=\"${TAG_DATE}\" />")
+		endforeach()
+
+		list(LENGTH RELEASE_LIST RELEASE_COUNT)
+		if (NOT RELEASE_COUNT EQUAL 0)
+			list(JOIN RELEASE_LIST "\n" RELEASE_STR)
+			set(APPSTREAM_RELEASES "<releases>\n${RELEASE_STR}\n</releases>")
+		endif()
+	endif()
+	message(${CMAKE_SOURCE_DIR}/res/io.mgba.mGBA.metainfo.xml.in )
```
endrift, line 500: Please remove this debug print

endrift (approved):

endrift:
@Jan200101 Can you update this PR so I can merge it please?

PR #3350, opened 2024-11-19
Qt: honor control characters in ScriptingTextBuffer::print

ahigerd:
My [previous PR](https://github.com/mgba-emu/mgba/pull/3314) fixed line wrapping for strings with embedded newlines, but in the process it completely broke newline handling.

This PR makes sure that the cursor moves to the next block / a new block after each line.

### DIFF
--- a/src/platform/qt/scripting/ScriptingTextBuffer.cpp
+++ b/src/platform/qt/scripting/ScriptingTextBuffer.cpp
@@ -49,26 +49,70 @@ void ScriptingTextBuffer::setBufferName(const QString& name) {
 	emit bufferNameChanged(name);
 }
 
+void ScriptingTextBuffer::lineBreak() {
+	bool nextBlockExists = m_shim.cursor.movePosition(QTextCursor::NextBlock, QTextCursor::MoveAnchor, 1);
+	if (!nextBlockExists) {
+		m_shim.cursor.movePosition(QTextCursor::EndOfBlock, QTextCursor::MoveAnchor, 1);
+		m_shim.cursor.insertBlock();
+	}
+}
+
+void ScriptingTextBuffer::carriageReturn() {
+	m_shim.cursor.movePosition(QTextCursor::StartOfLine, QTextCursor::MoveAnchor, 1);
+}
+
+void ScriptingTextBuffer::tab() {
+	QTextCursor& cursor = m_shim.cursor;
+	int column = cursor.positionInBlock();
+	int move = tabStop - (column % tabStop) + 1;
+	if (column + move >= m_dims.width()) {
+		lineBreak();
+	} else if (column + move <= cursor.block().length()) {
+		cursor.movePosition(QTextCursor::Right, QTextCursor::MoveAnchor, move - 1);
+	} else {
+		move = column + move - cursor.block().length();
+		cursor.movePosition(QTextCursor::EndOfLine, QTextCursor::MoveAnchor, 1);
+		cursor.insertText(QString(move, ' '));
+	}
+}
+
+void ScriptingTextBuffer::insertString(const QString& text) {
+	QTextCursor& cursor = m_shim.cursor;
+	if (cursor.positionInBlock() >= m_dims.width()) {
+		lineBreak();
+	}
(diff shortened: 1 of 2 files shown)


ahigerd:
After some further discussion in Discord, I've decided I don't want to submit this PR as-is -- it fixes the bug, but the strategy for how to fix it is inelegant and leaves bugs pertaining to `\r` lingering.

ahigerd:
Rewritten -- and added support for `\t` while I was at it.

src/platform/qt/scripting/ScriptingTextBuffer.cpp
```diff
@@ -49,26 +49,45 @@ void ScriptingTextBuffer::setBufferName(const QString& name) {
...
+void ScriptingTextBuffer::lineBreak() {
+	bool nextBlockExists = m_shim.cursor.movePosition(QTextCursor::NextBlock, QTextCursor::MoveAnchor, 1);
+	if (!nextBlockExists) {
+		m_shim.cursor.movePosition(QTextCursor::EndOfBlock, QTextCursor::MoveAnchor, 1);
+		m_shim.cursor.insertBlock();
+	}
+}
+
 void ScriptingTextBuffer::print(const QString& text) {
+	static const int tabStop = 4;
```
endrift, line 61: Let's move this to a class constant.

src/platform/qt/scripting/ScriptingTextBuffer.cpp
```diff
@@ -49,26 +49,45 @@ void ScriptingTextBuffer::setBufferName(const QString& name) {
...
-			}
-			m_shim.cursor.insertText(split.left(cut));
-			if (m_shim.cursor.atEnd()) {
-				m_shim.cursor.insertBlock();
+
+	QTextCursor& cursor = m_shim.cursor;
+	int lineWidth = m_dims.width();
+
+	for (const QChar& ch : text) {
+		if (ch == '\t') {
```
endrift, line 68: Do we want to do this with a switch statement? (_Can_ we do this with a switch statement?)

src/platform/qt/scripting/ScriptingTextBuffer.cpp
```diff
@@ -49,26 +49,45 @@ void ScriptingTextBuffer::setBufferName(const QString& name) {
...
+		} else if (ch == '\n') {
+			lineBreak();
+		} else if (ch == '\r') {
+			cursor.movePosition(QTextCursor::StartOfLine, QTextCursor::MoveAnchor, 1);
+		} else {
+			if (cursor.positionInBlock() >= lineWidth) {
+				lineBreak();
+			}
+			cursor.movePosition(QTextCursor::Right, QTextCursor::KeepAnchor, 1);
+			cursor.insertText(ch);
```
endrift, line 89: Inserting characters one byte at a time seems...potentially slow. Have you made sure that this doesn't cause performance issues?

endrift (commented):
Nothing major looks wrong, but I have a few concerns.

src/platform/qt/scripting/ScriptingTextBuffer.cpp
ahigerd: It COULD be done with a switch statement but I'm not sure it would make the code look any better.

ahigerd: Well... After refactoring to fix the performance issue, maybe it would help.

src/platform/qt/scripting/ScriptingTextBuffer.cpp
ahigerd: Done.

src/platform/qt/scripting/ScriptingTextBuffer.cpp
ahigerd: What do you think?

src/platform/qt/scripting/ScriptingTextBuffer.cpp
ahigerd: Refactored. Performance delta against master is less than 10% (compared to 1000% from the old version).

PR #3351, opened 2024-11-20
Shader configuration improvements

ahigerd:
Killing four birds with one stone:

* Shader uniforms haven't saved since 2017 because the code was written to assume the shader settings had its own OK button. I could have addressed this with a direct function call but I thought a `saveSettingsRequested()` signal would allow future settings panels to be self-contained.
* Unlike basically everything else in the settings window, loading a shader saves the user's choice immediately. This means the Cancel button doesn't undo the action. This PR will still apply the shader changes immediately but it won't save the settings to disk until the user clicks OK or Apply, and the shader changes will be reverted upon clicking Cancel.
* It's kind of annoying to work on shader development with a packaged release of mGBA because you can't select an unpacked shader and you don't have the build system to pack it for you. With this PR you can now select the manifest.ini out of an unpacked shader and it'll load just fine.
* Shader load errors fail silently in the current version of mGBA. This PR adds a simple warning message.

### DIFF
--- a/src/platform/qt/ShaderSelector.cpp
+++ b/src/platform/qt/ShaderSelector.cpp
@@ -16,6 +16,7 @@
 #include <QFileDialog>
 #include <QFormLayout>
 #include <QGridLayout>
+#include <QMessageBox>
 #include <QSpinBox>
 
 #include <mgba/core/version.h>
@@ -50,7 +51,7 @@ ShaderSelector::~ShaderSelector() {
 }
 
 void ShaderSelector::saveSettings() {
-	QString oldPath = config->getOption("shader");
+	QString oldPath = m_config->getOption("shader");
 	if (oldPath != m_shaderPath) {
 		if (m_shaderPath.isEmpty()) {
 			clearShader(true);
@@ -88,7 +89,8 @@ void ShaderSelector::selectShader() {
 #if !defined(USE_LIBZIP) && !defined(USE_MINIZIP)
 	QString name = GBAApp::app()->getOpenDirectoryName(this, tr("Load shader"), path.absolutePath());
 #else
-	QString name = GBAApp::app()->getOpenFileName(this, tr("Load shader"), "mGBA Shaders (*.shader)", path.absolutePath());
+	QString filters = QStringLiteral("%1 (*.shader manifest.ini)").arg(tr("mGBA Shaders"));
+	QString name = GBAApp::app()->getOpenFileName(this, tr("Load shader"), filters, path.absolutePath());
 #endif
 	if (!name.isNull()) {
 		loadShader(name);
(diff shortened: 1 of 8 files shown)


src/platform/qt/ShaderSelector.cpp
```diff
@@ -66,33 +89,47 @@ void ShaderSelector::selectShader() {
 #if !defined(USE_LIBZIP) && !defined(USE_MINIZIP)
 	QString name = GBAApp::app()->getOpenDirectoryName(this, tr("Load shader"), path.absolutePath());
 #else
-	QString name = GBAApp::app()->getOpenFileName(this, tr("Load shader"), "mGBA Shaders (*.shader)", path.absolutePath());
+	QString name = GBAApp::app()->getOpenFileName(this, tr("Load shader"), "mGBA Shaders (*.shader manifest.ini)", path.absolutePath());
```
ahigerd, line 92: Hmm... This really ought to be `QStringLiteral("%1 (*.shader manifest.ini)").arg(tr("mGBA Shaders"))`. Might as well work that in.

src/platform/qt/ShaderSelector.h
```diff
@@ -28,17 +28,19 @@ Q_OBJECT
...
 	void selectShader();
-	void loadShader(const QString& path);
-	void clearShader();
+	void loadShader(const QString& path, bool saveToSettings = false);
+	void clearShader(bool saveToSettings = false);
 	void buttonPressed(QAbstractButton*);
 
 signals:
-	void saved();
+	void saveSettingsRequested();
```
ahigerd, line 43: Strictly speaking, just a cosmetic change, but I think it more accurately reflects the semantics of the signal.

src/platform/qt/ShaderSelector.cpp
```diff
@@ -66,33 +89,48 @@ void ShaderSelector::selectShader() {
 #if !defined(USE_LIBZIP) && !defined(USE_MINIZIP)
 	QString name = GBAApp::app()->getOpenDirectoryName(this, tr("Load shader"), path.absolutePath());
 #else
-	QString name = GBAApp::app()->getOpenFileName(this, tr("Load shader"), "mGBA Shaders (*.shader)", path.absolutePath());
+	QString filters = QStringLiteral("%1 (*.shader manifest.ini)").arg(tr("mGBA Shaders")),
```
endrift, line 92:
```suggestion
	QString filters = QStringLiteral("%1 (*.shader manifest.ini)").arg(tr("mGBA Shaders"));
```

src/platform/qt/ShaderSelector.h
endrift: Sure, sounds good.

src/platform/qt/ShaderSelector.cpp
ahigerd: Now how did that happen? I compiled and ran this. O.o; Did I fat-finger a vim command while preparing the PR?

PR #3367, opened 2024-12-07
Update libretro.h with Swedish

yeager:


yeager:
Swedish translation in mGBA is now at 68%

PR #3376, opened 2024-12-16
Scripting: Fix Lua 5.1/LuaJIT and Lua 5.2 problems

nuive:
What's added:

-New option available for CMake USE_LUA option. Setting it to "JIT" will force CMake to look for LuaJIT when building.

What's fixed:

-Lua 5.1/JIT couldn't even load a script.
-Updated lua pairs/ipairs handling (Lua 5.1/JIT and 5.2 weren't able to manage things like mSTList iteration). Now pairs/ipairs work like Lua 5.4 (except for a lua_callk which stays lua_call).
-Custom objects show their propper names using Lua versions prior to 5.3.

What's not fixed:

I didn't find anything else needing a fix regarding specific Lua versions.


src/platform/cmake/FindLuaJIT.cmake
endrift, line 1: What is the provenance of this file? There's no info in the file itself, and it looks like it comes from somewhere

src/script/engines/lua.c
```diff
@@ -215,6 +215,66 @@ static const int _mScriptSocketNumErrors = sizeof(_mScriptSocketErrors) / sizeof
...
+				lua_pushstring(lua, (lua_toboolean(lua, idx) ? "true" : "false"));
+				break;
+			case LUA_TNIL:
+				lua_pushliteral(lua, "nil");
+				break;
+			default:
+				luaL_getmetafield(lua, idx, "__name");
+				int tt = lua_type(lua, -1);
+				const char* name = (tt == LUA_TSTRING) ? lua_tostring(lua, -1) :
+				                                         luaL_typename(lua, idx);
```
endrift, line 269: Don't copy source code directly from Lua, please. The license may permit this, but it'd be much better to replace the one call to this function with something like what Lua 5.1 does in its print function: [it just calls tostring on the argument directly](https://github.com/lua/lua/blob/v5.1/lbaselib.c#L34) to coerce it.

src/script/engines/lua.c
```diff
@@ -215,6 +215,66 @@ static const int _mScriptSocketNumErrors = sizeof(_mScriptSocketErrors) / sizeof
...
+	return (lua_type(lua, -1) == LUA_TNIL) ? 1 : 2;
+}
+
+static int _luaIpairs(lua_State* lua) {
+	luaL_checkany(lua, 1);
+	lua_pushcfunction(lua, _luaIpairsIter);
+	lua_pushvalue(lua, 1);
+	lua_pushinteger(lua, 0);
+	return 3;
+}
```
endrift, line 246: Isn't pairs/ipairs only in 5.2/5.3, not 5.1?

src/script/engines/lua.c
```diff
@@ -226,6 +286,7 @@ static const int _mScriptSocketNumErrors = sizeof(_mScriptSocketErrors) / sizeof
 #define lua_pushglobaltable(L) lua_pushvalue(L, LUA_GLOBALSINDEX)
 #endif
 
+
```
endrift, line 289: Please don't add a stray newline

src/script/engines/lua.c
```diff
@@ -415,6 +482,14 @@ struct mScriptEngineContext* _luaCreate(struct mScriptEngine2* engine, struct mS
 	lua_getglobal(luaContext->lua, "require");
 	luaContext->require = luaL_ref(luaContext->lua, LUA_REGISTRYINDEX);
 
+#if LUA_VERSION_NUM < 503
+	lua_pushcfunction(luaContext->lua, _luaIpairs);
+	lua_setglobal(luaContext->lua, "ipairs");
+
+	lua_pushcfunction(luaContext->lua, _luaPairs);
+	lua_setglobal(luaContext->lua, "pairs");
```
endrift, line 490: Aren't pairs/ipairs builtins in 5.2?

src/script/engines/lua.c
```diff
@@ -976,11 +1051,19 @@ bool _luaLoad(struct mScriptEngineContext* ctx, const char* filename, struct VFi
 		// Make the old _ENV the __index in the metatable
 		lua_newtable(luaContext->lua);
 		lua_pushliteral(luaContext->lua, "__index");
+#if LUA_VERSION_NUM >= 502
 		lua_getupvalue(luaContext->lua, -4, 1);
+#else
+		lua_getfenv(luaContext->lua, -4);
+#endif
```
endrift, line 1058: lua_getupvalue exists with the same signature in 5.1. What's different in 5.1 that this is needed?

endrift (changes requested):
Tagging in @ahigerd for this. I'm not sure why some of these changes were made, and I'm concerned about the provenance of some of the code.

src/platform/cmake/FindLuaJIT.cmake
CasualPokePlayer: It looks like it's coming mostly from CMake's [FindLua.cmake](https://github.com/Kitware/CMake/blob/3e2094fb9b35af4c262959117d203644d8f396ca/Modules/FindLua.cmake) file, but it's changed kind of weirdly for LuaJIT. It appears to be mostly just find and replace `lua` with `luajit` and such, then some changes to handle LuaJIT versions (2.1 and 2.0?), and there's kind of weird remanents remaining with lua functions getting a luajit replacement, but then also leaving in the older functions. The license is removed for some reason, and [even a comment mentioning FindLua was removed here](https://github.com/Kitware/CMake/blob/3e2094fb9b35af4c262959117d203644d8f396ca/Modules/FindLua.cmake#L158) for some reason.

CMakeLists.txt
```diff
@@ -768,7 +768,9 @@ endif()
 if(ENABLE_SCRIPTING)
 	list(APPEND ENABLES SCRIPTING)
 	find_feature(USE_JSON_C "json-c")
-	if(NOT USE_LUA VERSION_LESS 5.1)
+	if(USE_LUA STREQUAL "JIT")
+		find_feature(USE_LUA "Lua${USE_LUA}")
```
ahigerd, line 772: Might as well just write `LuaJIT` here since it's guaranteed to evaluate to that after the previous line.

src/platform/cmake/FindLuaJIT.cmake
```diff
@@ -0,0 +1,319 @@
...
+
+cmake_policy(PUSH)  # Policies apply to functions at definition-time
+cmake_policy(SET CMP0012 NEW)  # For while(TRUE)
+
+unset(_luajit_include_subdirs)
+unset(_luajit_library_names)
+unset(_luajit_append_versions)
+
+# this is a function only to have all the variables inside go away automatically
+function(_lua_get_versions)
```
ahigerd, line 48: This is irrelevant for LuaJIT as it always acts like 5.1.

src/platform/cmake/FindLuaJIT.cmake
```diff
@@ -0,0 +1,319 @@
...
+
+function(_lua_get_header_version)
+  unset(LUA_VERSION_STRING PARENT_SCOPE)
+  set(_hdr_file "${LUAJIT_INCLUDE_DIR}/lua.h")
+
+  if (NOT EXISTS "${_hdr_file}")
+    return()
+  endif ()
+
+  # At least 5.[012] have different ways to express the version
```
ahigerd, line 153: We know that it's 5.1, so simplify this.

src/script/engines/lua.c
ahigerd: No, they're in 5.1. You're thinking of the metamethods.

src/platform/cmake/FindLuaJIT.cmake
nuive: Sorry, it's bassed on CMake FindLua.cmake. I forgot to restore/modify comments.

The old Lua functions are there because of library name (lua51). I left them thinking LuaJIT could be built using different Lua versions in the future, but nowadays they could be removed and replaced those calls with "lua51" or "lua5.1" constants.

src/script/engines/lua.c
ahigerd: They're builtins in 5.1 and LuaJIT so I'm not sure what this is for.

src/script/engines/lua.c
ahigerd: This _is_ the breaking change between 5.1 and 5.2+ -- in 5.2+ the first upvalue of a function is the lexical environment (if it exists -- you cannot assume a function inherits a lexical environment), but in 5.1 the lexical environment is managed by getfenv/setfenv.

src/script/engines/lua.c
```diff
@@ -390,6 +451,8 @@ struct mScriptEngineContext* _luaCreate(struct mScriptEngine2* engine, struct mS
 
 	luaL_newmetatable(luaContext->lua, "mSTStruct");
 #if LUA_VERSION_NUM < 502
+	lua_pushliteral(luaContext->lua, "mSTStruct");
```
ahigerd, line 454: I become tempted to refactor these to reduce just how much repetition is involved here, but that's kinda out of scope for this PR.

src/script/engines/lua.c
nuive: That's right. In fact these modifications add __pairs metamethod to Lua 5.1, and removes the LUA_TTABLE type restriction for pairs/ipairs.

src/script/engines/lua.c
nuive: These lines are added in order to replace the original functions with the ones defined in this file. I don't know if there's a better method for this.

ahigerd: The only thing that comes to mind is that PUC-Rio had removed ipairs from prerelease versions of 5.2 before re-adding it in response to overwhelming outcry from the community. If this code is derived from some _ancient_ 5.1/5.2 compatibility hack then that might explain it, but pairs was never on the chopping block so I really don't know.

src/script/engines/lua.c
ahigerd: What's the use case? (And if you're implementing this, why not add __ipairs too?)

src/script/engines/lua.c
nuive: Yes, I'm definitely missing something here. I agree with this needing refactorization but I don't know how at the momment.

src/script/engines/lua.c
nuive: -Lua 5.1: pairs and ipairs work only with table objects. Also they don't check for __ respective metamethods.
-Lua 5.2: pairs and ipairs work only with table objects. They check for __ respective metamethods.
-Lua 5.3: pairs check for __pairs metamethod. ipairs check for __ipairs metamethod only in compatibility mode. Object type is irrelevant (only checks if its a table when using luaB_next iterator (the global next function).
-Lua 5.4: Only pairs check for its __ metamethod. Like in 5.3 object type is irrelevant (only checks if its a table when using luaB_next iterator (the global next function).

So the approach is done with Lua 5.4's functionality in mind.

It's used for mSTList iteration and mSTTable __pairs metamethod.

ahigerd: If you're trying to support 5.4 semantics then you need to add support for __len to the ipairs implementation.

src/script/engines/lua.c
ahigerd: (I guess that should probably be declared `static`.)

ahigerd:
I think the implementation of `_luaLoad` is missing some fenv fixes.

src/script/engines/lua.c
nuive: Oops... You're right!

I'll fix that.

nuive: Okay, this seems different... I think there's no easy way of overriding # operator, neither is a replaceable function like pairs and ipairs.

nuive:
> I think the implementation of `_luaLoad` is missing some fenv fixes.

Like what?

src/script/engines/lua.c
ahigerd: You're right, there isn't, but if you're going to provide the 5.4 semantics for ipairs then you can check for the metamethod manually.

ahigerd:
> > I think the implementation of `_luaLoad` is missing some fenv fixes.
> 
> Like what?

```c
      // Create new _ENV
      lua_newtable(luaContext->lua);
  
      // Make the old _ENV the __index in the metatable
      lua_newtable(luaContext->lua);
      lua_pushliteral(luaContext->lua, "__index");
      lua_getupvalue(luaContext->lua, -4, 1);
      lua_rawset(luaContext->lua, -3);
  
      lua_pushliteral(luaContext->lua, "__newindex");
      lua_getupvalue(luaContext->lua, -4, 1);
      lua_rawset(luaContext->lua, -3);
  
      lua_setmetatable(luaContext->lua, -2);
  
      lua_pushliteral(luaContext->lua, "script");
      lua_newtable(luaContext->lua);
```
This is _ENV manipulation which means it needs to be handled using setfenv instead. It's probably a little nontrivial to do correctly but if you don't do it then `require()` won't behave correctly.

src/platform/cmake/FindLuaJIT.cmake
endrift: I believe that's not true for LuaJIT 2.1, but also this file has a lot of stuff that definitely mishandles the versioning as it expects Lua versions

ahigerd: LuaJIT 2.1 still always acts like 5.1 in terms of semantics; it just supports some 5.2+ features as extensions.

nuive: That's because inside LuaJIT are defined LUAJIT_VERSION (2.x) and LUA_VERSION (5.1). LUA_VERSION is used into the library name. I've simplified the cmake file. As previously said LuaJIT'll always be Lua 5.1, all non LuaJIT references will be removed once I commit hte new changes.

nuive:
> ```c
>       // Create new _ENV
>       lua_newtable(luaContext->lua);
>   
>       // Make the old _ENV the __index in the metatable
>       lua_newtable(luaContext->lua);
>       lua_pushliteral(luaContext->lua, "__index");
>       lua_getupvalue(luaContext->lua, -4, 1);
>       lua_rawset(luaContext->lua, -3);
>   
>       lua_pushliteral(luaContext->lua, "__newindex");
>       lua_getupvalue(luaContext->lua, -4, 1);
>       lua_rawset(luaContext->lua, -3);
>   
>       lua_setmetatable(luaContext->lua, -2);
>   
>       lua_pushliteral(luaContext->lua, "script");
>       lua_newtable(luaContext->lua);
> ```
> 
> This is _ENV manipulation which means it needs to be handled using setfenv instead. It's probably a little nontrivial to do correctly but if you don't do it then `require()` won't behave correctly.

Yes, I know. In fact mGBA built with Lua 5.1 breaks if this is not handled propperly.

However if you look at the submited changes there's this:

```c
		// Create new _ENV
		lua_newtable(luaContext->lua);

		// Make the old _ENV the __index in the metatable
		lua_newtable(luaContext->lua);
		lua_pushliteral(luaContext->lua, "__index");
#if LUA_VERSION_NUM >= 502
		lua_getupvalue(luaContext->lua, -4, 1);
#else
		lua_getfenv(luaContext->lua, -4);
#endif
		lua_rawset(luaContext->lua, -3);

		lua_pushliteral(luaContext->lua, "__newindex");
#if LUA_VERSION_NUM >= 502
		lua_getupvalue(luaContext->lua, -4, 1);
#else
		lua_getfenv(luaContext->lua, -4);
#endif
		lua_rawset(luaContext->lua, -3);

		lua_setmetatable(luaContext->lua, -2);

		lua_pushliteral(luaContext->lua, "script");
		lua_newtable(luaContext->lua);

		if (dirname[0]) {
			lua_pushliteral(luaContext->lua, "require");
			lua_pushstring(luaContext->lua, dirname);
			lua_pushcclosure(luaContext->lua, _luaRequireShim, 1);
			lua_rawset(luaContext->lua, -5);

			lua_pushliteral(luaContext->lua, "dir");
			lua_pushstring(luaContext->lua, dirname);
			lua_rawset(luaContext->lua, -3);
		}

		if (name[0] == '@') {
			lua_pushliteral(luaContext->lua, "path");
			lua_pushstring(luaContext->lua, &name[1]);
			lua_rawset(luaContext->lua, -3);
		}

		lua_rawset(luaContext->lua, -3);
#if LUA_VERSION_NUM >= 502
		lua_setupvalue(luaContext->lua, -2, 1);
#else
		lua_setfenv(luaContext->lua, -2);
#endif
		luaContext->func = luaL_ref(luaContext->lua, LUA_REGISTRYINDEX);
		return true;
```

Woudln't this be enough? Am I missing any (set/get)fenv?

ahigerd:
Hmm... I may have misread the code. That might be correct after all.

src/script/engines/lua.c
nuive: I think I'm not getting this... What connection is between __len and ipairs? ipairs in Lua 5.4 is different from Lua 5.1, but I think I've adjusted it propperly in _luaIpairsIter and _luaIpairs functions.

ahigerd: *checks documentation* Oh. Huh. I was under the impression that ipairs was semantically equivalent to `for i = 1, #t do` in 5.4 since the __ipairs metamethod was removed, but apparently I was mistaken.

nuive:
I think I made all requested changes...

src/script/engines/lua.c
```diff
@@ -1032,6 +1063,39 @@ const char* _luaGetError(struct mScriptEngineContext* context) {
...
+	lua_pushstring(lua, name);
+	lua_setfield(lua, -2, "__name");
+	luaL_register(lua, NULL, funcs);
+#else
+	luaL_setfuncs(lua, funcs, 0);
+#endif
+	lua_pop(lua, 1);
+}
+
+const char* _luaTolstring(lua_State* lua, int idx, size_t* len) {
```
ahigerd, line 1078: @endrift Bikeshed question: Should this function wrap `luaL_tostring` for 5.3+, or should we polyfill `luaL_tostring` so that 5.3+ code compiles under 5.1/5.2?

(Tangentially related, but not a task for this PR: Is there a particular reason why the internal functions in this file aren't declared static?)

src/platform/cmake/FindLuaJIT.cmake
```diff
@@ -0,0 +1,225 @@
...
+_luajit_get_header_version()
+unset(_luajit_append_versions)
+
+if (LUAJIT_VERSION_STRING)
+  set(_luajit_library_names
+    luajit${LUAJIT_VERSION_MAJOR}${LUAJIT_VERSION_MINOR}
+    luajit${LUAJIT_VERSION_MAJOR}.${LUAJIT_VERSION_MINOR}
+    luajit-${LUAJIT_VERSION_MAJOR}.${LUAJIT_VERSION_MINOR}
+    luajit.${LUAJIT_VERSION_MAJOR}.${LUAJIT_VERSION_MINOR}
+    lua51
```
ahigerd, line 182: Does LuaJIT actually provide a `lua51` soname? That seems weird...

ahigerd (approved):
I think I'm basically satisfied?

src/platform/cmake/FindLuaJIT.cmake
nuive: Good point. That is an inconsistency.

I'm using a win64 docker and MinGW's LuaJIT has same library names as standard Lua 5.1. In unix that's not true, names would be something like luajit51, luajit5.1...

ahigerd: Hm, if that's the standard soname that's getting shipped for the package in MSYS2, then I guess it should stay as it is.

nuive: Well, I've added luajit51 variants in order to be able to build using unix systems. I don't know if I should left the luajit${LUAJIT_VERSION_MAJOR}${LUAJIT_VERSION_MINOR} variants into the list. I've found LuaJIT installations where the library was named something like libluajit2.1.so or something like that, but I supose those name conventions are customized. Should I remove ${LUAJIT_VERSION_MAJOR}${LUAJIT_VERSION_MINOR} variants from the list leaving only luajit51 and lua51 ones, or it's alrhigt like it is now? Currently possibly library names list has 12 items...

ahigerd:
@nuive It looks like you may have made a mistake with your branch handling here?

nuive:
> @nuive It looks like you may have made a mistake with your branch handling here?

Oops, I forgot about this... Yes, I modified some things because this was merged into master branch... I supose I should open a new PR referencing the new branch, am I right?

ahigerd:
@nuive Aye, that would be the way.

PR #3417, opened 2025-02-13
Qt: Fix use-after-free in shader settings

ahigerd:
User "Autofire" in Discord reported a crash when saving settings.

```
Thread 1 "mgba-qt" received signal SIGSEGV, Segmentation fault.
0x00007ffff2ba793b in QObject::disconnect (sender=0x555555eec800, signal=0x0, receiver=0x555558be4340, method=0x0)
    at /usr/src/debug/qt6-base/qtbase/src/corelib/kernel/qobject.cpp:3280
3280        const QMetaObject *smeta = sender->metaObject();
```

I tracked it down to a use-after-free: `Window::reloadDisplayDriver` calls `SettingsView::setShaderSelector` after destroying the old `ShaderSelector`, but the pointer in `SettingsView` didn't get cleared. Fortunately, QPointer exists for exactly this.

### DIFF
--- a/src/platform/qt/SettingsView.h
+++ b/src/platform/qt/SettingsView.h
@@ -7,10 +7,12 @@
 
 #include <QDialog>
 #include <QMap>
+#include <QPointer>
 #include <QTimer>
 
 #include "ColorPicker.h"
 #include "LogConfigModel.h"
+#include "ShaderSelector.h"
 
 #include <mgba/core/core.h>
 
@@ -25,7 +27,6 @@ namespace QGBA {
 class ConfigController;
 class InputController;
 class ShortcutController;
-class ShaderSelector;
 
 class SettingsView : public QDialog {
 Q_OBJECT
@@ -80,7 +81,7 @@ private:
 
 	ConfigController* m_controller;
 	InputController* m_input;
-	ShaderSelector* m_shader = nullptr;
+	QPointer<ShaderSelector> m_shader;
 	QLabel* m_dummyShader;
 	LogConfigModel m_logModel;
 	QTimer m_checkTimer;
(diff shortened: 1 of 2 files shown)


endrift:
I think a good question here is *why* that pointer didn't get cleared. I don't use QPointer anywhere else and tend to use either shared_ptr or actually keeping track of who's using it manually. I've been meaning to move more towards weak_ptr where applicable, too.

src/platform/qt/SettingsView.h
```diff
@@ -80,7 +81,7 @@ private slots:
 
 	ConfigController* m_controller;
 	InputController* m_input;
-	ShaderSelector* m_shader = nullptr;
+	QPointer<ShaderSelector> m_shader = nullptr;
```
endrift, line 84: Assuming we do use QPointer, the explicit initialization isn't needed anymore, as QPointer<T> isn't POD and the default constructor sets it to null.

ahigerd: Good point.

ahigerd:
@endrift It's dangerous to use `shared_ptr` and QObject parenting at the same time, and widgets necessarily use QObject parenting. It's also dangerous to depend on manual pointer management; that's _why_ smart pointers exist. (And why Rust exists.) QPointer really is the best option here.

(It would be nice if there was a stronger version of QPointer that would let you lock the object against destruction.)

endrift:
I absolutely agree about manual pointer management, which is why I try to avoid that, but I'm still curious why it wasn't getting cleared.

I might want to have a bit more of a conversation about this later after I log into Discord.

ahigerd:
I did a bit of digging and I discovered why:

https://github.com/mgba-emu/mgba/blob/master/src/platform/qt/Window.cpp#L1058-L1061

`reloadDisplayDriver` destroys the old one before creating the new one.

src/platform/qt/SettingsView.h
ahigerd: Fixed.

ahigerd:
... wait, that's what I said in the first place. ^^()

More specifically, it does so with nothing in the control flow in between that could possibly do anything about it.

endrift:
Oh, I see what the issue here is now. We don't clear the old one before deleting it.

PR #3438, opened 2025-03-16
Simplify Qt multimedia code

oltolm:
I think the unique pointers in `InputController` and `AudioProcessorQt` are unnecessary. I have refactored the code to not use unique pointers and simplified the code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.10)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/src/platform/cmake/")
 
 if(POLICY CMP0025)
--- a/src/core/scripting.c
+++ b/src/core/scripting.c
@@ -1343,6 +1343,8 @@ static uint8_t _readLuminance(struct GBALuminanceSource* luminance) {
 }
 #endif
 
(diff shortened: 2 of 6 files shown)


src/core/scripting.c
```diff
@@ -1343,6 +1343,8 @@ static uint8_t _readLuminance(struct GBALuminanceSource* luminance) {
 }
 #endif
 
+#pragma push_macro("CALLBACK")
```
endrift, line 1346: I have no idea what warning this fixes. Is it Windows-specific? Regardless, it doesn't belong in this PR.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.10)
```
endrift, line 1: Please rebase. I made this change a few days ago so it's not needed anymore.

endrift (changes requested):
While this does arguably simplify the AudioProcessorQt code, it does this by making a device and then effectively immediately deleting it as soon as the settings get applied. I'm not sure that's really a worthwhile change.

Further, I'm not sure the camera code is really simplified. You removed a variable, but in the process introduced the same setup-teardown-setup cycle. Also, we *really* don't want to create a camera early. There's no reason to instantiate one 99.9% of the time.

oltolm:
Yes, it's probably not worth it.

Issue #3234, opened 2024-06-17
Rumble should work even if 0x80000C8 is 0


afska:
Hi @endrift, thanks for fixing https://github.com/mgba-emu/mgba/issues/3232
I've just tested it and it works! However, I found a new issue:

[gba-rumble-example-no-0x80000C8-write.zip](https://github.com/user-attachments/files/15864590/gba-rumble-example-no-0x80000C8-write.zip)
This ROM works on the EZ Flash OmegaDE and insideGadgets carts, but mGBA doesn't autodetect its rumble features. Even when forcing it via override, the vibration doesn't work (it seems to be ignoring the Data and Direction port writes, based on the memory viewer).

According to GBATEK, this should work since the Control port only determines whether the addresses are readable or not.

endrift:
I'd like to check to make sure this works on official hardware, not just flash carts. Can you provide a multiboot build of the test ROM too?

afska:
Sure, here's a multiboot build:
[gba-rumble-example_mb.zip](https://github.com/user-attachments/files/15864810/gba-rumble-example_mb.zip)

Anders429:
Was this ever able to be verified on official hardware? I also observed this same behavior when experimenting with GPIO on a flash cart vs mGBA, but I don't have an official cart with rumble to try the multiboot ROM.

endrift:
I have yet to verify it. I have the hardware all set up to do it but have yet to do it.

PR #3453, opened 2025-04-06
Implement GPIO internal write latch

CasualPokePlayer:
Regardless of direction, a write to GPIO data actually succeeds fully with all 4 bits. However, this does not directly touch the pin state. It instead places it into an internal write only latch. This latch asserts bits onto the pin state if direction allows for such, otherwise the other side (e.g. RTC) will be the one asserting bits (but this does not end up touching the internal write latch).

The implementation here is likely not entirely accurate for direction changes from out to in (as that depends on each external device implementation), but it should be correct for in to out changes.

### DIFF
--- a/src/gba/cart/gpio.c
+++ b/src/gba/cart/gpio.c
@@ -49,6 +49,7 @@ void GBAHardwareInit(struct GBACartridgeHardware* hw, uint16_t* base) {
 
 void GBAHardwareReset(struct GBACartridgeHardware* hw) {
 	hw->readWrite = GPIO_WRITE_ONLY;
+	hw->writeLatch = 0;
 	hw->pinState = 0;
 	hw->direction = 0;
 	hw->lightCounter = 0;
@@ -64,6 +65,7 @@ void GBAHardwareReset(struct GBACartridgeHardware* hw) {
 void GBAHardwareClear(struct GBACartridgeHardware* hw) {
 	hw->devices = HW_NONE | (hw->devices & HW_GB_PLAYER_DETECTION);
 	hw->readWrite = GPIO_WRITE_ONLY;
+	hw->writeLatch = 0;
 	hw->pinState = 0;
 	hw->direction = 0;
 }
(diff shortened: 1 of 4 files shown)


src/gba/cart/gpio.c
```diff
@@ -74,16 +76,22 @@ void GBAHardwareGPIOWrite(struct GBACartridgeHardware* hw, uint32_t address, uin
...
+			hw->pinState = hw->writeLatch;
 		}
 		_readPins(hw);
 		break;
 	case GPIO_REG_DIRECTION:
 		hw->direction = value & 0xF;
+		if (!hw->p->vbaBugCompat) {
+			hw->pinState &= ~hw->direction;
+			hw->pinState |= hw->writeLatch & hw->direction;
+			_readPins(hw);
```
endrift, line 93: I don't think is this correct. I think the "write" will only affect on writing the data, not the direction.

endrift (commented):
While I think this is generally the right approach, I think you've got the latching backwards--what an external device would read is based on the latch, not the direction, I'm pretty sure. See also #3234. For simplicity, it might be best to split pinState into a "input" state and an "output" state instead of pinState and writeLatch.

src/gba/cart/gpio.c
CasualPokePlayer: I did test this on my own Emerald cart, changing the direction does in fact change the pin state.

CasualPokePlayer:
> While I think this is generally the right approach, I think you've got the latching backwards--what an external device would read is based on the latch, not the direction, I'm pretty sure. 

That wouldn't really make much sense, the internal latch here seems to be purely on the CPU (cartridge GPIO?) side, not the external side. The external side from testing appeared to not be able to touch this latch, and only reacted to its changes once direction allowed for this latch to assert bits (i.e. direction is going out).
> See also #3234.

That seems to be speaking of just the GPIO control register, which just affects whether the pin state is readable by the CPU, not external devices.

endrift:
I see, I hadn't looked at that in a while and forgot those were different registers. This PR seems mostly fine then. Please increment the savestate version though, as it repurposes an offset that had been used for something else previously.

endrift:
Might also make sense to reduce the size of the other uint16_t-but-actually-just-4-bit fields to uint8_t too.

src/gba/cart/gpio.c
CasualPokePlayer: Let's say you first just write 0 as the direction (i.e., all in, only external devices are asserting pin state here). You can proceed to write 0xF to data. This doesn't do anything to the pin state yet, as direction does not allow for such. If you proceed to then change the direction to 0xF (i.e., all out, only the CPU (or really, the write latch) is able to assert pin state here), then the pin state becomes 0xF, which external devices can react to.

PR #3454, opened 2025-04-07
Improve GBA RTC

CasualPokePlayer:
Implements various improvements to GBA RTC.

Primarily, this allows for RTC to actually be reset and set by games, instead of it being forced into matching host time. Time registers are now just incremented as host time increases, like other RTC implementations in mGBA (e.g. MBC3 / HuC3).

"Test mode" is also implemented to the degree it seems to do anything (which just seems to prevent commands from working, except exit test mode and reset commands). It is internally stored in bit 7 of the seconds byte, as documented in spec sheets (although this is impossible to verify it seems, due to get time commands no longer working).

Invalid / write-only commands now return a stream of 1-bits (rather than 0-bits), as on hardware. The PM flag is now properly implemented (so software can actually use 12 hour mode fully).

Only valid RTC control bits can actually be written now (unused bits are always 0; power-off bit is not settable but it is clearable).

The alarm command is "implemented" here as a stub (the alarm registers are just write only here, so nothing to do without alarms actually being implemented with cartridge IRQ; although I haven't verified yet if cartridge IRQ should actually be triggered or whatever).


src/gba/cart/gpio.c
```diff
@@ -112,11 +144,19 @@ void GBAHardwareInitRTC(struct GBACartridgeHardware* hw) {
...
-
-	hw->rtc.lastLatch = 0;
-	hw->rtc.offset = 0;
+	hw->rtc.time[RTC_MONTHS] = 1;
+	hw->rtc.time[RTC_DAYS] = 1;
+	// Normal first power-on would rather have the weekday at 0
+	// Under the usual 0 = Sunday, 6 would be the correct day for the GBA epoch (Jan 1 2000)
+	hw->rtc.time[RTC_WEEKDAY] = 6;
+	// 946684800 is unix time for the GBA epoch
+	hw->rtc.lastLatch = 946684800;
```
CasualPokePlayer, line 159: Note that this might be better to rather be the result from mktime rather (that would make it so initial RTC matches up with local time like before, if that is preferred).

endrift:
I can't look at this too closely at the moment, but I'd recommend looking at the RTC changes on the medudsa branch, as it might make sense to try to pull in some of that to reduce the distance.

CasualPokePlayer:
The medusa branch seems to not really be in much better shape (e.g. doesn't have setting RTC available), if anything it's in a much "worse" state as it appears to have just implemented some amalgamation of the GBA and DS's RTC within the GBA code?

src/gba/cart/gpio.c
CasualPokePlayer: With a little more thought, this isn't really a great way to do local time, since it wouldn't account for timezone changes due to DST or whatever.

An alternative is to just offset constantly within _rtcUpdateClock (like previous code did).

However, doing this at all isn't really desirable in every case. Using local time really only makes sense when real time is being used. For non-real time (like RTC_FIXED or RTC_FAKE_EPOCH, or maybe like BizHawk's custom impl for the purpose of recording movies) it doesnt make sense, as such values shouldnt need to be offset by local time (mGBA does end up offsetting them anyways, the user selection is offsetted by local time to "convert" it to UTC which is converted back to local time in the old code by localtime_r). Ideally, this should just be handled by mRTCSource (any time(0) call replaced with something to give offsetted time), or even mRTCSource being able to signal "I want the time converted to local time" (although if it's just done entirely in mRTCSource that would be shared amongst other RTC impls like MBC3/HuC3/TAMA5, none of which even do any kind of "local time conversion" (which in fairness makes sense since it's pretty meaningless outside of I guess automatic DST adjustment by the emulator).

PR #3485, opened 2025-04-29
Qt: improve AutorunScriptModel, add unit tests

ahigerd:
This PR depends on https://github.com/mgba-emu/mgba/pull/3483 as a prerequisite. (These changes will appear in this PR. You can view the last commit to review the changes in isolation.)

Qt's item models are notoriously tricky to get right. This PR adds some unit tests for `AutorunScriptModel` and makes sure all of them pass. It also cleans up the boundaries of `AutorunScriptModel` so that it can be unit tested.

Fixes of note:
* `setData` was not emitting `dataChanged`
* `removeRows` was not emitting `aboutToRemoveRows` / `rowsRemoved`
* `removeRows` did not have appropriate bounds checking, so clicking the Remove button would crash if no scripts were selected
* `moveRows` did not implement the correct API contract (which is painfully difficult to begin with)
* `moveRows` was not emitting `layoutAboutToBeChanged` / `layoutChanged`
* Qt5 requires registering the QDataStream operators, but that was missing. (Qt6 does not.)
* `#ifdef ENABLE_SCRIPTING` was missing in ConfigController and a couple places in Window.

While I was in here, I also took the opportunity to improve the serialization of `ScriptInfo`. Since endrift expressed that the struct might change in the future, I added versioning. And because I added versioning, I added forwards compatibility. (I took advantage of the fact that QString's serialization starts with a big-endian length, which will never realistically have anything but 0x00 in the first byte.) And because it now has forwards compatibility, I made the serialized data slightly less unreadable in qt.ini.

### DIFF
--- a/src/platform/qt/scripting/AutorunScriptModel.cpp
+++ b/src/platform/qt/scripting/AutorunScriptModel.cpp
@@ -5,22 +5,59 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 #include "scripting/AutorunScriptModel.h"
 
-#include "ConfigController.h"
-#include "LogController.h"
+#include <QtDebug>
+#include <iostream>
+
+QDataStream& operator<<(QDataStream& stream, const QGBA::AutorunScriptModel::ScriptInfo& object) {
+	stream << QGBA::AutorunScriptModel::ScriptInfo::VERSION;
+	stream << object.filename.toUtf8();
+	stream << object.active;
+	return stream;
+}
+
+QDataStream& operator>>(QDataStream& stream, QGBA::AutorunScriptModel::ScriptInfo& object) {
+	uint16_t version = 0;
+	stream >> version;
+	if (version == 1) {
+		QByteArray filename;
+		stream >> filename;
+		object.filename = QString::fromUtf8(filename);
+	} else {
+		qCritical() << QGBA::AutorunScriptModel::tr("Could not load autorun script settings: unknown script info format %1").arg(version);
+		stream.setStatus(QDataStream::ReadCorruptData);
+		return stream;
+	}
+	stream >> object.active;
+	return stream;
+}
 
 using namespace QGBA;
 
-AutorunScriptModel::AutorunScriptModel(ConfigController* config, QObject* parent)
+void AutorunScriptModel::registerMetaTypes() {
+	qRegisterMetaType<AutorunScriptModel::ScriptInfo>();
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
+	qRegisterMetaTypeStreamOperators<AutorunScriptModel::ScriptInfo>("QGBA::AutorunScriptModel::ScriptInfo");
+#endif
+}
+
+AutorunScriptModel::AutorunScriptModel(QObject* parent)
 	: QAbstractListModel(parent)
-	, m_config(config)
 {
-	QList<QVariant> autorun = m_config->getList("autorunSettings");
-	for (const auto& item: autorun) {
+	// Nothing to do
+}
+
+void AutorunScriptModel::deserialize(const QList<QVariant>& autorun) {
+	for (const auto& item : autorun) {
 		if (!item.canConvert<ScriptInfo>()) {
+			continue;
+		}
+		ScriptInfo info = qvariant_cast<ScriptInfo>(item);
+		if (info.filename.isEmpty()) {
 			continue;
 		}
 		m_scripts.append(qvariant_cast<ScriptInfo>(item));
 	}
+	emitScriptsChanged();
 }
 
 int AutorunScriptModel::rowCount(const QModelIndex& parent) const {
@@ -32,13 +69,14 @@ int AutorunScriptModel::rowCount(const QModelIndex& parent) const {
 
 bool AutorunScriptModel::setData(const QModelIndex& index, const QVariant& data, int role) {
 	if (!index.isValid() || index.parent().isValid() || index.row() >= m_scripts.count()) {
-		return {};
+		return false;
 	}
 
 	switch (role) {
 	case Qt::CheckStateRole:
 		m_scripts[index.row()].active = data.value<Qt::CheckState>() == Qt::Checked;
-		save();
+		emit dataChanged(index, index);
+		emitScriptsChanged();
 		return true;
 	}
 	return false;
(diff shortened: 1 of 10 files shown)


src/platform/qt/scripting/AutorunScriptModel.cpp
```diff
@@ -5,22 +5,58 @@
...
+	stream << object.filename.toUtf8();
+	stream << object.active;
+	return stream;
+}
+
+QDataStream& operator>>(QDataStream& stream, QGBA::AutorunScriptModel::ScriptInfo& object) {
+	QByteArray versionByte = stream.device()->peek(1);
+	int version = (uint8_t) versionByte[0];
+	if (version == 0) {
+		stream >> object.filename;
```
endrift, line 21: I'm not sure it's worth maintaining compat with a version that single digit users are using at the moment.

src/platform/qt/scripting/AutorunScriptModel.cpp
```diff
@@ -5,22 +5,58 @@
...
+	}
+	stream >> object.active;
+	return stream;
+}
 
 using namespace QGBA;
 
-AutorunScriptModel::AutorunScriptModel(ConfigController* config, QObject* parent)
+void AutorunScriptModel::registerMetaTypes()
+{
```
endrift, line 39:
```suggestion
void AutorunScriptModel::registerMetaTypes() {
```

src/platform/qt/scripting/AutorunScriptModel.cpp
```diff
@@ -5,22 +5,58 @@
...
+#endif
+}
+
+AutorunScriptModel::AutorunScriptModel(QObject* parent)
 	: QAbstractListModel(parent)
-	, m_config(config)
 {
-	QList<QVariant> autorun = m_config->getList("autorunSettings");
-	for (const auto& item: autorun) {
+	// initializers only
```
endrift, line 49: For consistency, I tend to say `// Nothing to do` IIRC.

src/platform/qt/scripting/AutorunScriptModel.cpp
```diff
@@ -86,34 +127,47 @@ bool AutorunScriptModel::moveRows(const QModelIndex& sourceParent, int sourceRow
...
 
 void AutorunScriptModel::addScript(const QString& filename) {
 	beginInsertRows({}, m_scripts.count(), m_scripts.count());
 	m_scripts.append(ScriptInfo { filename, true });
 	endInsertRows();
-	save();
+	emitScriptsChanged();
 }
 
-QList<QString> AutorunScriptModel::activeScripts() const {
```
endrift, line 115: wow why the heck did I use `QList<QString>` here?

src/platform/qt/scripting/AutorunScriptModel.h
```diff
@@ -17,42 +17,42 @@ Q_OBJECT
 
 public:
 	struct ScriptInfo {
+		static const uint8_t VERSION = 1;
```
endrift, line 20: Though I doubt it will ever exceed 255, I don't think it's really worth artificially limiting it to that. Perhaps uint16_t.

endrift (commented):
Looks pretty good; just a few minor things. Please make sure to rebase too.

src/platform/qt/scripting/AutorunScriptModel.h
ahigerd: Sure, that seems fair.

src/platform/qt/scripting/AutorunScriptModel.cpp
```diff
@@ -5,22 +5,59 @@
...
+
+void AutorunScriptModel::deserialize(const QList<QVariant>& autorun) {
+	for (const auto& item : autorun) {
 		if (!item.canConvert<ScriptInfo>()) {
 			continue;
 		}
+		ScriptInfo info = qvariant_cast<ScriptInfo>(item);
+		if (info.filename.isEmpty()) {
+			continue;
+		}
```
ahigerd, line 57: Added this check in here because `canConvert` doesn't detect if the variant will break during deserialization.

src/platform/qt/scripting/AutorunScriptModel.cpp
```diff
@@ -5,22 +5,59 @@
...
+
+QDataStream& operator>>(QDataStream& stream, QGBA::AutorunScriptModel::ScriptInfo& object) {
+	uint16_t version = 0;
+	stream >> version;
+	if (version == 1) {
+		QByteArray filename;
+		stream >> filename;
+		object.filename = QString::fromUtf8(filename);
+	} else {
+		qCritical() << QGBA::AutorunScriptModel::tr("Could not load autorun script settings: unknown script info format %1").arg(version);
```
ahigerd, line 26: It occurs to me to mention that this pops a message box.

src/platform/qt/scripting/AutorunScriptModel.cpp
ahigerd: I imagine because you were thinking of it parallel to `QList<ScriptInfo>`.

src/platform/qt/scripting/AutorunScriptModel.cpp
endrift: You should probably change it to qWarning

src/platform/qt/scripting/AutorunScriptModel.cpp
```diff
@@ -86,34 +128,47 @@ bool AutorunScriptModel::moveRows(const QModelIndex& sourceParent, int sourceRow
...
+	if (sourceRow == destinationChild - 1) {
 		return false;
 	}
 
-	auto item = m_scripts.takeAt(sourceRow);
-	m_scripts.insert(destinationChild, item);
-	save();
+	bool ok = beginMoveRows(QModelIndex(), sourceRow, lastSource, QModelIndex(), destinationChild);
+	if (!ok) {
+		qWarning() << tr("An error occurred while moving rows");
```
endrift, line 147: I don't think this warning is helpful. The user would visually notice, and a textual confirmation with no additional information is just clutter.

endrift (approved):
Looks good to me, apart from any further discussion on the review comments you might have.

src/platform/qt/scripting/AutorunScriptModel.cpp
ahigerd: You think so? It felt appropriate to give a popup to inform the user that the configuration that they had previously saved had gotten blown away.

endrift: That only makes sense if this only shows up once--which, unless we follow up by immediately wiping the corrupted information, won't necessarily be the case.

ahigerd: It is the case, though. When the model signals that it's been updated after loading the deserialized data, it triggers a save.

endrift: Ah. I guess that's fine then.

PR #3486, opened 2025-04-30
Qt: honor logToStdout for script logging

ahigerd:
Max wanted to see the scripting console output on stdout, but shimming the console object didn't work for capturing runtime errors. This will work for his use case, but maybe it's worth considering if we want to support this use case directly.

PR presented for discussion.

### DIFF
--- a/src/platform/qt/scripting/ScriptingController.cpp
+++ b/src/platform/qt/scripting/ScriptingController.cpp
@@ -17,6 +17,7 @@
 #include "input/GamepadButtonEvent.h"
 #include "input/GamepadHatEvent.h"
 #include "InputController.h"
+#include "LogController.h"
 #include "scripting/AutorunScriptView.h"
 #include "scripting/ScriptingTextBuffer.h"
 #include "scripting/ScriptingTextBufferModel.h"
@@ -39,6 +40,11 @@ ScriptingController::ScriptingController(ConfigController* config, QObject* pare
 		va_copy(argc, args);
 		QString message = QString::vasprintf(format, argc);
 		va_end(argc);
+		if (logger->p->m_config->getOption("logToStdout").toInt()) {
+			QTextStream out(stdout);
+			out << message << "\n";
+			out.flush();
+		}
 		switch (level) {
 		case mLOG_WARN:
 			emit logger->p->warn(message);


src/platform/qt/scripting/ScriptingController.cpp
```diff
@@ -39,6 +40,11 @@ ScriptingController::ScriptingController(ConfigController* config, QObject* pare
 		va_copy(argc, args);
 		QString message = QString::vasprintf(format, argc);
 		va_end(argc);
+		if (logger->p->m_config->getOption("logToStdout").toInt()) {
+			QTextStream out(stdout);
+			out << message << "\n";
+			out.flush();
```
ahigerd, line 46: `LogController.cpp` uses a QT_VERSION check to make sure that `endl` is defined, but this equivalent implementation works in all Qt versions. Maybe we could apply the same change there to remove a conditional compilation?

ahigerd: At the very least, IMO the `#define endl Qt::endl` should be replaced with `using Qt::endl;`.

ahigerd:
Oops, this PR depends on https://github.com/mgba-emu/mgba/pull/3485.

endrift:
This opens up a lot of architectural issues like "should logging from scripts go through the normal log controller flow", which this PR is kind of just sweeping under the rug. I don't think this is the right place to implement this and it needs more discussion.

ahigerd:
>  I don't think this is the right place to implement this and it needs more discussion.

Discussion is the reason I opened the PR. I didn't actually expect it to be merged as-is.

endrift:
So then let's ask the question: Should logging from the script go through the regular logging flow? If so, we'd need to re-plumb the console a bit, but it's probably a better approach, especially if we want to add options about the destination

ahigerd:
It does sound like it might be a good idea, especially with autorun scripts in the picture now. I like the idea of being able to see script errors in with emulation errors in the general logs, and it means that the logs would get captured even if the scripting console isn't open.

I think we might need to have some kind of log forwarding in place, so that messages always go to the scripting console but honor the logging configuration for what goes into the general logs.

Some side questions:
* Should we have separate log categories for script engine logs versus user script logs? Or is it enough to put everything into the `SCRIPT` category?
* What should the default log levels be?
* Is it meaningful to consider having log filtering in the scripting console? That would be enabled by such a re-plumbing. On the other hand, it introduces more questions about presentation and history tracking.

Also, pulling a comment from https://github.com/mgba-emu/mgba/pull/3483#discussion_r2069041172 over into this discussion:
> `mLog` itself isn't reentrant. Do we want to fix this now or later?

PR #3487, opened 2025-05-04
Qt: clean up logging output

ahigerd:
https://github.com/mgba-emu/mgba/pull/3483 had a side effect of causing quotes to be put around any QStrings that were logged. This PR adds the `noquote` manipulator to all of the calls.

I wish there was a way to make this the default behavior. The best way to do that would be something like:
```c++
QDebug& mqWarning() { return qWarning().noquote(); }
```
etc., and then change all of the call sites over to use these functions. If you'd prefer to do that, just say so; it's easy enough to do. (We can bikeshed what the functions should be named.)

I also added a few missing `tr()`'s while I was in here.

### DIFF
--- a/src/platform/qt/AudioDevice.cpp
+++ b/src/platform/qt/AudioDevice.cpp
@@ -31,7 +31,7 @@ AudioDevice::~AudioDevice() {
 
 void AudioDevice::setFormat(const QAudioFormat& format) {
 	if (!m_context || !mCoreThreadIsActive(m_context)) {
-		qInfo() << tr("Can't set format of context-less audio device");
+		qInfo().noquote() << tr("Can't set format of context-less audio device");
 		return;
 	}
 	mCoreSyncLockAudio(&m_context->impl->sync);
@@ -52,7 +52,7 @@ void AudioDevice::setInput(mCoreThread* input) {
 
 qint64 AudioDevice::readData(char* data, qint64 maxSize) {
 	if (!m_context->core) {
-		qWarning() << tr("Audio device is missing its core");
+		qWarning().noquote() << tr("Audio device is missing its core");
 		return 0;
 	}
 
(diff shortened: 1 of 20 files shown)


endrift:
The fact that this is necessary makes me think this switch was a mistake and should be reverted. Having this at every call site is tedious, and makes me question if this is indeed what we want to be using. The system was clearly not design for use with even just tr() if quote is the default and can't be changed.

Which raises the question of how LOG should be handled in the test. But in retrospect, the solution is obvious: just ifdef it separately if this is a unit test.

ahigerd:
I don't think that installing a QtMessageHandler was a mistake: I think it's valuable for us to capture warnings that originate within Qt code and keep track of them _somewhere._ They're totally getting discarded at the moment.

Qt's logging _isn't_ designed for localization. This much is true. It's meant to be developer-visible, not user-visible, which means that having a consistent message that doesn't change based on differing environments is valuable. So what _may_ have been a mistake was choosing to use it as a mechanism for user-visible messages.

But on the other hand, I _very much_ appreciate how the QtDebug library supports rendering of Qt data types. This is why I suggested wrapper functions to give us the best of both worlds.

Conditionally changing the definition of `LOG` is workable as a short-term fix for the unit test, but I think what it really indicates is a coupling issue -- ideally, `LogController` shouldn't depend on `ConfigController`. And coupling is definitely something that we haven't really paid a whole lot of attention to within the Qt frontend to date, but it's something that's going to be increasingly relevant as we add more unit tests to the frontend.

A possibility to consider: What if `LOG` didn't return `LogController::Stream` but instead `QDebug`? (See [QMessageLogger::warning](https://doc.qt.io/qt-5/qmessagelogger.html#warning-4) for example) This would allow us to define it without reference to `LogController` and allow it to accept general Qt data types, and it would still work for other categories besides just the `QT` category.

ahigerd:
Closing in favor of https://github.com/mgba-emu/mgba/pull/3489

PR #3489, opened 2025-05-09
Qt: improve log routing, clean up output

ahigerd:
This is an alternate implementation for https://github.com/mgba-emu/mgba/pull/3487. 

This PR reverts https://github.com/mgba-emu/mgba/pull/3483 except for the `tr()` fixes. The Qt message handler is still in place in order to make sure that Qt's own logging gets routed into mGBA's logging system.

To make up for not using Qt's logging system directly, I added a templated overload to `Stream::operator<<` that uses `QDebug` to format Qt data types. (Not that anything is using it _right now_ but it seems like a nice-to-have when debugging.)

I refactored `LogController` a little bit so that unit tests can just include `Log` without needing to worry about the rest of the architecture. If a `LogController` isn't instantiated (as designed, this can only happen if you don't link in `LogController.cpp.o`) then a base class implementation redirects log messages into Qt's logging system. (Yes, this does mean you could hypothetically have a recursive loop, but it's not possible to install the Qt message handler without having a `LogController` so that breaks the loop.)

While I was in here I noticed a bug: If `logToFile` or `logToStdout` are turned on, they get double-logged because each window has its own `LogController` to power the log view. I fixed that while I was in here.

### DIFF
--- a/src/platform/qt/Log.cpp
+++ b/src/platform/qt/Log.cpp
@@ -1,0 +1,78 @@
+/* Copyright (c) 2013-2025 Jeffrey Pfau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "Log.h"
+
+#include <QLoggingCategory>
+
+mLOG_DEFINE_CATEGORY(QT, "Qt", "platform.qt");
+
+using namespace QGBA;
+
+Log* Log::s_target = nullptr;
+
+Log::Stream Log::log(int level, int category) {
+	return Stream(s_target, level, category);
+}
+
+void Log::setDefaultTarget(Log* target) {
+	s_target = target;
+}
+
+Log::Log() {
+	// Nothing to do
+}
+
+Log::~Log() {
+	if (s_target == this) {
+		s_target = nullptr;
+	}
+}
+
+void Log::postLog(int level, int category, const QString& string) {
+	QLoggingCategory cat(mLogCategoryName(category));
+	switch (level) {
+	case mLOG_DEBUG:
+	case mLOG_STUB:
+		qCDebug(cat).noquote() << string;
+		return;
+	case mLOG_INFO:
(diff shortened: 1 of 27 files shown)


src/platform/qt/Log.cpp
```diff
@@ -0,0 +1,78 @@
...
+	switch (level) {
+	case mLOG_DEBUG:
+	case mLOG_STUB:
+		qCDebug(cat).noquote() << string;
+		return;
+	case mLOG_INFO:
+		qCInfo(cat).noquote() << string;
+		return;
+	case mLOG_ERROR:
+	case mLOG_GAME_ERROR:
```
endrift: `mLOG_GAME_ERROR` shouldn't be qCCritical. It won't ever be emitted by Qt logging anyway, but if anything it should be qCDebug.

endrift (approved):
Just one minor nit, then I can merge it.

src/platform/qt/Log.cpp
ahigerd: Done.

src/platform/qt/scripting/AutorunScriptModel.cpp
```diff
@@ -16,14 +15,21 @@ QDataStream& operator<<(QDataStream& stream, const QGBA::AutorunScriptModel::Scr
 }
 
 QDataStream& operator>>(QDataStream& stream, QGBA::AutorunScriptModel::ScriptInfo& object) {
+	static bool displayedError = false;
```
endrift, line 18: I don't like that this is static within the function, but I'm not sure there's any specifically better place to hold onto a value like this.

ahigerd: Yeah, that aligns with my own hesitation about it, but it's better than spamming the user with popups... The only other idea I have is to put it on GBAApp and access it through `GBAApp::app()`.

ahigerd:
@endrift 
> I'm still not sure about the rationale for all of this indirect routing is

Missed this question before. I'll just quote myself from the past:

> LOG() can't easily be used in a unit test context because LogController has a lot of transitive dependencies, some of which can't easily be constructed without standing up an entire GBAApp.

PR #3497, opened 2025-05-21
Qt: refactor popup window handling

ahigerd:
I noticed that some popups (memory logging, cheats) weren't getting deduplicated, and others (log view) weren't getting raised if they were already open when the menu command was clicked.

I also noticed some code duplication for overrides and sensors, so I was going to try to refactor them together, only to notice that the template magic I was about to write already existed.

So I went through all of the popups I could find and made sure they were recycled. In a few cases I made them modal dialogs instead of non-modal since they're clearly intended to be standalone interfaces. And in the case of MemoryAccessLogView, I had to go about deduplication a little differently because there are two ways to open it.

### DIFF
--- a/src/platform/qt/SensorView.cpp
+++ b/src/platform/qt/SensorView.cpp
@@ -16,7 +16,7 @@
 
 using namespace QGBA;
 
-SensorView::SensorView(InputController* input, QWidget* parent)
+SensorView::SensorView(std::shared_ptr<CoreController> controller, InputController* input, QWidget* parent)
 	: QDialog(parent, Qt::WindowTitleHint | Qt::WindowSystemMenuHint | Qt::WindowCloseButtonHint)
 	, m_input(input)
 	, m_rotation(input->rotationSource())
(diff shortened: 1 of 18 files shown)


src/platform/qt/MemoryView.h
```diff
@@ -40,6 +40,9 @@ public slots:
 	void update();
 	void jumpToAddress(uint32_t address);
 
+signals:
+  void openMemoryAccessLogView();
```
endrift, line 44: Wrong tab type

src/platform/qt/Window.cpp
```diff
@@ -580,6 +580,15 @@ std::function<void()> Window::openTView(A... arg) {
...
 }
 
+template <typename T, typename... A>
+std::function<void()> Window::openTViewModal(A... arg) {
+	return [=]() {
+		T* view = new T(arg...);
+		view->setAttribute(Qt::WA_DeleteOnClose);
+		view->open();
+	};
+}
```
endrift, line 590: Why doesn't this call `openView`?

src/platform/qt/SensorView.cpp
```diff
@@ -65,30 +65,34 @@ SensorView::SensorView(InputController* input, QWidget* parent)
...
-	connect(m_ui.timeOffset, &QRadioButton::clicked, [controller, this] () {
-		controller->setTimeOffset(m_ui.offsetSeconds->value());
-	});
-	connect(m_ui.offsetSeconds, qOverload<int>(&QSpinBox::valueChanged), [controller, this] (int value) {
-		if (m_ui.timeOffset->isChecked()) {
-			controller->setTimeOffset(value);
-		}
-	});
-	m_ui.timeButtons->checkedButton()->clicked();
+	if (controller) {
```
endrift, line 74: could probably just do `if (!controller) { return; }` to reduce the diff and also reduce the indent level for the remainder of the function

src/platform/qt/Window.cpp
```diff
@@ -580,6 +580,15 @@ std::function<void()> Window::openTView(A... arg) {
 	};
 }
 
+template <typename T, typename... A>
+std::function<void()> Window::openTViewModal(A... arg) {
```
endrift, line 584: Instead of having yet more similarly named mostly-overlapping functions we might want to add a keepalive flag to `openTView` if the only difference is the WA_DeleteOnClose, just like I did to the other functions. Wonder if there's some nicer way to handle this.

endrift (changes requested):
There are several distinct things going on in this PR and I'm not sure what the motivation behind all of them is. But mostly, I definitely do not like how MemoryAccessLogView::open works and would like that changed first off.

src/platform/qt/MemoryAccessLogView.cpp
```diff
@@ -15,6 +15,23 @@
...
+	if (!view) {
+		view = s_views[controllerId] = new MemoryAccessLogView(controller, parent);
+		view->setAttribute(Qt::WA_DeleteOnClose);
+		connect(controllerPtr.get(), &QObject::destroyed, view, &QWidget::close);
+	}
+	view->show();
+	view->activateWindow();
+	view->raise();
+	return view;
+}
```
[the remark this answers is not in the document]
ahigerd: Entirely fair. I'm fine with going for a different approach. (There is a leak that I hadn't considered but that would have been easily fixable, but that's neither here nor there.)

src/platform/qt/Window.cpp
ahigerd: That isn't the only difference. That one's just incidental to make sure the memory gets cleaned up. The core change is the use of `QDialog::open()` to make the popup modal, This solves the problem of having multiple windows open by blocking the user from getting to the command that opened the popup in the first place. I only chose this solution for the popups that can be treated as standalone behaviors where you're going in to do a specific thing and then you're done. I didn't do it for anything where being modal wouldn't make sense. (This is also the answer to the other question you asked in this function.)

src/platform/qt/Window.cpp
```diff
@@ -91,15 +92,34 @@
 
 using namespace QGBA;
 
+namespace QGBA {
+
+class WindowPopups {
```
ahigerd, line 97: By moving these into a private class, I was able to reduce the dependency graph in the header files, so changes to `PopupManager.h` don't cause a dozen files to rebuild.

ahigerd:
So it looks like I made a mess of this PR when I tried to separate things out the first time. I'm just going to close this PR and make some new ones, especially because I've had second thoughts about my implementation design about a few things.

PR #3517, opened 2025-06-22
GDB: Fix freeze when stepping over SWI instructions

felixjones:
GDB seems to completely freeze emulation whenever an `svc` instruction is stepped over (note: Not stepped into, just over)

After digging into this, it seems that the `SIGTRAP` sent to GDB after stepping over an `svc` is the cause of the problem, however it's not enough to just avoid `SIGTRAP` as GDB is expecting the mGBA to do _something_ after it sends a step over command

The solution I've landed with is to detect jumps into BIOS region (all SWI calls do this) and treat them as a "black box" for GDB (unless, of course, the user places an explicit breakpoint into BIOS memory)

The best way I could figure out to implement this solution is to set a temporary breakpoint after the instruction that jumped into BIOS space, and then continue execution until that breakpoint is hit

To support this plan I implemented internal support for temporary breakpoints that will remove when hit

This also provides a work around for mGBA spamming bad BIOS memory reads when the debugger attempts to step into BIOS code


endrift:
If you're going to add temporary breakpoints I'd like it to work everywhere, not just ARM hardware breakpoints. Also, the component for that commit is wrong.

felixjones:
> Also, the component for that commit is wrong.

Could you elaborate on this feedback, please? I'm not familiar with the components in use/available here

endrift:
> > Also, the component for that commit is wrong.
> 
> Could you elaborate on this feedback, please? I'm not familiar with the components in use/available here

You have commits that don't touch GDB directly, only the debugger support code. The component for that is Debugger.

felixjones:
> You have commits that don't touch GDB directly, only the debugger support code. The component for that is Debugger.

Thanks, I've re-written the commit history and fixed the commit message

endrift:
Last piece: can you handle temporary breakpoints for (ARM) software breakpoints too?

PR #3530, opened 2025-07-12
QT: drop QtLogging include

barroit:
QtLogging is only available since qtbase v6.5.0 and is redundant in LogController.cpp.


endrift:
This was superseded by #3489 getting merged.

Issue #3496, opened 2025-05-21
Qt: Enabled log levels reset when opening multiplayer window


ahigerd:
## Steps to reproduce
1. Tools -> View logs
2. Change the enabled log levels using the checkbox
3. File -> New multiplayer window

## Expected result
A new multiplayer window opens. Nothing associated with the first player should change.

## Actual result
A new multiplayer window opens. The enabled log levels reset back to the default (Warning, Error, and Fatal checked).

PR #3536, opened 2025-07-23
Flag for Window to skip reload of log levels (issue #3496)

sockeye2204:
#3496 

Component modified: Qt

To fix the above issue, a flag _m_skipResetLevels_ is added to Window that is false by default. When set to true (which is done in GBAApp::newWindow), the logging levels are not refreshed in Window::reloadConfig.

### DIFF
--- a/src/platform/qt/GBAApp.cpp
+++ b/src/platform/qt/GBAApp.cpp
@@ -131,6 +131,7 @@ Window* GBAApp::newWindow() {
 		return nullptr;
 	}
 	Window* w = new Window(&m_manager, m_configController, m_windows.count());
+	w->setSkipResetLevels(true);
 	connect(w, &Window::destroyed, [this, w]() {
 		m_windows.removeAll(w);
 		for (Window* w : m_windows) {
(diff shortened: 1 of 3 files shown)


ahigerd:
Thanks for the contribution, but adding a flag like this feels like a band-aid fix. If we can't figure out how to get reasonable behavior using the state we already have access to, then there's something deeper wrong with the design.

I have two possible ideas:

1. Only call `m_log.setLevels` if the LogView isn't currently visible. This feels reasonable because the only time this is actually a problem is when it changes while you're actively using it.
2. Create `LogView::loadConfig(ConfigController*)` with logic that will only replace the current settings with the default settings if you haven't changed anything (which means you have to track if the user has changed anything), and then call that instead of `m_log.setLevels`. This feels reasonable because it means that changing the default levels in the settings will reflect in the LogView, but only if it's already displaying the previous default settings.

I suppose the two options aren't necessarily mutually exclusive; it wouldn't be unreasonable to do both.

Might be worth popping into #development in Discord to discuss the plan.

PR #3549, opened 2025-08-08
Qt: add automated smart pointer for CoreController

ahigerd:
This is the first of a series of patches that intend to refactor the way mGBA handles popup windows.

Second PR: https://github.com/ahigerd/mgba/pull/3
Third PR: https://github.com/ahigerd/mgba/pull/4
Fourth PR: https://github.com/ahigerd/mgba/pull/5

## Description

`CorePointer` is a weak pointer to a boxed shared pointer to `CoreController` with change callbacks. That is:
* `CorePointerSource` mostly-transparently wraps around `shared_ptr<CoreController>`
* `CorePointer` acts as a `CoreController` smart pointer, but it returns the pointer contained by its associated `CorePointerSource`.
* `CoreConsumer` is an interface type exposing attach and detach callbacks.
* When `CorePointerSource` has a new `CoreController` assigned to it, each associated `CorePointer` will dispatch callbacks to the `CoreConsumer` that owns it.
* When the `CorePointerSource` is destroyed, all associated `CorePointer` objects become null pointers.

## Motivation

Primarily, I wanted to unify how all of the various popup dialogs that depend on `Window` to set them up get access to `CoreController`. This will simplify the next step of the popup refactor by allowing a single code path to handle all such view classes.

## Considered alternatives

My previous PR had used some arcane C++ template metaprogramming to avoid needing to introduce an interface class. I know you don't like multiple inheritance, but I think this is a lot more straightforward of an implementation that I expect to be easier to maintain going forward. If you insist, I can switch this back to a metaprogramming-based implementation. (However, it will beget even more awkward metaprogramming in the next phase because having a common base class makes it MUCH easier to detect what APIs a view supports.)

C++20 introduces some new template metaprogramming features that would make it significantly less ugly, but I don't expect you want to make that jump yet.

I could also use lambdas as delegates, with an API something along the lines of `m_controller.connectAttach(this, &FooView::onCoreAttached); m_controller.connectDetach(this, &FooView::onCoreDetached);`. This wouldn't change the implementation very much but it would increase the boilerplate on every class that uses it. (And I would still need the additional metaprogramming in the next phase.)

### DIFF
--- a/src/platform/qt/Window.cpp
+++ b/src/platform/qt/Window.cpp
@@ -221,7 +221,7 @@ void Window::argumentsPassed() {
 			m_gdbController = new GDBController(this);
 		}
 		if (m_controller) {
-			m_gdbController->setController(m_controller);
+			m_gdbController->setCoreSource(&m_controller);
 		}
 		m_gdbController->attach();
 		m_gdbController->listen();
@@ -580,10 +580,24 @@ std::function<void()> Window::openTView(A... arg) {
 	};
 }
 
+namespace {
+class RefAdaptor
+{
+public:
+	RefAdaptor(CorePointerSource& source) : source(&source) {}
+
+	inline operator CorePointerSource&() { return *source; }
+	inline operator CorePointerSource*() { return source; }
+	inline operator std::shared_ptr<CoreController>() { return *source; }
+
+	CorePointerSource* source;
+};
+}
+
 template <typename T, typename... A>
 std::function<void()> Window::openControllerTView(A... arg) {
 	return [=]() {
-		T* view = new T(m_controller, arg...);
+		T* view = new T(RefAdaptor(m_controller), arg...);
 		connect(m_controller.get(), &CoreController::stopping, view, &QWidget::close);
 		openView(view);
 	};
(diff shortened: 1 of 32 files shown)


ahigerd:
Next PR in the series: https://github.com/ahigerd/mgba/pull/3

src/platform/qt/Window.cpp
```diff
@@ -580,10 +580,24 @@ std::function<void()> Window::openTView(A... arg) {
...
+	RefAdaptor(CorePointerSource& source) : source(&source) {}
+
+	inline operator CorePointerSource&() { return *source; }
+	inline operator CorePointerSource*() { return source; }
+	inline operator std::shared_ptr<CoreController>() { return *source; }
+
+	CorePointerSource* source;
+};
+}
+
```
ahigerd, line 596: This is temporary and the next PR removes it. It's only here so I don't have to make all of the changes at the same time.

ahigerd:
Third PR in the series: https://github.com/ahigerd/mgba/pull/4

ahigerd:
Fourth PR in the series: https://github.com/ahigerd/mgba/pull/5

PR #3562, opened 2025-08-20
Update README.md Qt link

DanLaneyDLL:
The original link for the required Qt version has since expried, but the same executable can be found on Qt's archives. The link has been changed to use the archive link rather than the dead one.


endrift:
Can you please update the link in the translated readmes too?

DanLaneyDLL:
Done.
I will say, though, that the other translations seem to not be updated to match the english readme, with the JP readme being only 7 months out of date vs. the rest being over a year. Making that known now w/ the original "last updated" dates for each readme just so it's clear for yall.

image

Might be worth doing a once-over w/ those other languages just to make sure everything's up to date. Otherwise, thanks!

endrift:
I'm aware they're out of date. I don't have a set of translators I can poke about that though; it's wholly "whoever shows up".

PR #3568, opened 2025-08-22
Qt: Add Net Gate Functionality & Fix Chip Gate for non-JP versions of Battle Network 6

DeCoded-Void:
- Added "Net Gate" functionality for parity with shonumi/gbe-plus#87 . 
- Fixed an issue where the non-japanese versions of Mega Man Battle Network 6 are not getting the BattleChip Gate working as a result of qtitle changes, the ROM header was used to identify the game instead.
- Changed chip insertion behavior to be "hotswapped" rather than being kept inserted.
- Changes are only done in the qt end of things, battlechip.c and io related code is untouched.

This allows developers to create apps to easily interface with mGBA via TCP to insert chips in the chip gate.
Potential applications include: creating a wireless reader that reads real battlechips, using NFC tags to substitute as battle chips, phone/web app to use for chip selection, livestream chat to control inserted chips, etc.

Tested on Windows 11 64-bit with the windows:w64 build.

Screenshot 2025-08-22 091513


endrift:
Please back out the "qtitle" fix. It was a trivial one line fix when done properly, so instead of complicating the review process I just went ahead and did it myself. I can review this properly afterwards.

Rotockter:
any chance of getting the build ? for the live of me cant seam to build it myself

endrift:
I'm still waiting on them backing out the change I asked for 6 months ago before I can review the rest of it.

Rotockter:
Any tips for idiots try building mgba ? I tried to build his forked version with the net gate changes, but docker dont work because you cant get wsl to update and building with MSYS2 worked but the resulted mgba.exe with all dll's in the folder just wont start.

DeCoded-Void:
> I'm still waiting on them backing out the change I asked for 6 months ago before I can review the rest of it.

Somehow didn't get the notif for the incoming change conflict resolution from the fork until the pr got a new comment.

@Rotockter Used the w64 docker image with wsl2 backend, you may need to enable virtualization in the bios.



@endrift Adding this info in case it results in breaking changes or requires further research as a separate issue, it was work done after this pr:

Ended up getting a PIO (Programable In Out) implementation working on a physical GBA to have a rp2040 microcontroller to act as the chip gate. The GBA's clock and SO pin helped with framing as well as serve as interrupt triggers on the microcontroller's end. The weird values that seem to increment and decrement surprisingly helped with making sure I was reading the data correctly.

There would be phantom disconnects in BN6 requiring a reset in the emu. BN6 was the only game I have seen where it would respond with the "resync command" directly after confirmation (8FFF), where the game will ease its way into the indefinite communication loop. The state machine could break at this point but it'll fix itself once a resync is forced (in this case, passing the title screen and triggering a battle). This behavior doesn't seem to be present for my BN4 and BN5 copies.

The games seem to also have no problem with being probed with different Chip Gate IDs until the game sees the one it likes. For the microcontroller implementation, I'd test each flavor 5 times indefinitely until we get the 8FFF response where the MCU will remember the previous accepted flavor so it will attempt to use it as the first flavor the next time the GBA triggers the communication again. It's a workaround, but the results were consistent when it came to handling BN6 US and JP versions on real hardware, so the user doesn't have to do any sort of manual intervention and it seems instantaneous because this all happens before the title screen fully fades in.

Rotockter:
I cant update wsl it always fails and internet dont give me any solution for that. so cant use docker. cant you upload one of ya test builds of the fork ? 

kinda got interrested in this through a youtube short seeing first time the battle gate. currently i got mgba running in gamenative on android with your 2022 lua code that seam to only be for BN 4-5. Sending the 3 byte code from my pc worked on my ayn thor running Battle Network 4.5 through gamenative with the script. 

Its only realy convoluted to set everytime with: "open mgba, run BN4.5, open battle gate, restart bn4.5, open script window, find and load script, go fullscreen"

i wonder if you can autostart battle gate in the background. sadly the script net gate wont work without opening battel gate window.

goal is to make one of my micro controllers (esp32 etc) read out battle chips and sent to my any thor. struggeling with getting a connector working for the battle chips pcb contacts. buying edge connector didnt helped. now tinkering with 3d printing and paper clips.

DeCoded-Void:
The lua script is not a proper implementation because of it being made with the per frame limitation in that script. Communication with the accessory can't be done in a per frame basis with scripting alone, so the window is necessary to keep the emulated connection alive. The lua script just edits the "end result" per frame rather than simulate the actual peripheral interaction itself.

The current PR here was intended keeping things as high level as possible (in this case, mgba qt/gui) for accessibility since the lower level implementation is not feature complete, but already handles most of the time-sensitive related communication (assuming additional work in the future is going to be done to expand it for multiplayer capabilities, additional scripting abilities, etc.)

That does consequently mean that mGBA would still require its chip gate gui to remain open if you'd want to interface with it using an external/middleman program with this.

A dirty build was made based off of the fork to test its viability near the time of the pr for devs and hobbists working on chip gate related ideas to test, but it's unstable (leading to manual patching or workaround fixes, handling breaking changes to scripting API, etc.) Haven't recompiled it since then.

For chip gate related projects in the way you describe, gbe+ is recommended for the time being.

Although the "netgate" isn't really a set standard, hobbyists and collectors that do their own ways to do chip insertions from existing projects were able to get them working in the dirty build and gbe+ within a few minutes including:
- Wireless capabilities (via esp8266/32 or pico w)
- NFC stickers + readers to act as chips
- Live chat or secondary actor simulating chip insertions
- Macropad related projects
- Mobile Webapp 
- GBA multiboot homebrew program for the GBA to act as a chipgate itself --> communicate with PC via serial/uart --> network packet to the emu

You can use pogo pins or [these](https://www.digikey.com/en/products/detail/bourns-inc/70AAJ-6-M0/1232567) instead of paperclips if you want to use a microcontroller to read chips since they lower the risk of damaging chip pins.

Rotockter:
seams like mgba 0.10.5 in software render mode and your script are the only things that run in windows emulation on the android ayn thor. I would like to use gbe+ but that does run extreamly slow in windows emulation(1 frame per 20 sec). 
development/nightly version mgba 0.11 sadly does not run in windows emulation at all on android.
sadly there is no android gba emulator with battlechip gate with netgate. 

my paperclip solution as pins works so far, but the connection seam not to be fool proof and sometimes pins dont connect even then touching the contacts, like old gba cartridges into gameboy fails. 
not sure how to use your suggested pins. 
Also sounds like alot more people tried tinkering with this as i found so far. any links to other existing projects/information ?

DeCoded-Void:
Yes, for the most part, a lot of undocumented information of the battlechips is discussed and collaborated on in the MegaMan Battle Network Collectors community discord: https://discord.gg/Wa98sZza4g

There are members that have worked on or are currently working on battlechip related projects (chip reproduction, chip gate accessory research, disassembled chip gate board pics, code snippets, pcb schematics, etc.) so they can point you to someone or an existing message on there that may be able to answer some of the niche questions surrounding them.

Connectors were one of the hurdles for diy related projects, the ones used for the original accessory weren't the best quality leading to some collectors requiring to disassemble and bending connectors back on them. There is a worry about connectors scratching pins on the chips because of how scalped the original chips are.

@zarcha has a pogo pin implementation of the connector and is using an arduino nano:
image
[Link-Analyzer-Firmware](https://github.com/zarcha/Link-Analyzer-Firmware) - There is a shim programmed for it that allows it to be used with GBE+. That part currently undocumented atm, but confirmed to be working. It currently has some compatibility to the recently released Battle Chip Stadium emulation. It's spring loaded so contacts with pins are consistent.

@BlaXun has worked on an NFC implementation to substitute for chip reading

The 6 position spring contact connector linked above is what I currently use because of board assembly costs as a limitation. Using this for the implementation described in the pr with a few prototypes that can be used on that and physical hardware:
image image 
image






Interlinking some publicly available info/documentation around this that can be helpful for contributors here, since search engines are getting more unreliable lately:
- GBE+ dev's Battle Chip Gate Blog: [Shonumi's Blog](https://shonumi.github.io/articles/art16.html)
- Physical disassembly of battle chips: [Dip Chip Creation Guide](https://forums.therockmanexezone.com/dip-chip-creation-guide-t4416.html)
- Battlechip 3D model & PCB related resources:  [Megaman Exe Chip Project](https://github.com/goodtofufriday/megaman-battle--network-exe-chip-project)
- @goodtofufriday 's bridged pin spreadsheet - [Google Sheets](https://docs.google.com/spreadsheets/u/0/d/1beCqi77os6HziWFO5Ia3u95-WQ1cv9IMDcv7cEDRxCU/htmlview#)
- @zarcha 's memory address findings for the eeproms on bn6 link navi chips: [Google Doc](https://docs.google.com/document/d/1zp4swkQE4sntQGPm59474CyQrbgJAogHF1ESsc0C3sQ/)
- Chip Gate Homebrew: [forums.therockmanexezone.com](https://forums.therockmanexezone.com/) This is multiboot capable, an rp2040/rp2350 microcontroller is usable as the middleman that keeps communication alive on the gba's end and able to forward chip inserts declared by it though your choice of protocol (ex: netgate), effectively letting a physical gba act as a chipgate itself.
- Open Net Battle: [Github Repo](https://github.com/TheMaverickProgrammer/OpenNetBattle) - A community-made engine that have done some work on software based tools for planned support for the chips. A wip tool targeted for the project can be usable with the GBE+ implementation.

src/platform/qt/BattleChipView.cpp
```diff
@@ -11,6 +11,8 @@
 #include "GBAApp.h"
 #include "ShortcutController.h"
 #include "Window.h"
+#include <mgba/internal/gba/gba.h>
```
endrift, line 14: Should be a blank line between the C++ header and the C header include blocks

src/platform/qt/BattleChipView.cpp
```diff
@@ -20,6 +22,10 @@
 #include <QSettings>
 #include <QStringList>
 
+#include <QTcpServer>
```
endrift, line 25: This should be merged and sorted into the include block above.

src/platform/qt/BattleChipView.cpp
```diff
@@ -48,22 +54,23 @@ BattleChipView::BattleChipView(std::shared_ptr<CoreController> controller, Windo
...
 	m_model.setScale(size);
 
-	connect(m_ui.chipId, static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged), m_ui.inserted, [this]() {
-		m_ui.inserted->setChecked(false);
-	});
-	connect(m_ui.chipName, static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged), m_ui.chipId, [this](int id) {
-		if (id < 0) {
-			return;
+	connect(m_ui.chipName, static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged), this, [this](int idx) {
+		if (idx < 0) return;
```
endrift, line 58: Please follow the existing style guide--no single line blocks.

src/platform/qt/BattleChipView.cpp
```diff
@@ -48,22 +54,23 @@ BattleChipView::BattleChipView(std::shared_ptr<CoreController> controller, Windo
...
+		const auto keys = m_model.chipNames().keys();
+		if (idx >= 0 && idx < keys.size()) {
+			bool blocked = m_ui.chipId->blockSignals(true);
+			m_ui.chipId->setValue(keys[idx]);
+			m_ui.chipId->blockSignals(blocked);
```
endrift, line 63: This can be simplified:

```suggestion
		auto keys = m_model.chipNames().keys();
		if (idx < 0 || idx >= keys.size()) {
			return;
		}
		QSignalBlocker blocker(m_ui.chipId);
		m_ui.chipId->setValue(keys[idx]);
```

Though if I may ask, why are we blocking signals here now?

src/platform/qt/BattleChipView.cpp
```diff
@@ -48,22 +54,23 @@ BattleChipView::BattleChipView(std::shared_ptr<CoreController> controller, Windo
...
-	connect(m_ui.add, &QAbstractButton::clicked, this, &BattleChipView::addChip);
-	connect(m_ui.remove, &QAbstractButton::clicked, this, &BattleChipView::removeChip);
-	connect(m_ui.save, &QAbstractButton::clicked, this, &BattleChipView::saveDeck);
-	connect(m_ui.load, &QAbstractButton::clicked, this, &BattleChipView::loadDeck);
+	connect(m_ui.chipId,  static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged), this, &BattleChipView::onChipIdChanged);
+	connect(m_ui.insert,  &QAbstractButton::clicked, this, &BattleChipView::reinsert);
+	connect(m_ui.add,     &QAbstractButton::clicked, this, &BattleChipView::addChip);
+	connect(m_ui.remove,  &QAbstractButton::clicked, this, &BattleChipView::removeChip);
+	connect(m_ui.save,    &QAbstractButton::clicked, this, &BattleChipView::saveDeck);
+	connect(m_ui.load,    &QAbstractButton::clicked, this, &BattleChipView::loadDeck);
```
endrift, line 73: Please follow existing convention. Do not do inline alignment, and especially don't change existing code just to match a convention you added.

src/platform/qt/BattleChipView.cpp
```diff
@@ -119,9 +126,23 @@ BattleChipView::BattleChipView(std::shared_ptr<CoreController> controller, Windo
 		connect(download, &QDialog::accepted, this, &BattleChipView::updateData);
 		download->show();
 	}
+
+	if (m_ui.netGateEnable && m_ui.netGatePort && m_ui.netGateBind && m_ui.netGateStatus) {
+		connect(m_ui.netGateEnable, &QCheckBox::toggled, this, &BattleChipView::netGateToggled);
+		connect(m_ui.netGatePort,   &QLineEdit::textChanged, this, &BattleChipView::netGatePortChanged);
+		connect(m_ui.netGateBind,   &QLineEdit::textChanged, this, &BattleChipView::netGateBindChanged);
```
endrift, line 133: See above about style.

src/platform/qt/BattleChipView.cpp
```diff
@@ -119,9 +126,23 @@ BattleChipView::BattleChipView(std::shared_ptr<CoreController> controller, Windo
 		connect(download, &QDialog::accepted, this, &BattleChipView::updateData);
 		download->show();
 	}
+
+	if (m_ui.netGateEnable && m_ui.netGatePort && m_ui.netGateBind && m_ui.netGateStatus) {
```
endrift, line 130: This will never be false.

src/platform/qt/BattleChipView.cpp
```diff
@@ -133,9 +154,6 @@ void BattleChipView::setFlavor(int flavor) {
 }
 
 void BattleChipView::insertChip(bool inserted) {
-	bool blocked = m_ui.inserted->blockSignals(true);
-	m_ui.inserted->setChecked(inserted);
-	m_ui.inserted->blockSignals(blocked);
```
endrift, line 138: Why was this removed?

src/platform/qt/BattleChipView.cpp
```diff
@@ -144,12 +162,15 @@ void BattleChipView::insertChip(bool inserted) {
 }
 
 void BattleChipView::reinsert() {
-	if (m_ui.inserted->isChecked()) {
+	const bool keepInserted = m_ui.inserted->isChecked();
+	if (keepInserted) {
```
endrift, line 166: This has no functional change. Please revert this.

src/platform/qt/BattleChipView.cpp
```diff
@@ -144,12 +162,15 @@ void BattleChipView::insertChip(bool inserted) {
...
 		insertChip(false);
 		m_next = true;
 		m_frameCounter = UNINSERTED_TIME;
 	} else {
 		insertChip(true);
+		m_next = false;
+		m_frameCounter = UNINSERTED_TIME;
```
endrift, line 173: Why was this added? If it's needed then the if statement can be removed entirely as both branches are the same, just with the parameters truth values swapped based on condition.

src/platform/qt/BattleChipView.cpp
```diff
@@ -119,9 +126,23 @@ BattleChipView::BattleChipView(std::shared_ptr<CoreController> controller, Windo
...
+		connect(m_ui.netGateEnable, &QCheckBox::toggled, this, &BattleChipView::netGateToggled);
+		connect(m_ui.netGatePort,   &QLineEdit::textChanged, this, &BattleChipView::netGatePortChanged);
+		connect(m_ui.netGateBind,   &QLineEdit::textChanged, this, &BattleChipView::netGateBindChanged);
+
+		bool ok = false;
+		quint32 p = m_ui.netGatePort->text().toUInt(&ok);
+		if (ok && p > 0 && p <= 65535) {
+			m_netPort = static_cast<quint16>(p);
+		}
+		m_netBindStr = m_ui.netGateBind->text().trimmed();
```
endrift, line 140: This is the constructor. You need to set default values.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+void BattleChipView::netGatePortChanged(const QString& s) {
+	bool ok = false;
+	quint32 p = s.toUInt(&ok);
+	if (ok && p > 0 && p <= 65535) {
+		m_netPort = static_cast<quint16>(p);
+		if (m_netGateServer && m_netGateServer->isListening()) {
+			netGateStop();
+			netGateStart();
+		}
+	}
```
endrift, line 303: This can be simplified. An early return reduces the indent level and the explicit cast is unnecessary.

```suggestion
	if (!ok || p == 0 || p > 65535) {
		return;
	}
	m_netPort = p;
	if (m_netGateServer && m_netGateServer->isListening()) {
		netGateStop();
		netGateStart();
	}
```

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+	if (m_netGateServer && m_netGateServer->isListening()) {
+		netGateStop();
+		netGateStart();
+	}
+}
+
+void BattleChipView::netGateStart() {
+	if (!m_ui.netGateEnable || !m_ui.netGatePort || !m_ui.netGateBind || !m_ui.netGateStatus) {
+		return;
+	}
```
endrift, line 317: This is always false.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+		m_ui.netGateEnable->setChecked(false);
+		return;
+	}
+	netGateDisableFields(true);
+	netGateSetStatus(tr("Listening on %1:%2").arg(bindAddr.toString()).arg(m_netPort));
+}
+
+void BattleChipView::netGateStop() {
+	for (QTcpSocket* c : m_netClients) {
+		if (!c) continue;
```
endrift, line 345: No single line blocks.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+	if (m_netGateServer) {
+		m_netGateServer->close();
+	}
+
+	netGateDisableFields(false);
+	netGateSetStatus(tr("Stopped"));
+}
+
+void BattleChipView::netGateNewConnection() {
+	if (!m_netGateServer) return;
```
endrift, line 362: No single line blocks.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+	while (m_netGateServer->hasPendingConnections()) {
+		QTcpSocket* c = m_netGateServer->nextPendingConnection();
+		m_netClients.append(c);
+		m_netBufs.insert(c, QByteArray());
+		connect(c, &QTcpSocket::readyRead, this, [this, c]() { netGateReadyRead(c); });
```
endrift, line 367: No single line blocks.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+		connect(c, &QTcpSocket::disconnected, this, [this, c]() {
+			m_netBufs.remove(c);
+			m_netClients.removeAll(c);
+			c->deleteLater();
+		});
+	}
+}
+
+void BattleChipView::netGateReadyRead(QTcpSocket* sock) {
+	if (!sock) return;
```
endrift, line 377: No single line blocks.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+	QByteArray& buf = m_netBufs[sock];
+	buf += sock->readAll();
+
+	for (;;) {
+		int hdr = buf.indexOf(char(0x80));
+		if (hdr < 0) { buf.clear(); return; }
```
endrift, line 383: No single line blocks.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+		if (buf.size() - hdr < 3) {
+			if (hdr > 0) buf.remove(0, hdr);
```
endrift, line 385: No single line blocks.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+			return;
+		}
+		const unsigned char b0 = (unsigned char)buf.at(hdr + 0);
+		const unsigned char b1 = (unsigned char)buf.at(hdr + 1);
+		const unsigned char b2 = (unsigned char)buf.at(hdr + 2);
```
endrift, line 390:
```suggestion
		uint8_t b0 = buf.at(hdr + 0);
		uint8_t b1 = buf.at(hdr + 1);
		uint8_t b2 = buf.at(hdr + 2);
```

This is theoretically undefined behavior but mGBA is compiled with -fwrapv, which defines it. It's required for the core.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+		buf.remove(0, hdr + 3);
+		if (b0 != 0x80) continue;
```
endrift, line 392: No single line blocks.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+	const bool keepInserted = m_ui.inserted->isChecked();
+	if (keepInserted) {
+		insertChip(false);
+		m_next = true;
+		m_frameCounter = UNINSERTED_TIME;
+	} else {
+		insertChip(true);
+		m_next = false;
+		m_frameCounter = UNINSERTED_TIME;
+	}
```
endrift, line 414: This is reused code. Please factor it out into a method that's called in both places--after you explain why you changed it above to be the same.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+}
+
+void BattleChipView::netGateSetStatus(const QString& text) {
+	if (m_ui.netGateStatus) {
+		m_ui.netGateStatus->setText(text);
+	}
+}
```
endrift, line 421: This is just a single line inside of a block that's always true. Please just inline the callsites.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+
+void BattleChipView::netGateDisableFields(bool disable) {
+	if (!m_ui.netGatePort || !m_ui.netGateBind) return;
```
endrift, line 424: No single line blocks.

src/platform/qt/BattleChipView.cpp
```diff
@@ -261,3 +282,159 @@ void BattleChipView::updateData() {
...
+	const auto names = m_model.chipNames();
+	const QString name = names.value(id);
+	if (!name.isEmpty()) {
+		int idx = m_ui.chipName->findText(name);
+		if (idx >= 0) {
+			bool blocked = m_ui.chipName->blockSignals(true);
+			m_ui.chipName->setCurrentIndex(idx);
+			m_ui.chipName->blockSignals(blocked);
+		}
+	}
```
endrift, line 439: This can be slightly simplified:

```suggestion
	auto names = m_model.chipNames();
	QString name = names.value(id);
	if (name.isEmpty()) {
		return;
	}
	int idx = m_ui.chipName->findText(name);
	if (idx < 0) {
		return;
	}
	QSignalBlocker blocker(m_ui.chipName);
	m_ui.chipName->setCurrentIndex(idx);
```

src/platform/qt/BattleChipView.h
```diff
@@ -8,13 +8,20 @@
 #include "BattleChipModel.h"
 
 #include <QDialog>
+#include <QByteArray>
+#include <QString>
+#include <QList>
+#include <QHash>
```
endrift, line 14: Please keep this sorted

src/platform/qt/BattleChipView.ui
```diff
@@ -23,10 +23,10 @@
       <bool>true</bool>
      </property>
      <property name="dragDropMode">
-      <enum>QAbstractItemView::InternalMove</enum>
+      <enum>QAbstractItemView::DragDropMode::InternalMove</enum>
```
endrift, line 26: Please revert this change.

src/platform/qt/BattleChipView.ui
```diff
@@ -23,10 +23,10 @@
...
      </property>
      <property name="defaultDropAction">
-      <enum>Qt::MoveAction</enum>
+      <enum>Qt::DropAction::MoveAction</enum>
```
endrift, line 29: Please revert this change.

src/platform/qt/BattleChipView.ui
```diff
@@ -35,13 +35,13 @@
       </size>
      </property>
      <property name="movement">
-      <enum>QListView::Free</enum>
+      <enum>QListView::Movement::Free</enum>
```
endrift, line 38: Please revert this change.

src/platform/qt/BattleChipView.ui
```diff
@@ -35,13 +35,13 @@
...
      </property>
      <property name="isWrapping" stdset="0">
       <bool>true</bool>
      </property>
      <property name="resizeMode">
-      <enum>QListView::Adjust</enum>
+      <enum>QListView::ResizeMode::Adjust</enum>
```
endrift, line 44: Please revert this change.

src/platform/qt/BattleChipView.ui
```diff
@@ -50,7 +50,7 @@
       </size>
      </property>
      <property name="viewMode">
-      <enum>QListView::IconMode</enum>
+      <enum>QListView::ViewMode::IconMode</enum>
```
endrift, line 53: Please revert this change.

src/platform/qt/BattleChipView.ui
```diff
@@ -98,7 +98,7 @@
      <item>
       <spacer name="horizontalSpacer">
        <property name="orientation">
-        <enum>Qt::Horizontal</enum>
+        <enum>Qt::Orientation::Horizontal</enum>
```
endrift, line 101: Please revert this change.

src/platform/qt/BattleChipView.ui
```diff
@@ -129,18 +129,21 @@
...
            <string>Gate type</string>
           </property>
          </widget>
         </item>
-        <item row="1" column="1">
+        <item row="0" column="1">
          <widget class="QRadioButton" name="gateBattleChip">
+          <property name="toolTip">
+           <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-weight:700;&quot;&gt;Battle Chip Gate&lt;/span&gt;&lt;/p&gt;&lt;p&gt;Usable with &lt;span style=&quot; font-style:italic;&quot;&gt;Mega Man Battle Network 4&lt;/span&gt; and &lt;span style=&quot; font-style:italic;&quot;&gt;Rockman.EXE 4.5 Real Operation&lt;/span&gt;. &lt;/p&gt;&lt;p&gt;It can also be used to unlock minigames in &lt;span style=&quot; font-style:italic;&quot;&gt;Rockman Zero 3 &lt;/span&gt;upon inserting the &amp;quot;Meteor Red Sun&amp;quot; or &amp;quot;Blue Moon Ray&amp;quot; Battle Chip. &lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+          </property>
```
endrift, line 146: mGBA does not use tooltips at this time. Generally speaking, people don't read them, so if this information is helpful an not purely informational please try and find a better way to integrate it. Otherwise, please just remove it.

src/platform/qt/BattleChipView.ui
```diff
@@ -149,15 +152,21 @@
           </property>
          </widget>
         </item>
-        <item row="2" column="1">
+        <item row="1" column="1">
          <widget class="QRadioButton" name="gateProgress">
+          <property name="toolTip">
+           <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-weight:700;&quot;&gt;Progress Gate&lt;/span&gt;&lt;/p&gt;&lt;p&gt;Usable with &lt;span style=&quot; font-style:italic;&quot;&gt;Mega Man Battle Network 5&lt;/span&gt;.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+          </property>
```
endrift, line 159: Ditto about tooltips.

src/platform/qt/BattleChipView.ui
```diff
@@ -149,15 +152,21 @@
...
            <string notr="true">Progress &amp;Gate</string>
           </property>
          </widget>
         </item>
-        <item row="3" column="1">
+        <item row="2" column="1">
          <widget class="QRadioButton" name="gateBeastLink">
+          <property name="toolTip">
+           <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-weight:700;&quot;&gt;Beast Link Gate&lt;/span&gt;&lt;/p&gt;&lt;p&gt;Usable with &lt;span style=&quot; font-style:italic;&quot;&gt;Rockman EXE 6&lt;/span&gt;, including support for &lt;span style=&quot; font-style:italic;&quot;&gt;Mega Man Battle Network 6&lt;/span&gt;.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+          </property>
```
endrift, line 169: Ditto about tooltips.

src/platform/qt/BattleChipView.ui
```diff
@@ -210,6 +219,71 @@
...
+        </property>
+       </widget>
+      </item>
+      <item>
+       <layout class="QFormLayout" name="formLayout_6">
+        <item row="0" column="0" colspan="2">
+         <widget class="QCheckBox" name="netGateEnable">
+          <property name="toolTip">
+           <string>&lt;p&gt;&lt;b&gt;Net Gate&lt;/b&gt; - Sends Battle Chip data to mGBA over TCP.&lt;/p&gt; &lt;p&gt;It can be used with an external service/app that opens a TCP socket to submit Chip IDs.&lt;/p&gt; &lt;p style=&quot;margin:0&quot;&gt;&lt;b&gt;Protocol (3 bytes)&lt;/b&gt;:&lt;/p&gt;&lt;span style=&quot;font-family:'Courier New';&quot;&gt;&lt;ol style=&quot;margin:0; padding-left:1.2em&quot;&gt; &lt;li&gt;&lt;tt&gt;0x80&lt;/tt&gt;&lt;/li&gt; &lt;li&gt;&lt;tt&gt;CHIP_ID_HI&lt;/tt&gt;&lt;/li&gt; &lt;li&gt;&lt;tt&gt;CHIP_ID_LO&lt;/tt&gt;&lt;/li&gt; &lt;/ol&gt;&lt;/span&gt;</string>
+          </property>
```
endrift, line 235: Ditto about tooltips

src/platform/qt/BattleChipView.ui
```diff
@@ -210,6 +219,71 @@
...
+        <item row="1" column="1">
+         <widget class="QLineEdit" name="netGatePort">
+          <property name="text">
+           <string>2000</string>
+          </property>
+          <property name="maxLength">
+           <number>5</number>
+          </property>
+          <property name="placeholderText">
+           <string>1-65535</string>
```
endrift, line 257:
```suggestion
           <string>1 &ndash; 65535</string>
```

src/platform/qt/BattleChipView.ui
```diff
@@ -210,6 +219,71 @@
...
+         </widget>
+        </item>
+        <item row="2" column="1">
+         <widget class="QLineEdit" name="netGateBind">
+          <property name="text">
+           <string>127.0.0.1</string>
+          </property>
+          <property name="placeholderText">
+           <string>e.g., 127.0.0.1</string>
+          </property>
```
endrift, line 275: Unsure a placeholder is useful here when there's already a default.

src/platform/qt/BattleChipView.ui
```diff
@@ -225,7 +299,7 @@
      <item>
       <widget class="QDialogButtonBox" name="buttonBox">
        <property name="standardButtons">
-        <set>QDialogButtonBox::Close|QDialogButtonBox::Reset</set>
+        <set>QDialogButtonBox::StandardButton::Close|QDialogButtonBox::StandardButton::Reset</set>
```
endrift, line 302: Please revert this.

endrift (changes requested):
My concerns with this implementation are largely stylistic. Please adhere to existing style throughout and don't change nearby style either.

One major change I'd like is that the actual server logic should be pulled out into C code that's usable across implementations and operates directly on GBASIOBattlechipGate instances. The UI will need to synchronize with it, but it would let this be used with other frontends if we want to add that. mGBA already includes TCP server code, so you can see how to do it in gdb-stub.c. It will complicate the integration, but it's more forward-thinking in the process. I'm actually not sure it makes sense to allow both the view and a server to both be operating on the chip deck at the same time, so we might want to make it just disable the controls while it's operating.

There are also changes that were made here for reasons that aren't obvious to me, so please elaborate on why those were changed.

As for the Qt Designer changes I'm asking you to revert, I know you didn't do those manually; the Qt 6 version of Designer does them automatically. However, I'm maintaining the older Qt 5 style for now since it works with Qt 6 already and allows it to compile with older versions of Qt 5 as well.

endrift:
It also seems to have merge conflicts, so please rebase it and fix those.

DeCoded-Void:
A lot of kludges on my end.

Some of the rationale behind some of it is handling potential user misconfiguration (ex: user blanking out defaults without changing them before attempting to start the qt server) and trying to handle certain quirks in the game like where if the chipId is immediately switched in the state machine, the game could ignore that insert. 

Going to be refactoring according to the style guide and suggestions and to make sure my interpretation is correct.

I definitely do want it to work so users will still be able to use mgba's current chip gate gui, while also being able to use multiple clients at the same time (ex: being able to remotely send chips while the user also locally is able to do it themselves). A potential issue may pop up is with overfeeding data and handling how and what to prioritize on the network end of things.

endrift:
If you want to allow multiple different sources of inputs it sounds like all the more reason to have a central object for coordinating the state that multiple things can plug into and synchronize with.

FYI, I might not merge anything for 0.11 as I'm winding down, depending on how much of a change this is, but I am interested in pulling it into 0.12 once it's cleaned up

PR #3576, opened 2025-09-01
All: Support loading of ".SAV" files as well as ".sav"

LoxoSoftware:
I added some code that checks if the savefile has a ".SAV" extension instead of ".sav" so that on UNIX-like systems changing file extensions between other emulators is no longer required.
Let me know if you have any issue on Windows

### DIFF
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -255,16 +255,20 @@ bool mCoreLoadSaveFile(struct mCore* core, const char* path, bool temporary) {
 
 #if defined(ENABLE_VFS) && defined(ENABLE_DIRECTORIES)
 bool mCoreAutoloadSave(struct mCore* core) {
-	if (!core->dirs.save) {
-		return false;
-	}
-	int savePlayerId = 0;
-	char sav[16] = ".sav";
-	mCoreConfigGetIntValue(&core->config, "savePlayerId", &savePlayerId);
-	if (savePlayerId > 1) {
-		snprintf(sav, sizeof(sav), ".sa%i", savePlayerId);
-	}
-	return core->loadSave(core, mDirectorySetOpenSuffix(&core->dirs, core->dirs.save, sav, O_CREAT | O_RDWR));
+    if (!core->dirs.save) {
+        return false;
+    }
+    int savePlayerId = 0;
+    char sav[16];
+    if (mDirectorySetOpenSuffix(&core->dirs, core->dirs.save, ".SAV", O_RDONLY))
+        snprintf(sav, sizeof(sav), ".SAV");
+    else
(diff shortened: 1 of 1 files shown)


ahigerd:
I have reservations about this implementation. I'm not sure I like prioritizing the uppercase version over the lowercase version, and it feels incomplete to support a single alternative capitalization. It also doesn't handle `.SA2` for multiplayer, and it forces the UI to have to handle more cases.

I will also note that we do have long-term plans to provide save file management features in the frontend. This will make case sensitivity a non-issue because you'd be able to pick the file you want to use as your save file and mGBA will automatically handle renaming it appropriately.

(Nitpick: I think the prefix for the second commit is supposed to be `Core:`.)

LoxoSoftware:
> I have reservations about this implementation. I'm not sure I like prioritizing the uppercase version over the lowercase version, and it feels incomplete to support a single alternative capitalization. It also doesn't handle `.SA2` for multiplayer, and it forces the UI to have to handle more cases.
> 
> I will also note that we do have long-term plans to provide save file management features in the frontend. This will make case sensitivity a non-issue because you'd be able to pick the file you want to use as your save file and mGBA will automatically handle renaming it appropriately.
> 
> (Nitpick: I think the prefix for the second commit is supposed to be `Core:`.)

The ".SAV" name being a priority was a mistake by my part, it should be trivial to switch it up; also that would make the system transparent: everything works like before but now you can also load .SAV files.

I didn't support ".SAx" names because I tought it would be unnecessary since most emulators don't even have multiplayer and the lower case name is what mgba uses.

For the nitpick, I used "All" because I couldn't find "Core" in the CONTRIBUTING file

endrift:
The CONTRIBUTING.md is extremely out of date. It hasn't been updated significantly since before I added GB support in 0.4. The list of components should be seen as a rough guide, not a hard list. The git log is a much better reference.

I'm not really going to review this in depth yet, but I will say that you did _not_ adhere to the coding style _at all_, despite explicitly pointing out in the aforementioned guide that you need to do that.

LoxoSoftware:
> The CONTRIBUTING.md is extremely out of date. It hasn't been updated significantly since before I added GB support in 0.4. The list of components should be seen as a rough guide, not a hard list. The git log is a much better reference.
> 
> I'm not really going to review this in depth yet, but I will say that you did _not_ adhere to the coding style _at all_, despite explicitly pointing out in the aforementioned guide that you need to do that.

If you don't want this change merged, that's fine.
You could have said so from the beginning.

P.S.: Please update your documentation if you want to be clear with contributors, how was I supposed to know when even the README said that I should look at the CONTRIBUTING file

PR #3577, opened 2025-09-02
Update CONTRIBUTING.md to notify future contributors that "Filing pull request" section is outdated  

Mattias-Software:
Solves this issue:
https://github.com/mgba-emu/mgba/pull/3576#issuecomment-3244101428


endrift:
Alternatively, I've updated it and made it more clear that the list isn't exhaustive (though the verbiage already said that).

PR #3589, opened 2025-09-15
Debugger: CLIDebugger - fix bug CLIDebuggerRunCommand always return false

KurobaM:
Bugfix


src/debugger/cli-debugger.c
```diff
@@ -1097,7 +1097,11 @@ bool CLIDebuggerRunCommand(struct CLIDebugger* debugger, const char* line, size_
 	if (result < 0) {
 		debugger->backend->printf(debugger->backend, "Command not found\n");
 	}
-	return false;
+	if (result <=0) {
+		return false;
+	} else {
+		return true;
+	}
```
endrift:
```suggestion
	return result > 0;
```

endrift (changes requested):
Apart from the style nit, it's just simpler this way

PR #3594, opened 2025-09-21
Scripting: Fix Lua 5.1/LuaJIT and Lua 5.2 problems

nuive:
Reopening #3376 which was automatically and mistakenly closed due to some repo changes.


ahigerd (approved):

nuive:
I made some small changes to _luaPrintShim. Additions when no console was present broke Lua 5.1/JIT and 5.2 compatibility.

src/script/engines/lua.c
```diff
@@ -976,11 +995,19 @@ bool _luaLoad(struct mScriptEngineContext* ctx, const char* filename, struct VFi
 		// Make the old _ENV the __index in the metatable
 		lua_newtable(luaContext->lua);
 		lua_pushliteral(luaContext->lua, "__index");
+#if LUA_VERSION_NUM >= 502
 		lua_getupvalue(luaContext->lua, -4, 1);
+#else
+		lua_getfenv(luaContext->lua, -4);
+#endif
```
endrift, line 1002: perhaps instead we should have ifdefs:

```C
#if LUA_VERSION_NUM >= 502
#define lua_getfenv(L, I) lua_getupvalue((L), (I), 1)
#define lua_setfenv(L, I) lua_setupvalue((L), (I), 1)
#endif
```

or similar. What say you, @ahigerd?

src/script/engines/lua.c
```diff
@@ -1032,6 +1063,39 @@ const char* _luaGetError(struct mScriptEngineContext* context) {
...
+	luaL_setfuncs(lua, funcs, 0);
+#endif
+	lua_pop(lua, 1);
+}
+
+const char* _luaTolstring(lua_State* lua, int idx, size_t* len) {
+#if LUA_VERSION_NUM < 503
+	if (!luaL_getmetafield(lua, idx, "__tostring")) {
+		if (luaL_getmetafield(lua, idx, "__name")) {
+			const char *name = lua_tostring(lua, -1);
```
endrift: tiny nit:

```suggestion
			const char* name = lua_tostring(lua, -1);
```

src/script/engines/lua.c
ahigerd: The semantics are too different for me to like doing that as a general definition. I would be more okay with it if we made a new symbol that gets defined as `lua_*upvalue` for 5.2+ and `lua_*fenv` for 5.1, but I don't want it to be something that would appear to a future maintainer to have the behaviors of fenvs.

ahigerd: That said I think I prefer the way it is now. We don't use it in that many places and I like the clarity of being explicit here.

Issue #3355, opened 2024-11-25
Linkin problem (synchronization perhaps?)


nuive:
Tried using different GB pokémon games.

Linking has some strange problems using most recent dev build.

Firstly I though it was related to Lua scripting, but I think this is another thing. Linking works more or less without a Lua script running, but if a Lua script is running or has run it won't work, I've tried at least 40 times. A script as simple as a print("Hello world") will break this functionality until you restart mGBA.

This works flawlessly with mGBA 0.10.3 but I've found an interesting thing using dev builds. When you try to start a link session (for example when talking to a Pokémon Center link receptionist) 2nd mGBA window will go into fast forward mode for some strange reason. If you can stablish a connection, speed will turn to normal speed and all will work. The problem with Lua scripts is that the connection never stablish. Best case first window will go into Trade Center and 2nd window will freeze.

An interesting test would be try to do this changing windows without a keyboard. I use a keyboard because I'm blind, and perhaps the fast forward thing has some relation with alt+tab combination used to switch between windows (tab is fast forward, after all). I've tried switching windows with alt+escape too with same results but... who knows? Perhaps it is something related to keyboard input...

nuive:
Okay, more tests:

mGBA 0.10.4 has not this problem, so it seems a 0.11 issue.
I've been trying building previous dev versions. Due to FFmpeg I cannot build anything prior to:

commit 6853080b9847c939f71ffd44bad65cdba81126c3
Author: Vicki Pfau <vi@endrift.com>
Date:   Mon Nov 6 22:16:14 2023 -0800

    FFmpeg: Fix deprecation warnings


But in that commit this problem is already present.

nuive:
I've found the conflicting commit: https://github.com/mgba-emu/mgba/commit/6ecc182

In fact, if you revert this commit into the master branch link connections work using Lua scripts (obviously issue #3046 reappears and script breakpoints don't work). With this commit, even loading a print("Hello world") script file in any of the windows (even before starting a multi window) breaks linking completely.

The thing is... I have no idea why is this happening.

nuive:
More info:

This is not related to any concrete script. Opening and closing the Scripts window is enough to break linking until mGBA is closed.

nuive:
Okay, I've got new info.

It seems all is related to the debugger and not to Lua Scripts specifically.

Currently I'm using last mGBA dev build and I can do the following in order to make it crash (this happens in previous versions).

[...]

PR #3595, opened 2025-09-21
GBA SIO: Fix some race conditions

nuive:
This fixes some crashes in mGBA that occurred when opening multiple instances and having the debugger active in any of them, which caused race conditions reflected in the logs provided in issue #3355
These crashes would close all instances of the emulator.

These changes didn't break any of the games I tried.


endrift:
I am less worried about games visibly breaking and more worried about determinism (for replayability reasons, such as in netplay of course). Right now clearly it's not deterministic, but I have worries that this, in some cases, might make things a bit less deterministic. I'd have to review both the changes you're making and the surrounding code, but any non-relative change to cycle counting is going to look suspect to me, i.e.

```C
coordinator->nextHardSync = -HARD_SYNC_INTERVAL;
```

nuive:
Race conditions were present in SIO, but SIO was not the problem... closed.

Issue #3593, opened 2025-09-20
Test failures with Qt 6.9.2 on master branch


apteryks:
Hi,

When running the test suite of mgba from the master branch (commit c6aa8f5), I see:
```
1/26 Testing: platform-qt-autoscript
1/26 Test: platform-qt-autoscript
Command: "/tmp/guix-build-mgba-0.10.5-0.c6aa8f5.drv-0/build/qt/test-qt-autoscript"
Directory: /tmp/guix-build-mgba-0.10.5-0.c6aa8f5.drv-0/build/qt
"platform-qt-autoscript" start time: Sep 20 08:36 UTC
Output:
----------------------------------------------------------
********* Start testing of AutorunScriptModelTest *********
Config: Using QtTest library 6.9.2, Qt 6.9.2 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 14.3.0), unknown unknown
PASS   : AutorunScriptModelTest::initTestCase()
PASS   : AutorunScriptModelTest::testAdd()
PASS   : AutorunScriptModelTest::testEdit()
PASS   : AutorunScriptModelTest::testMoveUp()
PASS   : AutorunScriptModelTest::testMoveDown()
PASS   : AutorunScriptModelTest::testRemove()
PASS   : AutorunScriptModelTest::testSerialize()
PASS   : AutorunScriptModelTest::testDeserialize()
QWARN  : AutorunScriptModelTest::testDeserializeInvalid() Trying to construct an instance of an invalid type, type id: 1024
PASS   : AutorunScriptModelTest::testDeserializeInvalid()
QWARN  : AutorunScriptModelTest::testDeserializeV1() Trying to construct an instance of an invalid type, type id: 1024
FAIL!  : AutorunScriptModelTest::testDeserializeV1() Compared values are not the same
   Actual   (model->rowCount()): 0
   Expected (2)                : 2
   Loc: [/tmp/guix-build-mgba-0.10.5-0.c6aa8f5.drv-0/source/src/platform/qt/test/autoscript.cpp(159)]
PASS   : AutorunScriptModelTest::cleanupTestCase()
Totals: 10 passed, 1 failed, 0 skipped, 0 blacklisted, 1ms
********* Finished testing of AutorunScriptModelTest *********
Fontconfig error: No writable cache directories
Fontconfig error: No writable cache directories
<end of output>
Test time =   0.05 sec
----------------------------------------------------------
Test Failed.
[...]

PR #3599, opened 2025-09-26
Test: fix Qt6 test failures

ahigerd:
Fixes https://github.com/mgba-emu/mgba/issues/3593

### DIFF
--- a/src/platform/qt/scripting/AutorunScriptModel.cpp
+++ b/src/platform/qt/scripting/AutorunScriptModel.cpp
@@ -8,6 +8,7 @@
 #include "Log.h"
 
 QDataStream& operator<<(QDataStream& stream, const QGBA::AutorunScriptModel::ScriptInfo& object) {
+	stream.setVersion(QDataStream::Qt_5_0);
 	stream << QGBA::AutorunScriptModel::ScriptInfo::VERSION;
 	stream << object.filename.toUtf8();
 	stream << object.active;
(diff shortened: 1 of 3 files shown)


endrift:
Might want to use something newer, like Qt_5_6. The minimum version of Qt mGBA supports is somewhere around 5.12 (I forget exactly), so 5.0 is pretty ancient in comparison, but 5.12 might be slightly too new.

ahigerd:
> Might want to use something newer, like Qt_5_6. The minimum version of Qt mGBA supports is somewhere around 5.12 (I forget exactly), so 5.0 is pretty ancient in comparison, but 5.12 might be slightly too new.

The choice of `Qt_5_0` was intentional as that's the version that changed the definition of `QMetaType::UserType`. Nothing since then impacts the serialization format for the types in use here, except for Qt6 changing the enum yet again.

endrift:
What has changed in the middle? I'm more concerned about if we add things and then forget to increment the value

ahigerd:
There's a reason I put the stream version directly in the function that needs it: to make sure the impact can't spread farther than that, and it's a little tricky to look up. Qt's own code does it this way (QDateTime for example hard-codes Qt_5_6 for its serialization format). 

It looks like between 5.0 and 5.6 it started serializing all-caps mode and style strategies for QFont, some adjustment to time zones for QDateTime, added serialization for textured QBrush objects, and... that seems to be it?

PR #3602, opened 2025-09-29
ARM, SM83: Avoid single stepping if CPU is halted or blocked

nuive:
Single step functions would run an instruction even if CPU was halted or blocked.  This disrupted the order of events as well as synchronization between the different components. SIO even caused emulator crashes randomly.


endrift:
Unfortunately, making single stepping *always* do something was an intentional semantic design choice. It is assumed that something will happen if you call step. I assume there are some callsites that still try to call step when the core is halted due to external concerns. I would try to root out where and why, and how to prevent that.

PR #3603, opened 2025-09-29
Debugger: Make debugger aware of thread status when using threads

nuive:
When mGBA used Threading, the debugger ignored this and would run even when the thread status was not RUNNING. This issue has been addressed by allowing the debugger to know thread's current status at every momment.
Mutex handling has also been added (when using threads) to prevent unwanted or unpredictable changes during critical sections.

Between this and #3602 issue #3355 should be fixed.

### DIFF
--- a/include/mgba/debugger/debugger.h
+++ b/include/mgba/debugger/debugger.h
@@ -212,10 +212,13 @@ struct mDebuggerPlatform {
 	void (*nextInstructionInfo)(struct mDebuggerPlatform* d, struct mDebuggerInstructionInfo* info);
 };
 
+struct mCoreThreadInternal;
+
 struct mDebugger {
 	struct mCPUComponent d;
 	struct mDebuggerPlatform* platform;
 	enum mDebuggerState state;
+	struct mCoreThreadInternal* threadImpl;
 	struct mCore* core;
 	struct mScriptBridge* bridge;
 	struct mStackTrace stackTrace;
@@ -245,9 +248,13 @@ void mDebuggerInit(struct mDebugger*);
 void mDebuggerDeinit(struct mDebugger*);
 
 void mDebuggerAttach(struct mDebugger*, struct mCore*);
+#ifndef DISABLE_THREADING
+void mDebuggerSetThread(struct mDebugger*, struct mCoreThreadInternal*);
+void mDebuggerUnsetThread(struct mDebugger*);
+#endif
 void mDebuggerAttachModule(struct mDebugger*, struct mDebuggerModule*);
 void mDebuggerDetachModule(struct mDebugger*, struct mDebuggerModule*);
-void mDebuggerRunTimeout(struct mDebugger* debugger, int32_t timeoutMs);
+void mDebuggerRunTimeout(struct mDebugger*, int32_t);
 void mDebuggerRun(struct mDebugger*);
 void mDebuggerRunFrame(struct mDebugger*);
 void mDebuggerEnter(struct mDebugger*, enum mDebuggerEntryReason, struct mDebuggerEntryInfo*);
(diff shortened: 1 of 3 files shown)


src/debugger/debugger.c
```diff
@@ -97,6 +101,20 @@ void mDebuggerAttach(struct mDebugger* debugger, struct mCore* core) {
...
+#ifndef DISABLE_THREADING
+void mDebuggerSetThread(struct mDebugger* debugger, struct mCoreThreadInternal* thread) {
+	if (!debugger->threadImpl) {
+		debugger->threadImpl = thread;
+	}
+}
+
+void mDebuggerUnsetThread(struct mDebugger* debugger) {
+	if (debugger->threadImpl) {
+		debugger->threadImpl = 0;
```
endrift, line 113:
```suggestion
		debugger->threadImpl = NULL;
```

include/mgba/debugger/debugger.h
```diff
@@ -245,9 +248,13 @@ void mDebuggerInit(struct mDebugger*);
...
 
 void mDebuggerAttach(struct mDebugger*, struct mCore*);
+#ifndef DISABLE_THREADING
+void mDebuggerSetThread(struct mDebugger*, struct mCoreThreadInternal*);
+void mDebuggerUnsetThread(struct mDebugger*);
+#endif
 void mDebuggerAttachModule(struct mDebugger*, struct mDebuggerModule*);
 void mDebuggerDetachModule(struct mDebugger*, struct mDebuggerModule*);
-void mDebuggerRunTimeout(struct mDebugger* debugger, int32_t timeoutMs);
+void mDebuggerRunTimeout(struct mDebugger*, int32_t);
```
endrift, line 257: Please leave the name `timeoutMs` in the signature. I remove parameter names that are obvious, but `timeoutMs` includes the unit, which just `int32_t` doesn't.

src/debugger/debugger.c
```diff
@@ -127,45 +145,105 @@ void mDebuggerRunTimeout(struct mDebugger* debugger, int32_t timeoutMs) {
...
 	switch (debugger->state) {
 	case DEBUGGER_RUNNING:
-		if (!debugger->platform->hasBreakpoints(debugger->platform)) {
-			debugger->core->runLoop(debugger->core);
-		} else {
-			debugger->core->step(debugger->core);
-			debugger->platform->checkBreakpoints(debugger->platform);
+		bool hasBreakpoints = debugger->platform->hasBreakpoints(debugger->platform);
+#ifndef DISABLE_THREADING
+		if (impl) {
```
endrift, line 156: Almost all of these checks can be put outside of the switch statement by just doing an unlock before and lock afterwards. The only exceptions are the early exit if the thread is paused, which doesn't really seem like it belongs in this function anyway (are there any other callsites that can't do the check itself?), and the two cases at the end, which you can do an early check for before unlocking. By doing the checks external to the switch statement, you can reduce duplication of code and massively reduce the diff size at the same time.

endrift (changes requested):
I'm really not sure what this fixes. It looks like it's just leaking the thread abstraction into the debugger, but only for one check on one callsite that could be done outside the function. Am I missing something?

nuive:
> I'm really not sure what this fixes. It looks like it's just leaking the thread abstraction into the debugger, but only for one check on one callsite that could be done outside the function. Am I missing something?

I tried it, but if I do that inside _mCoreThreadRun all commands entered in the debugger window never reach debugger. It never pauses, it never shows help...

src/debugger/debugger.c
nuive: I thought 0 was the prefered C sintax and NULL for C++?

endrift:
I believe this PR is obviated by 141a82e5aad69cf5ec46f5c442f0864ed0cca808. Let me know.

Issue #3580, opened 2025-09-06
Pokemon Gold: Sprite Error


MarkuzZz5:
I'm shiny hunting on Pokemon Gold. 
When I have multiple game windows open, I experience serious sprite errors in some encounters.
This only occurs when I'm playing multiple games at once.
It occurs at normal speed and in fast forward.
If I only have one game open, everything is normal.

I play on Windows, 64-bit, with a Bluetooth controller.
Could it be due to my weak PC?

endrift:
Can you post screenshots of this error? Which version of mGBA are you using? This sounds like multiple games are using the same save file, which can lead to corruption.

MarkuzZz5:
<img width="907" height="917" alt="Image" src="https://github.com/user-attachments/assets/266656ba-94d9-4748-aff5-16286585862e" />

<img width="1081" height="907" alt="Image" src="https://github.com/user-attachments/assets/ce6716ea-9e89-4d28-ab96-a9ddb4f51907" />

It's possible that they're using the same save file, because sprites spawn relatively simultaneously, and sometimes they get slightly mixed up.
Strangely, this only happens since the latest version, before that everything worked
How can I solve this problem?

fpi1337:
It doesn't happen for me. But then again, I'm using the AppImage on Debian 13.

<img width="1916" height="1054" alt="Image" src="https://github.com/user-attachments/assets/d6c5dcc4-bf9c-45dc-bb6b-38e0c0d6c3d6" />

endrift:
It still sounds like they're sharing save files, a thing that's supposed to be disallowed. Can you create a bug report zip (File > Report Bug) with all four games open and upload it here?

PR #3611, opened 2025-10-20
Apply an exclusive lock in VFileOpenFD if opening the file as writable

CasualPokePlayer:
Untested, should prevent situations such as #3580 from occurring.


endrift:
Two suggestions:

1. Let's put this in the map function if it's MAP_WRITE
2. Check the advisory lock on Unix while when opening

CasualPokePlayer:
1. I assume this would end up meaning we should unlock it at unmap. Which then leads to the question, how do we know if the unmap is with a read-only mapping or a writable mapping? I would guess each mapping would need to be tracked, for both Windows and POSIX platforms (Granted, the API seems inconsistent in whether or not it expects multiple mappings to occur for the same VFile, the Windows code implies that is possibly expected given its HandleMappingList vector, although the POSIX path with _POSIX_MAPPED_FILES undefined ends up having a `writable` member that's set on each _vfdMap call?)
2. I assume this just means replacing F_SETLK with F_GETLK and such and responding accordingly?

Issue #3547, opened 2025-08-05
Checking "Mute" when inactive also mutes when fast-forwarding


zippy37:
I've been trying to figure out why my audio was muting while fast-forwarding even though I had my fast-forward mute option unchecked. 

<img width="451" height="74" alt="Image" src="https://github.com/user-attachments/assets/4b26351c-bd2c-421e-86e7-065fc3856369" />

That's when I remembered I'd turned on the settings to pause and mute the game while the game window was inactive and minimized. I figured there wasn't a reason not to do both, and so I did. That's when I started to notice the audio issues. I played around with all of the checkboxes and found that the culprit was the mute while inactive checkbox. So when I unchecked it and kept all the other ones checked, it no longer muted while fast-forwarding.

<img width="381" height="57" alt="Image" src="https://github.com/user-attachments/assets/b096ccb2-f64c-4d00-9793-2d0d54359ae6" />

Fortunately for me, I don't lose functionality from this since clicking on the taskbar, alt-tabbing apparently counts as minimized, so the game pauses and mutes appropriately anyway. But still figure it's worth mentioning.

PR #3644, opened 2025-12-07
Fix unexpected mute when window becomes focused while fast-forwarding

ahigerd:
Reported by user "Vapoure" in Discord.

## Steps to reproduce

1. In Settings -> Audio/Video -> Fast Forward, check Mute
2. Uncheck Mute
3. Ensure that none of the "Pause" or "Mute" options are checked in the Interface tab
4. Start a game that will produce audio
5. Toggle fast-forward (shift+tab or menu option)
6. Switch to another window
7. Switch back to the mGBA window

## Expected results

There should be no audio interruptions, because none of the Pause or Mute options are turned on.

## Actual results

The audio becomes muted when the window regains focus.

## Additional details

It turns out that there's a difference between the default behavior of the Fast Forward Mute option and the "off" behavior -- this issue cannot be reproduced if the `fastForwardMute` line in `config.ini` does not exist. I'm not sure of the historical background for why this option's unconfigured behavior would be expected to differ from either the checked or unchecked behavior, but the logic for working with the -1 value is clearly mistaken.

I also separated the "inactive" and "minimized" boolean flags because there's a potential edge case that the window could receive a `showEvent` without gaining the focus, which would result in unexpected behavior. (This is possible with some window managers.)

### DIFF
--- a/src/platform/qt/Window.cpp
+++ b/src/platform/qt/Window.cpp
@@ -964,6 +964,9 @@ void Window::gameStopped() {
 	for (auto& action : m_platformActions) {
 		action->setEnabled(true);
 	}
+	for (auto& action : m_nonMpActions) {
+		action->setEnabled(true);
+	}
 	for (auto& action : m_gameActions) {
 		action->setEnabled(false);
 	}
--- a/CHANGES
+++ b/CHANGES
@@ -50,6 +50,7 @@ Other fixes:
  - Qt: Fix potential crash when configuring shortcuts
  - Qt: Fix regression where loading BIOS creates a save file (fixes mgba.io/i/3359)
  - Qt: Fix selecting high tiles in tile and map views (fixes mgba.io/i/3461)
+ - Qt: Fix some actions disabled in multiplayer staying disabled after closing game
  - Qt: Fix unexpected mute when window becomes focused while fast-forwarding
 Misc:
  - 3DS: Change title ID to avoid conflict with commercial title (fixes mgba.io/i/3023)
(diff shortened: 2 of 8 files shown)


ahigerd:
This appears to be related to https://github.com/mgba-emu/mgba/issues/3547 and might fix that problem as well.

src/platform/qt/Window.cpp
```diff
@@ -781,7 +781,7 @@ void Window::hideEvent(QHideEvent* event) {
 		m_controller->setPaused(true);
 	}
 	if (m_config->getOption("muteOnMinimize").toInt()) {
-		m_inactiveMute = true;
+		m_minimizedMute = false;
```
endrift:
```suggestion
		m_minimizedMute = true;
```

I think this is what you meant? Otherwise this never gets set to true.

ahigerd: Whoops. Yeah, think you're right.

src/platform/qt/CoreController.cpp
```diff
@@ -303,7 +303,7 @@ void CoreController::loadConfig(ConfigController* config) {
 	m_autoload = config->getOption("autoload", true).toInt();
 	m_autofireThreshold = config->getOption("autofireThreshold", m_autofireThreshold).toInt();
 	m_fastForwardVolume = config->getOption("fastForwardVolume", -1).toInt();
-	m_fastForwardMute = config->getOption("fastForwardMute", -1).toInt();
+	m_fastForwardMute = config->getOption("fastForwardMute", false).toBool();
```
ahigerd: Uh. Wha...? This is suddenly failing to compile when it worked before... How on _earth_ did I break this?

ahigerd: Or rather, how on earth did it ever work before?

Issue #3614, opened 2025-10-26
mGBA cannot read "NUMPAD+" from "qt.ini" upon startup


96F:
Tested it on mGBA 0.10.5

If you map "NUMPAD+" to any of the Shortcuts, like Fast Forward, it will be saved as "+" character and will be usable during your game session.
However, if you restart the emulator, the "NUMPAD+" from said shortcut will be gone and you have to remap it, despite it being saved in the "qt.ini" file. It seems like the emulator is having trouble reading the "+" character as "NUMPAD+" for some reason. All other numpad keys seem to work fine.

RetroEdit:
Confirmed in latest dev build b3f1ea318463f18b03b967eda409b7de6fe745a1 . Thanks for the detailed report!

endrift:
Should be fixed now.

PR #3686, opened 2026-01-20
Qt: Fix persistence of NUMPAD+ shortcut

andybaish:
NUMPAD+ shortcuts were saved as '+' in qt.ini but would not persist across restarts. Previously, NUMPAD+ was interpreted as a modifier-only shortcut. This change ensures only true modifier shortcuts follow the modifier path, allowing NUMPAD+ to be restored correctly.

Fixes issue #3614.

### DIFF
--- a/src/platform/qt/ShortcutController.cpp
+++ b/src/platform/qt/ShortcutController.cpp
@@ -254,7 +254,7 @@ bool ShortcutController::loadShortcuts(std::shared_ptr<Shortcut> item) {
 		QString s = shortcut.toString();
 		if (s.endsWith('+') &&
 			(s.contains("Ctrl") || s.contains("Shift") ||
-			s.contains("Alt")  || s.contains("Meta"))) {
+			s.contains("Alt") || s.contains("Meta"))) {
 			updateKey(item, toModifierShortcut(s));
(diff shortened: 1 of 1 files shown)


src/platform/qt/ShortcutController.cpp
```diff
@@ -251,10 +251,13 @@ bool ShortcutController::loadShortcuts(std::shared_ptr<Shortcut> item) {
 	loadGamepadShortcuts(item);
 	QVariant shortcut = m_config->getQtOption(item->name(), KEY_SECTION);
 	if (!shortcut.isNull()) {
-		if (shortcut.toString().endsWith("+")) {
-			updateKey(item, toModifierShortcut(shortcut.toString()));
+		QString s = shortcut.toString();
+		if (s.endsWith('+') &&
+			(s.contains("Ctrl") || s.contains("Shift") ||
+			s.contains("Alt")  || s.contains("Meta"))) {
```
endrift: typo
```suggestion
			s.contains("Alt") || s.contains("Meta"))) {
```

endrift (approved):

endrift:
Thanks!

PR #3692, opened 2026-01-29
Scripting: Commands to read / write palettes

Jengolus:
This PR adds scripting commands to manipulate color palettes and corresponding tests. They can be used as follows:

```
local palette = emu:readPalette(0)
print("Colors of palette 0:")
print("Index", "Red", "Green", "Blue")
for idx, color in ipairs(palette) do
    local values = util.unpackColor(color)
    print(idx, values.red, values.green, values.blue)
    -- Or equivalently:
    --print(idx, values[1], values[2], values[3])
end

local red = util.packColor(255, 0, 0)
-- Does nothing on GBA
emu:writePalette(0, palette[1], red, palette[3], palette[4])
```

Motivation: There isn't currently any way to directly modify palettes, especially for SGB games. In particular, I am working on a socket interface between the pixel art software Aseprite and mGBA to preview images directly in the games they should appear in, but I cannot change the colors on SGB.

Design choices:
- packColor takes red, green, and blue from 0 to 255, not 0 to 31
- unpackColor converts the channels as 255 * channel // 31, not 8 * channel
- writePalette accepts the colors as four integers and thus doesn't work for GBA. I tried to make it work using a table, but couldn't access the table's values from C. I can change that if given instructions.

### DIFF
--- a/src/core/scripting.c
+++ b/src/core/scripting.c
@@ -696,9 +696,9 @@ mSCRIPT_DEFINE_STRUCT(mCore)
 	mSCRIPT_DEFINE_DOCSTRING("Write the value of the register with the given name")
 	mSCRIPT_DEFINE_STRUCT_METHOD(mCore, writeRegister)
 
-	mSCRIPT_DEFINE_DOCSTRING("Read a 16-bit value from the given palette index (0-indexed)")
+	mSCRIPT_DEFINE_DOCSTRING("Read a 16-bit value encoding the RGB channels in 5 bits each from the given palette index (0-indexed). See util.unpackColor")
 	mSCRIPT_DEFINE_STRUCT_METHOD(mCore, readPalette)
-	mSCRIPT_DEFINE_DOCSTRING("Write a 16-bit value to the given palette index (0-indexed)")
+	mSCRIPT_DEFINE_DOCSTRING("Write a 16-bit value encoding the RGB channels in 5 bits each to the given palette index (0-indexed). See util.packColor")
 	mSCRIPT_DEFINE_STRUCT_METHOD(mCore, writePalette)
 
 	mSCRIPT_DEFINE_DOCSTRING("Save state and return as a buffer. See C.SAVESTATE for possible values for `flags`")


endrift:
First and foremost: There are several peculiar consistency issues with your code that imply you didn't review it yourself before submitting it. Is this code AI generated? Please review your own code for these issues before asking us to review it. Please also ensure you adhere to the existing coding and style guides used elsewhere in the code.

Second, when it comes to opinionated matters:

1. Do not scale the colors when reading/writing. Everyone has different opinions for how this should be handled, and it's a lossy operation, so we should not enforce it in the scripting API. The script itself can do whatever scaling it wants.
2. Writing palettes should only write a single entry in the palette, not the entire palette at once. It would solve your table issue (which itself is strange considering you construct a table in C already).

Please bring this PR to non-draft quality before asking for further review.

src/core/scripting.c
```diff
@@ -415,88 +415,68 @@ static void _mScriptCoreWriteRegister(struct mCore* core, const char* regName, i
...
-		return &mScriptValueNull;
-	}
-
-	if (index >= paletteCount) {
-		return &mScriptValueNull;
-	}
-
-	if (palette == 0) {
-		return &mScriptValueNull;
+		return 0x8000;
```
endrift, line 438: What is the logic behind returning 0x8000 here?

src/core/scripting.c
```diff
@@ -415,88 +415,68 @@ static void _mScriptCoreWriteRegister(struct mCore* core, const char* regName, i
...
-		struct mScriptValue* key = mScriptValueCreateFromUInt(i + 1);
-		struct mScriptValue* val = mScriptValueCreateFromUInt(palette[4*index + i]);
-
-		mScriptTableInsert(tbl, key, val);
+	if (index >= count) {
+		return 0x8000;
 	}
 
-	return tbl;
+	return palette[index];
```
endrift: This needs to be a LOAD_16.

src/core/scripting.c
```diff
@@ -415,88 +415,68 @@ static void _mScriptCoreWriteRegister(struct mCore* core, const char* regName, i
...
-		return;
-	}
-
-	uint32_t values[4] = {c0, c1, c2, c3};
-
-	for (int i = 0; i < 4; i++) {
-		palette[4*index + i] = values[i];
-	  video->renderer->writePalette(video->renderer, 4*index + i, values[i]);
-	}
+	palette[index] = color;
```
endrift: This needs to be a STORE_16, but it also needs to be before the ->writePalette calls.

src/core/scripting.c
```diff
@@ -711,9 +691,9 @@ mSCRIPT_DEFINE_STRUCT(mCore)
 	mSCRIPT_DEFINE_DOCSTRING("Write the value of the register with the given name")
 	mSCRIPT_DEFINE_STRUCT_METHOD(mCore, writeRegister)
 
-	mSCRIPT_DEFINE_DOCSTRING("Get the colors of the palette of the given index (0-indexed). Returns a table of integers which encode the colors in 5 bits each. See util.unpackColor.")
+	mSCRIPT_DEFINE_DOCSTRING("Read a 16-bit value from the given palette index (0-indexed)")
 	mSCRIPT_DEFINE_STRUCT_METHOD(mCore, readPalette)
-	mSCRIPT_DEFINE_DOCSTRING("Set the colors of the palette of the given index (0-indexed). Takes four integers encoding the colors in 5 bits each. See util.packColor. Does nothing on GBA.")
+	mSCRIPT_DEFINE_DOCSTRING("Write a 16-bit value to the given palette index (0-indexed)")
```
endrift: Why did you remove mention of format and util.packColor?

src/core/scripting.c
Jengolus: To my knowledge and testing, the highest bit is ignored, so I figured it would make a good error marker since `readPalette` cannot return a `nil` anymore.

src/core/scripting.c
Jengolus: Done for both `readPalette` and `writePalette.`

src/core/scripting.c
Jengolus: I felt that mimicked the other docstrings more closely; added it back in now.

PR #3701, opened 2026-02-09
Changing StartupWMClass

Kyuyrii:
Yo! I'm the publisher of the Snap version, specifically the [Kyu version].

This change is just to avoid this problem here:

Captura de imagem_20260209_184406

### DIFF
--- a/res/mgba-qt.desktop
+++ b/res/mgba-qt.desktop
@@ -10,4 +10,4 @@ Comment=Nintendo Game Boy Advance Emulator
 Categories=Game;Emulator;
 MimeType=application/x-gameboy-advance-rom;application/x-agb-rom;application/x-gba-rom;application/x-gameboy-rom;application/x-gameboy-color-rom;
 Keywords=emulator;Nintendo;advance;gba;gbc;gb;Game Boy Advance;Game Boy Color; Game Boy;
-StartupWMClass=mGBA
+StartupWMClass=io.mgba.mGBA


endrift:
@hebo6 you originally added that line. Does this change look correct to you?

Kyuyrii:
Just to add some information I discovered, WMClass in Xorg is different from WMClass in Wayland.

This is probably why certain problems like this occur, since many .desktop files may have been created before the migration of the various desktop environments to Wayland.

endrift:
Yes, I suspected that might be the case. It might be worth trying to set the application name in the QCoreApplication to projectName, which will match the desktop file. See if that works instead.

endrift:
Poke. I added the change I mentioned in my last post. Can you see if it works without this PR?

Kyuyrii:
I'm currently using Gnome, so I can't test it right now, but I'll test it as soon as possible.

Kyuyrii:
I built mGBA today, but the problem persists; the mGBA window is recognized as io.mgba.mGBA.

StartupWMClass=io.mgba.mGBA is needed to resolve the issue.

ReillyBrogan:
This is completely incorrect. Instead `StartupWMClass` should be removed and the desktop file renamed to `io.mgba.mGBA.desktop`

ReillyBrogan:
Ah, that is already done in main. In that case `StartupWMClass` just needs to be removed completely.

electricbrass:
When using appimages, the desktop file filename can't necessarily be relied on, as tools like Gear Lever tend to rename them. And so leaving out StartupWMClass in desktop files for appimages frequently causes issues on GNOME in particular in my experience.

PR #3726, opened 2026-03-08
Add scripting history

endrift:
My C++ is a bit rusty and @ahigerd has more Qt experience for me so I want to run this by him before I merge it.

### DIFF
--- a/src/platform/qt/HistoryLineEdit.cpp
+++ b/src/platform/qt/HistoryLineEdit.cpp
@@ -75,3 +75,9 @@ void HistoryLineEdit::keyPressEvent(QKeyEvent* keyEvent) {
 	}
 	setIndex(newIndex);
 }
+
+void HistoryLineEdit::appendLine(const QString& string) {
+	int rowCount = model()->rowCount();
+	model()->insertRows(rowCount, 1);
+	model()->setData(model()->index(rowCount, 0), string, Qt::DisplayRole);
+}
--- a/src/platform/qt/HistoryLineEdit.h
+++ b/src/platform/qt/HistoryLineEdit.h
@@ -31,6 +31,7 @@ signals:
 
 public slots:
 	void setIndex(int);
+	void appendLine(const QString& string);
 
 protected:
 	void keyPressEvent(QKeyEvent*) override;
--- a/src/platform/qt/scripting/ScriptingView.h
+++ b/src/platform/qt/scripting/ScriptingView.h
@@ -5,7 +5,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 #pragma once
 
+#include <QStringList>
+
 #include "ui_ScriptingView.h"
+
+class QStringListModel;
 
 namespace QGBA {
 
@@ -20,7 +24,7 @@ public:
 	ScriptingView(ScriptingController* controller, ConfigController* config, QWidget* parent = nullptr);
 
 private slots:
-	void submitRepl();
+	void submitRepl(const QString& line);
 	void load();
 	void loadMostRecent();
 
@@ -38,6 +42,7 @@ private:
 	ScriptingController* m_controller;
 	QStringList m_mruFiles;
 	QTextDocument* m_blankDocument;
+	QStringListModel* m_history;
 };
 
 }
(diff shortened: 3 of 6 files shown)


src/platform/qt/DebuggerConsoleController.h
```diff
@@ -8,7 +8,7 @@
 #include "DebuggerController.h"
 
 #include <QMutex>
-#include <QStringList>
+#include <QStringListModel>
```
adamhigerd, line 11: Nit: Use a forward declaration and #include in the cpp

src/platform/qt/scripting/ScriptingView.h
```diff
@@ -5,6 +5,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 #pragma once
 
+#include <QStringListModel>
```
adamhigerd: Same nit: Use a forward declaration and #include in the cpp

src/platform/qt/HistoryLineEdit.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "HistoryLineEdit.h"
+
+#include <QKeyEvent>
+
+using namespace QGBA;
+
+HistoryLineEdit::HistoryLineEdit(QWidget* parent)
+	: QLineEdit(parent)
+{
+	installEventFilter(this);
```
adamhigerd: Nominally fine, but is there any particular reason why you went for an event filter instead of `keyPressEvent`?

src/platform/qt/HistoryLineEdit.cpp
```diff
@@ -0,0 +1,81 @@
...
+		newIndex = m_model->rowCount();
+		break;
+	default:
+		return false;
+	}
+	setIndex(newIndex);
+	return true;
+}
+
+void HistoryLineEdit::appendLine(const QString& string) {
```
adamhigerd, line 79: Instead of depending on the parent to call `appendLine` you might consider connecting to the `returnPressed` signal internally to simplify the external contract.

src/platform/qt/HistoryLineEdit.cpp
```diff
@@ -0,0 +1,81 @@
...
+	int rowCount = model()->rowCount();
+	model()->insertRows(rowCount, 1);
```
adamhigerd, line 81: Thoughts for future enhancement: I wonder if you should put a cap on the history size and/or deduplicate consecutive history entries.

src/platform/qt/HistoryLineEdit.h
```diff
@@ -0,0 +1,40 @@
+/* Copyright (c) 2013-2026 Jeffrey Pfau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#pragma once
+
+#include <QAbstractItemModel>
```
adamhigerd: Use a forward declaration

src/platform/qt/HistoryLineEdit.cpp
adamhigerd: On that note, it might even be a good idea to add a `commandEntered(QString)` signal that you can emit in the slot after you check that it's non-empty; would knock 4 lines of code out of each use.

src/platform/qt/HistoryLineEdit.cpp
endrift: Was refactoring code that used this approach because it was on a different object. I should move it to keyPressEvent though, you're right.

src/platform/qt/HistoryLineEdit.cpp
endrift: I had a similar thought, but indeed wanted to leave it for future work.

src/platform/qt/HistoryLineEdit.cpp
endrift: I've simplified the usage some, but since the DebugConsoleController has to handle history internally due to how the CLI is designed I can't do that connection. Regardless, everything else should be resolved now.

PR #3752, opened 2026-04-01
Scripting: added more APIs

Mte90:
## New APIs Added
1. Memory Access Tracing
`emu:traceMemory(enabled, callback)`
Enables real-time memory access monitoring. The callback receives:
- address: Memory address accessed
- value: Value read/written  
- isWrite: Boolean indicating write operation
2. Instruction Tracing
`emu:traceInstructions(enabled, callback)`
Monitors instruction execution. The callback receives:
- pc: Program counter
- opcode: Executed opcode
- isThumb: Boolean indicating Thumb vs ARM mode
3. Register Watch
```
emu:watchRegister("r0", callback)
emu:watchRegister("pc", callback)
```
Watches CPU registers for changes. Supports r0-r12, sp, lr, pc. Callback receives old and new values.
4. Interrupt Tracing
`emu:traceInterrupts(enabled, callback)`
Tracks interrupt firing.
5. Bus Transaction Log
```
log = emu:getBusTransactionLog()
emu:clearBusTransactionLog()
```
Provides a log of all bus transactions with cycle, type, address, and value.
## Use Cases
These APIs enable:
- Dynamic analysis of ROM behavior without static analysis
- Debugging tools to trace code execution and memory access patterns  
- Emulator features like cheat detection, memory breakpoints, and performance profiling
- Test automation for validating emulator correctness against known test cases
- Coverage analysis to identify untested code paths
## Implementation Details
- Added callback fields to mScriptCoreAdapter struct
- Extended mCoreCallbacks struct with new callback types
- Added memory access callbacks in GBA memory read/write functions
- Added instruction execution callback in ARM CPU loop
- All callbacks use existing mScript infrastructure for consistency

PS: this code was generated with qwen3.5-122b

### DIFF
--- a/include/mgba/core/interface.h
+++ b/include/mgba/core/interface.h
@@ -23,6 +23,26 @@ CXX_GUARD_START
 		} \
 	} while (0);
 
+#define mCALLBACKS_INVOKE_IE(BOARD, PC, OPCODE, THUMB) do {\
+		size_t c; \
+		for (c = 0; c < mCoreCallbacksListSize(&(BOARD)->coreCallbacks); ++c) { \
+			const struct mCoreCallbacks* callbacks = mCoreCallbacksListGetConstPointer(&(BOARD)->coreCallbacks, c); \
+			if (callbacks->instructionExecuted) { \
+				callbacks->instructionExecuted(callbacks->context, PC, OPCODE, THUMB); \
+			} \
+		} \
+	} while (0);
+
+#define mCALLBACKS_INVOKE_MEM(BOARD, ADDR, VALUE, WRITE) do {\
+		size_t c; \
+		for (c = 0; c < mCoreCallbacksListSize(&(BOARD)->coreCallbacks); ++c) { \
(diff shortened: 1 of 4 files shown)


endrift:
We do not accept code written by LLMs.

Also this would tank performance.

Mte90:
Ok, good to know I wanted to improve the scripting APIs but I understand your point to LLMs (the reason why I wanted to disclose it).
About performance I saw also that, maybe I can improve it but I will use again LLMs because I am not skilled at all with C++ so I will not bother you again about that topic.

endrift:
I am accepting scripting API requests, and there are ways to go about _some_ of this without tanking performance, and the rest of it with only tanking performance when it's in use (which wasn't the case here). If you can give me a list of features you want I can toss them in the list.

Mte90:
Well I am looking for those APIs that this PR implemented but as you saying just call this APIs when they are used probably it's the first thing to do to avoid the majority of performance issues.
I am experimenting with an analyzer of what a rom is doing so those APIs are the missing piece for that.

Issue #3637, opened 2025-11-25
Improve UX when loading ROMs from a temporary/non-writable path


ahigerd:
[This commit](https://github.com/mgba-emu/mgba/commit/df1e8689b04afceb6a8340865ca86ad9987efc4e) introduced a warning that is displayed when loading a ROM from a temporary directory, which could happen if double-clicking a .gba file contained inside a .zip file. We also have the existing warning that is displayed when loading a ROM from a non-writable directory (e.g. Program Files due to UAC).

In both of these cases, the user is informed that they may lose save data, but mGBA doesn't offer any mechanism to improve on the situation; the user must understand the problem and manually move the ROM to a better place.

Instead, mGBA could instruct the user to copy the ROM to a better location and provide a file dialog to select a more appropriate path, performing the copy operation automatically. The chosen path should be revalidated for suitability.

This will not only make the solution easier for the user, it is much less likely that the user will click through the warning without paying attention. A user that _really_ doesn't want to do it can still click the Cancel button in the file dialog.

Polish ideas to consider:
* If there are any folders already added to the library, open the file dialog to one of the paths there by default.
* Store the chosen path in a setting so that it can be recalled next time.
* Maybe if the user has already configured the library, the game or the selected folder should be automatically added to it.
* A link to a FAQ might be a good idea.

PR #3754, opened 2026-04-03
Add new copy dialog when there are save issues detected (#3637)

endrift:
Can you review this @ahigerd?

### DIFF
--- a/src/platform/qt/CoreManager.cpp
+++ b/src/platform/qt/CoreManager.cpp
@@ -5,9 +5,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 #include "CoreManager.h"
 
+#include <QMessageBox>
+
+#include "GBAApp.h"
 #include "CoreController.h"
 #include "LogController.h"
 #include "VFileDevice.h"
+#include "utils.h"
 
 #include <QDir>
 
@@ -71,16 +75,41 @@ CoreController* CoreManager::loadGame(const QString& path) {
 		}
 		archive->close(archive);
 	}
-	QDir dir(info.dir());
-	QDir tmpdir(QDir::tempPath());
-	if (info.canonicalFilePath().startsWith(tmpdir.canonicalPath())) {
-		LOG(QT, ERROR) << tr("The ROM appears to be loaded from a temporary directory. This will likely lead to data loss (e.g. saves, screenshots, etc.) if you continue. "
-			"Please put the ROM in a more suitable location and then re-open it. If you are loading the ROM from an archive, please extract the archive first.");
-	}
 	if (!vf) {
 		// Open bare file
 		vf = VFileOpen(info.canonicalFilePath().toUtf8().constData(), O_RDONLY);
 	}
+
+	if (!vf) {
+		return nullptr;
+	}
+
+	QDir dir(info.dir());
+	QDir tmpdir(QDir::tempPath());
+	if (info.canonicalFilePath().startsWith(tmpdir.canonicalPath())) {
+		bool bad = false;
+		if (m_config) {
+			mCoreOptions opts;
+			mCoreConfigMap(m_config, &opts);
+			bad = bad || !opts.savegamePath;
+			bad = bad || !opts.savestatePath;
+			bad = bad || !opts.screenshotPath;
+			bad = bad || !opts.cheatsPath;
+			mCoreConfigFreeOpts(&opts);
+		} else {
+			bad = true;
+		}
+		if (bad) {
+			QString newPath = saveFailed(vf, tr("Temporary file loaded"),
+			                             tr("The ROM appears to be loaded from a temporary directory, perhaps automatically extracted from an archive (e.g. a zip file)."),
+			                             "*." + info.suffix());
+			if (!newPath.isEmpty()) {
+				vf->close(vf);
+				return loadGame(newPath);
+			}
+		}
+	}
+
 	return loadGame(vf, info.fileName(), dir.canonicalPath());
 }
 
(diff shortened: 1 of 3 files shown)


ahigerd:
With pleasure! :eyes:

src/platform/qt/CoreManager.cpp
```diff
@@ -178,7 +215,70 @@ CoreController* CoreManager::loadBIOS(int platform, const QString& path) {
...
-	cc->setPath(path, info.dir().canonicalPath());
 	emit coreLoaded(cc);
 	return cc;
 }
+
+QString CoreManager::saveFailed(VFile* vf, const QString& title, const QString& summary, const QString& filter) {
+	QMessageBox error;
+	error.setIcon(QMessageBox::Critical);
+	error.setWindowTitle(title);
+	error.setText(tr("%1 Would you like to move the ROM to a different location? If you don't, this will likely lead to data loss (e.g. saves, screenshots, etc.).").arg(summary));
```
ahigerd: s/move/copy/

src/platform/qt/CoreManager.cpp
```diff
@@ -178,7 +215,70 @@ CoreController* CoreManager::loadBIOS(int platform, const QString& path) {
...
+					ok = false;
+				}
+				break;
+			}
+		}
+
+		if (ok) {
+			return newPath;
+		}
+	}
```
ahigerd, line 275: Perhaps attempt to delete the destination file if it was created but not fully populated? I can imagine an edge case where you run into a disk-full condition and the partial ROM is eating up the last megabyte of space on a tiny SD card.

src/platform/qt/CoreManager.cpp
```diff
@@ -71,16 +75,41 @@ CoreController* CoreManager::loadGame(const QString& path) {
...
+			bad = bad || !opts.savestatePath;
+			bad = bad || !opts.screenshotPath;
+			bad = bad || !opts.cheatsPath;
+			mCoreConfigFreeOpts(&opts);
+		} else {
+			bad = true;
+		}
+		if (bad) {
+			QString newPath = saveFailed(vf, tr("Temporary file loaded"),
+			                             tr("The ROM appears to be loaded from a temporary directory, perhaps automatically extracted from an archive."),
```
ahigerd: Considering our userbase, perhaps "archive (e.g. zip file)"

src/platform/qt/CoreManager.cpp
```diff
@@ -178,7 +215,70 @@ CoreController* CoreManager::loadBIOS(int platform, const QString& path) {
...
+	error.setDefaultButton(QMessageBox::Ok);
+	if (error.exec() == QMessageBox::Ignore) {
+		return QString();
+	}
+
+	auto retry = [this]() {
+		QMessageBox retry;
+		retry.setIcon(QMessageBox::Critical);
+		retry.setWindowTitle(tr("Copy failed"));
+		retry.setText(tr("Failed to copy ROM. Do you want to try again?"));
```
ahigerd: The engineer in me would like to be able to report _why_ the copy failed. `QIODevice::errorString()` might be a good place to start, though you might have better luck with `QFileDevice::error()` mapping into something where you can control/localize the messages.

src/platform/qt/CoreManager.cpp
```diff
@@ -178,7 +215,70 @@ CoreController* CoreManager::loadBIOS(int platform, const QString& path) {
...
+	error.setWindowTitle(title);
+	error.setText(tr("%1 Would you like to move the ROM to a different location? If you don't, this will likely lead to data loss (e.g. saves, screenshots, etc.).").arg(summary));
+	error.setStandardButtons(QMessageBox::Ok | QMessageBox::Ignore);
+	error.setDefaultButton(QMessageBox::Ok);
+	if (error.exec() == QMessageBox::Ignore) {
+		return QString();
+	}
+
+	auto retry = [this]() {
+		QMessageBox retry;
```
ahigerd: I broadly prefer to use `QMessageBox::critical()` over constructing the object longhand if I'm going to be using `exec()` instead of `open()`. Your codebase, your style preference, but I think that's how it's been done elsewhere as well.

src/platform/qt/CoreManager.cpp
ahigerd: Stylistically I would also consider `%1\n\n` for readability.

ahigerd: I might also suggest `summary + "\n\n" + tr("...")` to keep the first line out of the translation, but that's entirely optional.

src/platform/qt/CoreManager.cpp
```diff
@@ -178,7 +215,70 @@ CoreController* CoreManager::loadBIOS(int platform, const QString& path) {
 	if (m_multiplayer) {
 		cc->setMultiplayerController(m_multiplayer);
 	}
-	cc->setPath(path, info.dir().canonicalPath());
 	emit coreLoaded(cc);
 	return cc;
 }
+
+QString CoreManager::saveFailed(VFile* vf, const QString& title, const QString& summary, const QString& filter) {
+	QMessageBox error;
```
ahigerd: Ditto what I said below about using the static method.

src/platform/qt/CoreManager.cpp
endrift: Yeah, I had tried to figure out how to make that clearer. I guess that works.

src/platform/qt/CoreManager.cpp
endrift: I don't think we need to do that, at least not yet.

src/platform/qt/CoreManager.cpp
endrift: the signature is so long that you kinda have to know what the order is, but it is what I do already elsewhere in the codebase I suppose.

Issue #1871, opened 2020-08-31
Sprites updated with HDMA differs from hardware


GValiente:
Test rom:
[test.zip](https://github.com/mgba-emu/mgba/files/5148306/test.zip)

The first sprite lines in the image with length = 64 pixels are not shown on hardware.
Those lines are present in the last line of the sprite's image data.

![test](https://user-images.githubusercontent.com/13147012/91683188-30661c80-eb54-11ea-9f09-73c7219e90d3.png)

GValiente:
New [test ROM](https://github.com/mgba-emu/mgba/files/5524058/mgba_1871.zip).

The two biggest lines in the middle of the diamond are not visible on a real GBA:

![mgba](https://user-images.githubusercontent.com/13147012/98814175-9e1ca780-2425-11eb-8dcf-5b3719aaa336.png)

fleroviux:
NBA also had this issue. The fix is to emulate the OAM update delay.
Like on the DS and tested by your OAM update delay test:
Render OAM one scanline ahead (for the next line) starting in the last line of V-blank.

See this commit: https://github.com/fleroviux/NanoboyAdvance/commit/eda1af70535c1dc0d63168b7c350a5f64db02acb

Issue #2048, opened 2021-02-22
HDMA sprite border not present on hardware


GValiente:
[Test ROM](https://github.com/mgba-emu/mgba/files/6025000/cube_test.zip)

The line that crosses the cube (drawn with sprites written with HDMA) is not present on hardware, because it is below other sprites:

![cube_test](https://user-images.githubusercontent.com/13147012/108767041-2e56a680-7556-11eb-8284-70bf1a1f83ba.png)

endrift:
I wonder if this is because I don't yet draw sprites a scanline early. Do you know if you can reproduce it with a setup that doesn't use HDMA?

GValiente:
I'm sure I can't replicate it without using HDMA.

fleroviux:
> wonder if this is because I don't yet draw sprites a scanline early.

From looking at this in NBA, it does look like drawing sprites a scanline early would fix this.

endrift:
I am _pretty_ sure this is a duplicate of #1871 with just a different test ROM

endrift:
Confirmed duplicate per #3756

PR #3756, opened 2026-04-05
Implement HBlank OAM staging behaviour to match hardware

felixjones:
Encountered the same issue in a game project I am working on where 1px thick lines rendered with OAM sprites are rendered on the wrong scanline

Fixes https://github.com/mgba-emu/mgba/issues/1871 by staging OAM to the renderer 1 scanline earlier

This is similar to how NanoBoyAdvance solved it https://github.com/nba-emu/NanoBoyAdvance/commit/eda1af70535c1dc0d63168b7c350a5f64db02acb

A lot of this is a copy-paste of `*WriteOAM` and then modified for the staging concept, so the approach might be a bit unwieldy

And I couldn't for the life of me get the cinema locally set up, so that cinema test I added was a guess and uses a screenshot from this particular change

### DIFF
--- a/src/gba/renderers/video-software.c
+++ b/src/gba/renderers/video-software.c
@@ -1046,23 +1046,12 @@ static void _stageSpriteLayer(struct GBAVideoSoftwareRenderer* softwareRenderer,
 
 	softwareRenderer->forceTarget1 = false;
 
-	int savedNWindows = softwareRenderer->nWindows;
-	struct Window savedWindows[MAX_WINDOW];
-	memcpy(savedWindows, softwareRenderer->windows, sizeof(savedWindows));
-	struct WindowControl savedCurrentWindow = softwareRenderer->currentWindow;
-	softwareRenderer->nWindows = 1;
-	softwareRenderer->windows[0].endX = GBA_VIDEO_HORIZONTAL_PIXELS;
-	softwareRenderer->windows[0].control.packed = 0xFF;
-
 	int32_t savedCycles = softwareRenderer->spriteCyclesRemaining;
 	softwareRenderer->spriteCyclesRemaining = GBARegisterDISPCNTIsHblankIntervalFree(softwareRenderer->dispcnt) ? OBJ_HBLANK_FREE_LENGTH : OBJ_LENGTH;
 
 	softwareRenderer->spriteLayerMask = GBAVideoSoftwareRendererPreprocessSpriteLayer(softwareRenderer, y);
 
 	softwareRenderer->spriteCyclesRemaining = savedCycles;
-	softwareRenderer->nWindows = savedNWindows;
-	memcpy(softwareRenderer->windows, savedWindows, sizeof(savedWindows));
-	softwareRenderer->currentWindow = savedCurrentWindow;
 }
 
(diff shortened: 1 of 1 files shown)


src/gb/extra/proxy.c
```diff
@@ -200,6 +200,8 @@ static bool _parsePacket(struct mVideoLogger* logger, const struct mVideoLoggerD
 			break;
 		}
 		break;
+	case DIRTY_STAGE_OAM:
+		break; // GB (probably) does not have staged OAM updates
```
felixjones: Feels like something that needs verifying

endrift: It does not. It does dumb shit with pausing the dot clock.

endrift:
This probably also fixes #2048. Please cross-check.

fwiw, I'm not sure how I feel about this approach. I also don't intend to merge this before 0.11 regardless, but it is something I was targeting for fixing in 0.12 regardless.

felixjones:
> This probably also fixes #2048. Please cross-check.

Yes can confirm #2048 is also fixed

cube_test-0

> fwiw, I'm not sure how I feel about this approach. I also don't intend to merge this before 0.11 regardless, but it is something I was targeting for fixing in 0.12 regardless.

I struggled to think of a better approach, considering real OAM writes still need to happen. Double buffering them was all I could think of

endrift:
The way I was planning on implementing this is by doing what actually happens on hardware: it renders the OAM to another buffer a frame early and then compositing into the final image, then rendering the current state of OAM to that buffer. This may require another buffer to do efficiently, but so be it.

endrift:
That way I believe does not require modifying the API, which is something I want to avoid here since it would break mVLs

felixjones:
> The way I was planning on implementing this is by doing what actually happens on hardware: it renders the OAM to another buffer a frame early and then compositing into the final image, then rendering the current state of OAM to that buffer. This may require another buffer to do efficiently, but so be it.

Looked into rendering the incoming line of OAM at HBlank start with the goal of not modifying the API, feels much cleaner as it's shuffling existing code around, but what do you think?

src/gba/renderers/video-software.c
```diff
@@ -1016,6 +1025,46 @@ void GBAVideoSoftwareRendererPostprocessBuffer(struct GBAVideoSoftwareRenderer*
...
+
+	softwareRenderer->forceTarget1 = false;
+
+	int savedNWindows = softwareRenderer->nWindows;
+	struct Window savedWindows[MAX_WINDOW];
+	memcpy(savedWindows, softwareRenderer->windows, sizeof(savedWindows));
+	struct WindowControl savedCurrentWindow = softwareRenderer->currentWindow;
+	softwareRenderer->nWindows = 1;
+	softwareRenderer->windows[0].endX = GBA_VIDEO_HORIZONTAL_PIXELS;
+	softwareRenderer->windows[0].control.packed = 0xFF;
```
felixjones: This is a big hack to pre-render all sprites regardless of the hardware windows. Would be nice to get rid of it

felixjones: This does not inspire confidence. This hack was in place because I was noticing issues with windows not clipping sprites correctly, and I was investigating that some more only to notice the code isn't needed. I must have squash away whatever issue I was facing

There goes my confidence in this PR. I think this will be a rabbit hole of debugging cases like this.

PR #3759, opened 2026-04-08
Updated german readme to the english one

tolik518:
I figured that the German README has gotten a little bit dated, so I updated it with the additions and changes that were done to the original README.


endrift:
Thanks!

PR #3784, opened 2026-05-18
fix(lua/gb): emu:readRegister() returns garbage in high bytes

bebecue:
while playing with lua scripting with the gb core, i found sometimes emu:readRegister() returns 16 bits value for 8 bit registers.

assuming cpu.b = 0x12, emu:readRegister("b") might return 0x12, 0x7F12 or any others.

the issue is in _GBCoreReadRegister(), param `void* out` is casted to `uint8_t*`, and `*value = cpu->b` will only write to the lowest byte, leave whatever garbage in high bytes untouched.


bebecue:
I forget to switch to another branch, github is telling me "Merging is blocked", closing this one.

PR #3829, opened 2026-08-08
android port

Ishan09811:


.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
...
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
```
ahigerd: Pin to hashes instead of version tags. It's a security issue.

.idea/caches/deviceStreaming.xml
ahigerd, line 1: If this is a cache, it probably shouldn't be committed.

src/platform/android/app/src/main/assets/nointro.dat
ahigerd, line 1: If this is the path it _has_ to be in, can you use a build step to copy it out of `res/` instead of keeping a duplicate?

src/platform/android/app/src/main/cpp/audio/oboe_audio_player.cpp
```diff
@@ -0,0 +1,192 @@
+#include "oboe_audio_player.h"
+#include <android/log.h>
+#include <algorithm>
+#include <chrono>
+
+#define LOG_TAG "oboe_audio_player"
+#define LOGI(...) __android_log_print(ANDROID_LOG_INFO, LOG_TAG, __VA_ARGS__)
+#define LOGD(...) __android_log_print(ANDROID_LOG_DEBUG, LOG_TAG, __VA_ARGS__)
+#define LOGE(...) __android_log_print(ANDROID_LOG_ERROR, LOG_TAG, __VA_ARGS__)
```
ahigerd: See `src/platform/qt/LogController.cpp` around line 58 for how to configure mGBA's built-in logging system. Should be pretty simple, and then you can pull full mGBA logs using logcat.

ahigerd:
Coding style notes: (I'm going to edit this comment as I see things to bring up instead of making a ton of individual comments)

* Use hard tabs in C/C++ code (I know, I would prefer spaces too; if I have to deal with it so do you :stuck_out_tongue:)
* Use an `m_` prefix on private C++ member variables.
* We try to use the `override` keyword on overridden C++ virtual functions.
* No one-line `if` statements -- always put the code on a second line. (And definitely don't put multiple semicolon-delimited statements on the same line.)
* Don't forget the license boilerplate at the top of each file. (You're welcome to put yourself in as the copyright author.)

src/platform/android/app/src/main/cpp/audio/oboe_audio_player.h
```diff
@@ -0,0 +1,130 @@
+#pragma once
+
+#include <oboe/Oboe.h>
+#include <algorithm>
+#include <atomic>
+#include <cstdint>
+#include <vector>
+
+class SpscRingBuffer {
```
ahigerd: Move this into the .cpp file since it's not used anywhere else.

src/platform/android/app/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
+cmake_minimum_required(VERSION 3.22.1)
+project(mgba-android)
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_C_STANDARD 11)
+set(CMAKE_C_STANDARD_REQUIRED ON)
+
+# android studio causes freezes when using CMAKE_SOURCE_DIR
```
ahigerd, line 17: Ewwww ^^() Oh well. We do what we have to. Can't complain.

src/platform/android/app/src/main/cpp/core.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+JNIEXPORT jint JNICALL
+Java_org_mgba_1emu_mgba_core_Core_nativeGetPlatform(JNIEnv* env, jobject /*thiz*/) {
+    std::lock_guard<std::mutex> lock(g_coreMutex);
+    return g_core ? g_core->getPlatform() : -1;
+}
+
+JNIEXPORT void JNICALL
+Java_org_mgba_1emu_mgba_core_Core_nativeSetConfigInt(JNIEnv* env, jobject thiz, jstring jKey, jint value) {
+    const char* key = env->GetStringUTFChars(jKey, nullptr);
```
ahigerd: I'm not going to _insist_ on it (unless endrift does) but the explicit get/release semantics would definitely benefit from a RAII container. Maybe something like:

```c++
class JavaUTFString : public std::string_view
{
public:
    JavaUTFString(JNIEnv* env, jstring jKey)
    : std::string_view(env->GetStringUTFChars(jKey, nullptr)), m_env(env), m_jKey(jKey)
    {
        // nothing to do
    }

    ~JavaUTFString()
    {
        m_env->ReleaseStringUTFChars(m_jKey, data());
    }

private:
    JNIEnv* m_env;
    const jstring& m_jKey;
};
```

As a `std::string_view` subclass this will provide all of the C++ string semantics you would like (including using `==` for string comparison) with minimal overhead.

src/platform/android/app/src/main/cpp/core_interface.cpp
```diff
@@ -0,0 +1,541 @@
...
+#include <mgba-util/audio-resampler.h>
+#include <mgba/internal/gb/gb.h>
+#include <mgba-util/vfs.h>
+}
+
+#define LOG_TAG "core"
+#define LOGI(...) __android_log_print(ANDROID_LOG_INFO, LOG_TAG, __VA_ARGS__)
+#define LOGE(...) __android_log_print(ANDROID_LOG_ERROR, LOG_TAG, __VA_ARGS__)
+
+class CoreImpl : public CoreInterface {
```
ahigerd: Going to echo what we talked about in Discord just to keep track of it in the PR -- is there a compelling reason to use an interface class with a single implementation?

src/platform/android/app/src/main/cpp/core_interface.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+    void configureCoreAudioBuffer(int sourceRate) {
+        if (!m_core || sourceRate <= 0) return;
+
+        const double samplesPerFrame =
+                static_cast<double>(sourceRate) *
+                static_cast<double>(m_core->frameCycles(m_core)) /
+                static_cast<double>(m_core->frequency(m_core));
+
+        auto bufferFrames = static_cast<size_t>(std::ceil(samplesPerFrame * 2.0));
```
ahigerd: I think the `* 2` belongs outside of the conversion to int so you don't end up with an odd-length buffer.

ahigerd:
Not relevant for the MVP, of course, but how's i18n going to end up looking for this in the long run? We should ideally integrate it into the i18n workflow that's already in place.

ahigerd:
Kotlin feels a lot like Java-flavored TypeScript.

src/platform/android/app/src/main/java/org/mgba_emu/mgba/renderer/gl/EmulationThread.kt
```diff
@@ -0,0 +1,113 @@
...
+
+    @Volatile
+    var paused: Boolean = false
+
+    private var framesThisSecond = 0
+    private var lastFpsTimestampNs = 0L
+
+    override fun run() {
+        running.set(true)
+        val frameIntervalNs = NANOS_PER_SECOND / GBA_FPS
```
ahigerd: Nitpick: This is a constant, you could put it down in the companion.

src/platform/android/app/src/main/java/org/mgba_emu/mgba/utils/SaveDataStore.kt
```diff
@@ -0,0 +1,39 @@
...
+    fun save(gameCode: String, saveBytes: ByteArray): Boolean {
+        if (saveBytes.isEmpty()) return true
+        val target = fileFor(gameCode)
+        val temp = File(saveDir, "${target.name}.tmp")
+        return try {
+            temp.writeBytes(saveBytes)
+            temp.renameTo(target)
+        } catch (_: IOException) {
+            false
+        }
```
ahigerd: Feels like this deserves a worker function (`saveBytesToFile` or something) that can be shared across the various places you use the pattern.

src/platform/android/gradle/wrapper/gradle-wrapper.jar
ahigerd, line 1: Is it really necessary to include this binary in the repo? If so, what is it?

ahigerd:
See what you can do to keep the generated files out of the repo if they can easily be rebuilt. (I also question the necessity of the IDE configuration files, but I'll leave that to @endrift if it's okay or not.)

ahigerd:
Okay, done with my review. Really not a whole lot to critique outside of style issues. Well done!

src/platform/android/app/src/main/cpp/audio/oboe_audio_player.h
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        readIndex.store(r + toRead, std::memory_order_release);
+        return toRead;
+    }
+
+private:
+    static constexpr size_t kChannels = 2;
+
+    static size_t nextPowerOfTwo(size_t v) {
```
ahigerd: Turns out an equivalent (but faster) function exists in mGBA already as `toPow2()`.

src/platform/android/app/src/main/assets/nointro.dat
Ishan09811: i am already doing it, just forgot add in .gitignore :)

src/platform/android/gradle/wrapper/gradle-wrapper.jar
Ishan09811: yeah, otherwise people will not be able to build it :)

src/platform/android/app/src/main/cpp/audio/oboe_audio_player.h
Ishan09811: then how i will be able to define this ``std::unique_ptr<SpscRingBuffer> ringBuffer`` in the header file ?

ahigerd: With a forward declaration: `class SpscRingBuffer;` is enough to declare the unique_ptr member, and the full definition will be visible everywhere it gets used.

src/platform/android/app/src/main/cpp/core.h
```diff
@@ -0,0 +1,125 @@
...
+class Core {
+public:
+	Core() {
+		s_audioRateOwner.store(this, std::memory_order_release);
+	}
+
+	~Core() {
+		Core* expected = this;
+		s_audioRateOwner.compare_exchange_strong(expected, nullptr, std::memory_order_acq_rel, std::memory_order_acquire);
+	}
```
ahigerd: Style nitpick: These aren't quite trivial enough to inline, IMO. Put the implementations into the .cpp file.

src/platform/android/app/src/main/cpp/core.h
```diff
@@ -0,0 +1,125 @@
...
+
+	static constexpr size_t kMaxCoreAudioBufferFrames = 0x4000;
+	static constexpr size_t kResampledAudioBufferFrames = 1024;
+	static constexpr size_t kAudioDrainFrames = 1024;
+	static constexpr size_t kMaxAudioProcessIterations = 2;
+	static constexpr size_t kAudioRingBufferFrames = 4096;
+
+	uint32_t getRomCRC32() const;
+
+	enum Platform { PLATFORM_UNKNOWN = -1, PLATFORM_GB = 1, PLATFORM_GBC = 2, PLATFORM_SGB = 3, PLATFORM_GBA = 4 };
```
ahigerd: Style nitpick: Go ahead and expand this to one per line.

ahigerd:
You missed spaces vs. tabs in `oboe_audio_player` (and the `.cpp` has mixed whitespace).

src/platform/android/app/src/main/cpp/core.cpp
```diff
@@ -0,0 +1,455 @@
...
+std::vector<uint8_t> Core::exportSaveData() {
+	if (!m_core)
+		return { };
+	void* sram = nullptr;
+	size_t size = m_core->savedataClone(m_core, &sram);
+	if (size == 0 || sram == nullptr) {
+		return { };
+	}
+
+	std::vector<uint8_t> result(static_cast<uint8_t*>(sram), static_cast<uint8_t*>(sram) + size);
```
ahigerd, line 307: ~~Very much a nitpick, but it's _slightly_ faster (and slightly less code) to use the `(const T* data, size_t size)` overload than the `(ForwardIterator start, ForwardIterator end)` overload.~~

ahigerd: Wait, disregard, this is vector, not string.

src/platform/android/app/src/main/cpp/no_intro_parser.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+void noIntroShutdown() {
+	if (g_noIntroDB) {
+		NoIntroDBDestroy(g_noIntroDB);
+		g_noIntroDB = nullptr;
+	}
+}
+
+bool noIntroLookupCRC32(uint32_t crc32, NoIntroMetadata& out) {
```
ahigerd: Oh, I missed this before. Pass out parameters by pointer, not by reference.

(Also maybe this should be `[[no_discard]]` so callers can't ignore errors.)

ahigerd: (Rationale: You can't tell the difference at the call site between a mutable reference and a const reference. Having to write `&metadata` makes it more obvious that the value will be mutated.)

src/platform/android/app/src/main/cpp/audio/oboe_audio_player.cpp
```diff
@@ -93,6 +93,12 @@ class SpscRingBuffer {
 	std::atomic<uint64_t> underrun{0};
 };
 
+OboeAudioPlayer::OboeAudioPlayer() = default;
```
ahigerd, line 100: Huh. I didn't know you could do that. I'm not _complaining_. I've just always put `= default` in the header file.

src/platform/android/app/src/main/cpp/core.cpp
```diff
@@ -0,0 +1,461 @@
...
+}
+
+void Core::setConfigInt(const char* key, int value) {
+	if (!m_core)
+		return;
+	mCoreConfigSetIntValue(&m_core->config, key, value);
+}
+
+void Core::setConfigString(const char* key, const char* value) {
+	mCoreConfigSetValue(&m_core->config, key, value);
```
ahigerd, line 380: Missing `!m_core` check

src/platform/android/app/src/main/cpp/jni_bridge.cpp
```diff
@@ -0,0 +1,230 @@
...
+}
+
+JNIEXPORT jint JNICALL
+Java_org_mgba_1emu_mgba_core_Core_nativeGetPlatform(JNIEnv* env, jobject /*thiz*/) {
+    std::lock_guard<std::mutex> lock(g_coreMutex);
+    return g_core ? g_core->getPlatform() : -1;
+}
+
+JNIEXPORT void JNICALL
+Java_org_mgba_1emu_mgba_core_Core_nativeSetConfigInt(JNIEnv* env, jobject thiz, jstring jKey, jint value) {
```
ahigerd, line 207: No mutex guard or `!g_core` check... but if it doesn't _need_ a mutex guard, then that might be okay?

ahigerd: Also with the next two functions

src/platform/android/app/src/main/cpp/utils/jni_string.h
```diff
@@ -0,0 +1,24 @@
+#include <jni.h>
```
ahigerd, line 12: Missing `#pragma once`

src/platform/android/app/src/main/cpp/jni_bridge.cpp
Ishan09811: i never call these functions after emulation has been started so no problem probably

src/platform/android/app/src/main/java/org/mgba_emu/mgba/utils/FileUtils.kt
```diff
@@ -0,0 +1,29 @@
+
+package org.mgba_emu.mgba.utils
+
+import java.io.File
+import java.io.IOException
+
+object FileUtils {
+    fun File.writeBytesAtomically(bytes: ByteArray): Boolean {
```
ahigerd, line 25: Oh, huh. Extension methods like C#. That's kinda cool.

.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  build:
```
ahigerd, line 10: @endrift Do you think it would be a good idea to add a conditional to this workflow to only run the build if the Android frontend or a shared dependency (core, resources) are modified? Or is it not likely to matter and we can just let it run unconditionally?

ahigerd (approved):
The Oboe files still have funky whitespace but other than that I think I'm satisfied with the current state of this branch.

PR #3836, opened 2026-08-24
Interpolator fix

Photosounder:
I asked ChatGPT 5.6 Sol to inspect the audio interpolation code and it found some issues. Here's what it says:

> The sinc interpolator calculated its kernel incorrectly:
> 
> - `sinc(0)` was initialized to `0` instead of its limiting value, `1`.
> - The sinc lookup table was truncated after `width` radians rather than spanning the full window.
> - Fractional offsets discarded the sample index’s sign, producing incorrect and asymmetric kernel distances.
> - Upsampling scaled the kernel unnecessarily, while downsampling lacked the required anti-aliasing cutoff.
> 
> The fix builds the complete sinc table, calculates each tap from its signed distance to the fractional source position, preserves the source bandwidth when upsampling, and applies an appropriate low-pass cutoff when downsampling.
> 
> Regression tests cover exact source positions, fractional impulse symmetry, and Nyquist-frequency rejection during downsampling.

### DIFF
--- a/src/util/CMakeLists.txt
+++ b/src/util/CMakeLists.txt
@@ -46,6 +46,7 @@ set(TEST_FILES
 	test/geometry.c
 	test/hash.c
 	test/image.c
+	test/interpolator.c
 	test/sfo.c
 	test/string-parser.c
 	test/string-utf8.c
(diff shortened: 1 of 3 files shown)


endrift:
We do not accept AI contributions.

Photosounder:
lmao okay then take this as a bug report. You can look yourself at the code and see that the peak of the sinc function shouldn't be zero for starters.

endrift:
I asked the person who wrote the reference code for that for input and he said yes, there are issues, but
image

PR #3840, opened 2026-08-26
Qt: Add options for audio-buffer-sizes of 48 and 32 samples

72374:
I did notice a small difference in delay between an audio-buffer-size of 64 samples and 32 samples.
Would be nice if these 2 options for 32 samples and 48 samples would be added too. :)

I do not know how long the actual delays were, when i compared those. Sound-cards typically can use 2 or 3 buffers, and there is other processing, such as a mixer, for combining the audio of multiple applications; so the difference of the delay between buffer-sizes can increase. 3 buffers of 64 samples at 48 kHz would be 4 ms, which seems long enough to be perceivable.

32 samples was the lowest audio-buffer-size that worked properly on my computer while using USB-connected speakers; lower values (16 and 24) resulted in seemingly slowed down audio.

Playing games (F-Zero, and Kirby & The Amazing Mirror) with an audio-buffer-size of 32 samples, and Sync set to Audio, worked fine, without audible issues with the sound, even with 27x resolution, as long as the energy-mode of my computer was set to Performance, and rewinding was disabled.

Operating-system: Debian GNU/Linux Testing (Forky)
CPU: Core i7-12700K
GPU: Radeon RX 6750 XT
Audio-device: Creative Pebble Pro

### DIFF
--- a/src/platform/qt/SettingsView.ui
+++ b/src/platform/qt/SettingsView.ui
@@ -140,8 +140,18 @@
                <string>1536</string>
               </property>
               <property name="currentIndex">
-               <number>9</number>
+               <number>11</number>
               </property>
+              <item>
+               <property name="text">
+                <string>32</string>
+               </property>
+              </item>
+              <item>
+               <property name="text">
+                <string>48</string>
+               </property>
+              </item>
               <item>
                <property name="text">
                 <string>64</string>


endrift:
4ms is one quarter of a frame. I'm not really sure how that would be perceptible.

72374:
On its own, i also consider 4 ms below what is meaningfully perceivable, but it is not the total delay.
It is the lowest additional delay i expect to be there; it could be much longer, depending on software and hardware.
Is there a reliable way to measure the total audio-latency from mGBA and the audio-processing by the system?
Then i could measure that.

According to an automatic audio-latency-test in a webbrowser, the audio-latency for the audio-output on my system varies between at least 29 ms and 60 ms; with consistent timing during a single test, but sometimes with different results when trying again.
If the test works properly, the difference i perceived could be due to such a variation; but the difference seems fairly consistent to me.
Though, even if the difference in the delay caused by the buffer-size is not particularly noticeable on its own, lowering it, could make it more likely that the delay is sufficiently low.

72374:
The difference in delay that i noticed, was simply additional visual input-lag with a buffer-size of 64 samples, compared to 32 samples. Even without sound, the difference in responsiveness is noticeable when i compare the options. Without that input-lag, the timing of the audio also aligns properly with the visuals (which is why i was focused on the audio when writing the messages in this pull-request).

---

When using Bluetooth-connected headphones (with over 100 ms of audio-delay), sound-effects were noticeably more delayed compared to the images when using a buffer-size of 32 samples instead of 64 samples.
For synchronisation between audio and video, this seems to be an entire frame (~16.67 ms) of timing-difference, ± whatever the time-difference in sound-processing is, depending on wether audio or video is delayed.
The difference in audio-delay, when comparing sound-effects of a game to the sound of pressing keys, was almost not noticeable.

---

With a USB-soundcard/-amplifier (Creative Sound Blaster Play! 4), a buffer-size of 32 samples also worked fine. 

When using a headphone-connector of a "Gigabyte Z690 GAMING X DDR4"-mainboard, audio did crackle for a few seconds after starting mGBA for the first time after starting the computer, while buffer-size was set to 32 or 48, but after that, the audio was normal, even after restarting mGBA.
I guess that some part of the audio-processing-stack did detect the buffer-underruns and adjusted the audio-buffers accordingly. This might also be the reason for the differences in delay measured with the tool, that i mentioned in my previous message. The other delay-comparisons mentioned in this message, were done before using the headphone-connector on the mainboard during the same session.

PR #3849, opened 2026-09-04
CMake: Fix minizip unztell64 detection outside default library paths

Davey-Hughes:
The baseline build breaks on MacOS with pkg-config from homebrew.

## Technical Details
MINIZIP_LIBRARIES is a bare library name, so check_function_exists linked the probe without the directory pkg-config reported. On a prefix the linker does not search by default the probe fails to link, HAVE_UNZTELL64 comes back false, and the workaround below renames unztell64 to unzTell64 against a minizip that exports the lowercase name, so libmgba fails to link.

Pass MINIZIP_LIBRARY_DIRS to the probe as -L flags. CMAKE_REQUIRED_FLAGS is used instead of CMAKE_REQUIRED_LINK_OPTIONS because the latter needs CMake 3.14 and this project requires only 3.12.


Davey-Hughes:
I'm not sure if this PR should target master due to the merge authorization- lmk if I should change it.

PR #3862, opened 2026-09-16
Switch: Fix crash on exit

JostenSyon:
Exiting mGBA on Switch (in-app Exit, both via forwarder and hbmenu) data-aborts with 2168-0002.

Two problems in the audio teardown in `src/platform/switch/main.c`:

- `audoutExit()` was called without a matching `audoutInitialize()` (the port uses audren/audrv), so drop it.
- The audio mempool was never detached. Shut the voice down, detach the mempool *before* closing the driver (detaching after close dereferences cleared driver state and aborts inside `audrvMemPoolDetach`), stop the audio renderer and free the buffers.

Verified on hardware: crash reproduced on exit with the old teardown (PC in `audrvMemPoolDetach`, null deref), gone with this order. Built with devkitA64/libnx.


src/platform/switch/main.c
```diff
@@ -1103,15 +1103,18 @@ int main(int argc, char* argv[]) {
 
 	mGUIDeinit(&runner);
 
+	audrvVoiceStop(&audrenDriver, 0);
+	audrvMemPoolDetach(&audrenDriver, mempool);
```
endrift, line 1107: I'm pretty sure neither of these lines are necessary. The voice is already stopped when a game is unloaded, and I don't think the mempool matters at all if the voice isn't doing anything. The Close function will clean it up anyway.

endrift (changes requested):

src/platform/switch/main.c
```diff
@@ -1103,15 +1103,18 @@ int main(int argc, char* argv[]) {
...
 	audrvClose(&audrenDriver);
+	audrenStopAudioRenderer();
```
endrift, line 1109: This line is also unnecessary given the following line, per the switch-examples.

endrift:
I'm guessing this was written by Claude, which is disallowed.

endrift:
Also you said the crash happens in a function that doesn't get called previously? It doesn't crash normally (audoutExit is a no-op if audout never got initialized), so I'm not sure what is going on here.

mGBA does not have an official forwarder either, so it seems like whatever's going on is not part of some official port.

PR #3867, opened 2026-09-20
WSOLA + low pass filter

Davey-Hughes:
I recently implemented a WSOLA + low pass filter in RetroArch, and wanted to port the work to this project as well. With the way RetroArch works, there isn't a single PR to point to, but [this is one of them, for example](https://github.com/libretro/RetroArch/pull/19554).

The WSOLA time stretch algorithm allows for resampling the audio when changing speed, which preserves pitch rather than increasing pitch or dropping samples (which causes crackling). The low pass filter can help cut out harsh high pitched sounds that may still remain when playing sped up.

image

The default value for the low pass filter is 22KHz which is a decent default, but allowing for Hz settings can be a bit confusing and is different from how RetroArch ended up implementing the low pass filter. 48KHz is a sentinel value that turns the low pass filter off (though it's essentially not in effect above 22KHz). There's definitely an argument for simplifying this option which I can do if it's desired.

I can add some more technical details to the PR description if needed.


shenef:
Did you read CONTRIBUTING.md?

Davey-Hughes:
> Did you read CONTRIBUTING.md?

Yes- is there an issue with something like the license or coding style?

endrift:
We don't accept LLM-written code.

PR #3871, opened 2026-09-21
Update StartupWMClass to match the application

electricbrass:
At some point, the app_id was updated to an rDNS style id, but the desktop file was not updated to match, resulting in missing icons and other issues from the desktop file and the window not being matched.


electricbrass:
Totally missed the existing PR #3701 for this

